PR #10, opened 2026-02-02
Add missing rows.Err() check in EnsureParticipantsBatch

etbyrd:
Go best practices: also matches the pattern used in MessageExistsBatch 20 lines above.


wesm (approved):
LGTM. Just rebased to kick off a new CI build

Issue #23, opened 2026-02-03
Add list-accounts command


galligan:
There's no way to see which accounts are configured without poking around in `~/.msgvault/tokens/`. 

**Proposal:** Add `msgvault list-accounts` that shows configured accounts with basic stats.

```bash
$ msgvault list-accounts
ACCOUNT                  MESSAGES    LAST SYNC
personal@gmail.com       234,567     2024-02-02 10:15
work@company.com         45,123      2024-02-02 10:15
```

With `--json`:
```json
[
  {"email": "personal@gmail.com", "message_count": 234567, "last_sync": "2024-02-02T10:15:00Z"},
  {"email": "work@company.com", "message_count": 45123, "last_sync": "2024-02-02T10:15:00Z"}
]
```

Useful for:
- Verifying which accounts are set up
- Scripting (pipe to other commands)
- Debugging sync issues

etbyrd:
https://github.com/wesm/msgvault/pull/24

PR #24, opened 2026-02-03
Account CLI updates

etbyrd:
- there was no `list-accounts` command
  - Added `list-accounts` that shows all accounts, and a flag, `--json` for programmatic use
  - Added fields `last_sync` and `messages` count requested in https://github.com/wesm/msgvault/issues/23
- `display_name` was in the schema but not exposed:
  - Added a new function for setting/updating display name `UpdateSourceDisplayName`
  - Added a `--display-name` flag to `add-account`
  - Added a new command `update-account` that uses `UpdateSourceDisplayName`
  
  preview of `list-accounts`:
  
Screenshot 2026-02-02 at 4 58 51 PM

### DIFF
--- a/cmd/msgvault/cmd/addaccount.go
+++ b/cmd/msgvault/cmd/addaccount.go
@@ -62,9 +62,14 @@ Examples:
 		if oauthMgr.HasToken(email) {
 			// Still create the source record - needed for headless setup
 			// where token was copied but account not yet registered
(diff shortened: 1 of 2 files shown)


wesm:
Thank you, I will get to this in the next couple days, don't worry about the rebase conflict I'll take care of it and push changes here

PR #47, opened 2026-02-04
Fixed path traversal bug

hughdbrown:
Vulnerability Fixed: Path traversal in attachment export (internal/export/attachments.go:118)

  Changes made:

  1. Added ValidateContentHash() function (internal/export/attachments.go:22-36) - Validates that content hashes are exactly 64 hexadecimal characters (valid SHA-256 hash format). This prevents path traversal attacks like ../../../etc/passwd.
  2. Updated Attachments() function to use the new validation instead of the minimal len < 2 check.
  3. Added two new tests:
    - TestPathTraversalInContentHash - Demonstrates the attack is blocked
    - TestContentHashValidation - Table-driven tests covering valid hashes, path traversal sequences, slashes, backslashes, wrong lengths, non-hex chars, and null bytes
  4. Updated existing tests to use valid 64-char hex hashes in both internal/export/attachments_test.go and internal/tui/actions_test.go.

### DIFF
--- a/internal/export/attachments.go
+++ b/internal/export/attachments.go
@@ -4,6 +4,8 @@ package export
 
 import (
 	"archive/zip"
+	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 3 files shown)


wesm:
Thanks for investigating and fixing this

PR #53, opened 2026-02-04
Fix cache after deletion

robelkin:
As part of the deletions that I was doing, I noticed that there was a cache invalidation issue where the Parquet cache wasn't updated, and so the deletions (if they'd only partially been done or been done on a different filter for some of the emails) were just getting deleted a second time, which is obviously not super necessary. This is a quick fix to take care of that.


wesm (approved):
Good catch. I just rebased and will merge this once the CI runs

PR #60, opened 2026-02-04
 Fix path traversal bug in attachments

hughdbrown:
The export-attachment command's --output flag passes user-supplied paths directly to os.OpenFile without validation. The command's own documentation shows a scripted workflow where email-derived attachment filenames are piped into -o:

  jq -r '.attachments[] | "\(.content_hash)\t\(.filename)"' | \
    while IFS=$'\t' read -r hash name; do
      msgvault export-attachment "$hash" -o "$name"
    done

  An email attachment named ../../.ssh/authorized_keys would write outside the working directory.

  The Fix (3 lines of logic)

  Added ValidateOutputPath() to internal/export/attachments.go — it cleans the path with filepath.Clean() and rejects relative paths that start with .. (meaning they escape the working directory after normalization). Absolute paths are allowed since they represent an explicit user choice.

  The validation is called early in runExportAttachment before any file I/O.

  Changes Made

  - internal/export/attachments.go: Added ValidateOutputPath() function (7 lines)
  - internal/export/attachments_test.go: Added TestValidateOutputPath with 8 test cases
  - cmd/msgvault/cmd/export_attachment.go: Added validation call before file operations (5 lines)

### DIFF
--- a/internal/export/attachments.go
+++ b/internal/export/attachments.go
@@ -212,6 +212,11 @@ func ValidateOutputPath(outputPath string) error {
 	if filepath.VolumeName(cleaned) != "" {
 		return fmt.Errorf("output path %q contains a drive or UNC prefix; use a relative path", outputPath)
 	}
+	// Reject rooted paths (leading / or \) which are drive-relative on Windows
+	// and absolute on Unix. filepath.IsAbs misses these on Windows.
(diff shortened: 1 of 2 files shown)


wesm:
I'm pushing some fixes from reviews on the other path traversal PR, and I rebased. Bear with me

Issue #14, opened 2026-02-02
Feature request: IMAP support (iCloud, Fastmail, etc.)


danielbrodie:
Hey Wes,

Really excited about msgvault — the architecture is exactly what I've been looking for. Local-first, DuckDB-powered search, MCP server built in. This is the dream.

The catch for me: I'm on iCloud Mail, not Gmail. I've been using Bichon for archiving, but it doesn't have any CLI or programmatic API — just a WebUI. The whole reason I want a local archive is to query it with agents, and msgvault nails that with the MCP server.

Would love to see IMAP support so folks on iCloud, Fastmail, ProtonMail (via Bridge), etc. can use this. I know the Gmail API gives you nice things like labels and the History API for incremental sync, but basic IMAP would cover a lot of ground for people who aren't in the Google ecosystem.

Happy to help test if you get something working — I've got about 180k emails in my iCloud archive.

Thanks for building this!

Darthagnon:
I use Yahoo, Gmail and Outlook and IMAP/POP3 support would be amazing! I don't really trust any of those providers to keep their proprietary APIs stable over time, but there's usually some hack to enable IMAP.

twleung:
Add another vote for IMAP support!

PR #93, opened 2026-02-06
Design for Yahoo Mail integration

hughdbrown:
PRD: docs/design/yahoo-mail-prd.md
detailed task breakdown: docs/design/yahoo-mail-tasks.md


wesm:
since there is an IMAP support PR #14 I'll evaluate that against this PRD

PR #111, opened 2026-02-09
Fix CPU pinning and UX issues in batch deletion

wesm:
## Summary

- **Fix CPU pinning during batch deletion**: `ExecuteBatch` was calling `MarkMessageDeletedByGmailID` 1000 times in a tight loop after each batch API call. Each call did a full table scan because `source_message_id` lacked a standalone index (only existed as the second column in `UNIQUE(source_id, source_message_id)`, unusable for `source_message_id`-only lookups). Fixed by batching DB writes with `IN (...)` clauses and adding a standalone index.
- **Fix Ctrl+C spamming hundreds of warnings**: The retry-failed-IDs loop and batch-fallback loop in `ExecuteBatch` had no `ctx.Done()` check, so cancellation caused every remaining ID to fail at the rate limiter and log a warning line.
- **Show resumed position on restart**: `OnStart` now accepts `alreadyProcessed` so the progress bar immediately shows the checkpointed position instead of `0/N`.
- **Add `InitSchema` to `delete-staged`**: Was the only command missing it, so the new index wouldn't be created for existing databases.

## Test plan

- [x] All existing tests pass (`make test`)
- [x] Run `delete-staged` on a large batch, Ctrl+C, rebuild, restart — verify progress jumps to resumed position and no warning spam on cancel

### DIFF
--- a/cmd/msgvault/cmd/deletions.go
+++ b/cmd/msgvault/cmd/deletions.go
@@ -505,14 +505,16 @@ func isTTY() bool {
 
 // CLIDeletionProgress reports deletion progress to the terminal.
 type CLIDeletionProgress struct {
-	total     int
-	startTime time.Time
-	lastPrint time.Time
-	tty       bool
+	total        int
+	resumeOffset int // messages already processed before this run
+	startTime    time.Time
+	lastPrint    time.Time
+	tty          bool
 }
 
 func (p *CLIDeletionProgress) OnStart(total, alreadyProcessed int) {
 	p.total = total
+	p.resumeOffset = alreadyProcessed
 	p.startTime = time.Now()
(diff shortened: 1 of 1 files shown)


hughdbrown:
The events I see are:
- one commit [7bb9e0b](https://github.com/wesm/msgvault/commit/7bb9e0b1ebcab4d0b71e04c9edbed4ce61613583)
- two roborev reviews with warnings
- merge of the pull request

Roborev identified a number of warnings:
- High: SQL Injection in Batch Deletion
- Medium: Path Traversal on Deletion Checkpoint
- a medium and a low that I don't quite get.

Does this mean that Roborev did these reviews and the branch was merged without fixing the identified warnings?

I'd just like to know how the project manages code reviews and whether they are considered blocking.

PR #124, opened 2026-02-11
Feature python pkg

hughdbrown:
Create a python package that exposes an object model for working with the msgvalut.

### DIFF
--- a/scripts/msgvault_sdk/README.md
+++ b/scripts/msgvault_sdk/README.md
@@ -1,0 +1,235 @@
+# msgvault_sdk
+
+Python SDK for programmatic access to [msgvault](https://github.com/hughdbrown/msgvault) email archives.
+
+Read, query, filter, and mutate your locally archived Gmail data from Python scripts or Jupyter notebooks.
+
+## Installation
+
+```bash
+# Install with uv (recommended)
+cd scripts/msgvault_sdk
+uv sync
+
+# With pandas support (for DataFrame export)
+uv sync --extra pandas
+
+# With pip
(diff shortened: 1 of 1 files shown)


hughdbrown:
> If the SDK interacted only with "public APIs" (the CLI and HTTP API) that would be a different story.

The spec and task list were generated from my design-skill using the text in docs/design/prompt.md. There is no mention in prompt.md that the python code should operate against the sqlite database directly. It seems plausible to me that if the prompt.md were modified to specify a particular implementation direction that was satisfactory, then the design skill could be re-run to create a new spec and task list, resulting in a different (but conforming) implementation.

wesm:
> It seems plausible to me that if the prompt.md were modified to specify a particular implementation direction that was satisfactory, then the design skill could be re-run to create a new spec and task list, resulting in a different (but conforming) implementation.

Sure, agreed. But now in this new world where code is now abundant, cheap, and easy to throw away and recreate, we have to be even more thoughtful about what code we create and put in users' hands. I think it's fine if you want to have your own Python library to interact with the sqlite database, but to me it is too much scope creep for this repository at least. If you want to create SDK library to interact with msgvault via the Go binary and its public APIs (which are currently only the CLI and HTTP API), we can talk about that and I think it probably should be a separate msgvault-sdk library.

wesm:
I'm going to close this since this feels like too much scope creep for a project that is as young as this one -- we have only scratched the surface with the core system and there is too much not-yet-done (for example, web UI + mobile search / retrieval experience) there. Feel free to pursue this as a separate project, though, and I'll be happy to link to it from a community projects section in the README

Issue #134, opened 2026-02-15
Unclear: Token has been expired or revoked.


veitsanner:
After synchronizing my account for about a week, I get the following error when executing `msgvault sync`:
```
Errors:
  ***Redacted***@gmail.com: get token source: refresh token: oauth2: "invalid_grant" "Token has been expired or revoked." (run 'add-account' first)
Error: 1 account(s) failed to sync
...
```
Executing `msgvault add-account  ***Redacted***@gmail.com` does not help. I still get the same error for `msgvault sync`.

Any advice on how to refresh the token?

wesm:
Delete the existing token at ~/.msgvault/tokens/$email.json and then add-account again (this has no affect on the existing database). This happened to me also and I haven't gotten around to automatic token refresh, but I will do that

VeitSanner:
Thanks, the workaround resolves this.

bchoor:
I found that using a token on an app in testing (for Google) will expire within 7 days. You can make it production, but it'll complain not being verified, you just have to click the extra checkbox when authenticating to get your token. I found that even with token refreshes, it'll expire - but haven't had any issues when using production app. 

reference:
https://support.google.com/cloud/answer/15549945?hl=en 

"Authorizations by a test user will expire seven days from the time of consent. If your OAuth client requests [an offline access type](https://developers.google.com/identity/protocols/oauth2/web-server#offline) and receives a refresh token, that token will also expire."

That said, I would recommend documentation that clearly explains the test vs production oauth tokens, as this is not a bug but how google oauth works by design.

rodboev:
The token-expiry behavior described here is handled on current main by PR #135, including invalid_grant detection and reauthorization tests. This issue can probably be closed.

PR #135, opened 2026-02-15
Handle expired/revoked OAuth tokens with auto-reauth and --force flag

hughdbrown:
## Summary

- Add `--force` flag to `add-account` for manual token deletion and re-authorization
- Auto-reauth expired tokens during `sync`, `sync-full`, `verify`, and `delete-staged` — detects `invalid_grant`, deletes the stale token, and re-runs the browser OAuth flow in interactive sessions
- Guard auto-reauth against transient errors (network, context cancellation) and non-interactive sessions (cron, CI)
- Reject `--headless --force` flag combination (--force requires browser OAuth)
- Test coverage for `isAuthInvalidError` (8 cases) and `getTokenSourceWithReauth` (9 cases) via extracted `tokenReauthorizer` interface

## Test plan

- [x] `go test -tags fts5 ./cmd/msgvault/cmd/... -run "TestIsAuthInvalidError|TestGetTokenSourceWithReauth" -v` — all 17 cases pass
- [x] `make test` — full suite passes
- [x] `make lint` — no new issues
- [ ] Manual: `add-account --force` deletes token and re-authorizes
- [ ] Manual: `sync` with expired token triggers auto-reauth in interactive terminal
- [ ] Manual: `sync` with expired token in non-interactive session returns actionable error

Fix for Issue #134

### DIFF
--- a/cmd/msgvault/cmd/addaccount.go
+++ b/cmd/msgvault/cmd/addaccount.go
@@ -23,7 +23,8 @@ By default, opens a browser for authorization. Use --headless to see instruction
 for authorizing on headless servers (Google does not support Gmail in device flow).
 
 If a token already exists, the command skips authorization. Use --force to delete
-the existing token and re-authorize (useful when a token has expired or been revoked).
+the existing token and start a fresh OAuth flow (most token issues are handled
+automatically during sync and verify).
 
 Examples:
   msgvault add-account you@gmail.com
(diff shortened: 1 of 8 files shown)


hughdbrown:
For my curiosity, how was the `--headless silently bypasses --force` review generated? I used Claude and got a clean bill of health:

```
## Summary

This commit improves the --force flag UX across the msgvault CLI by:

1. Restoring sync-full as the primary next-step when an account is already authorized
2. Adding an info message when --force is used but no token exists
3. Differentiating error messages based on whether a token file exists (missing token → suggest add-
account; invalid token → suggest add-account --force)

## Issues

No issues found.

The changes are clean, consistent across all four commands, and directly address the feedback from
the previous review. The HasToken branching logic is applied uniformly, and the fmt.Errorf format
strings correctly include the email parameter.
```

PR #193, opened 2026-03-13
Add IMAP account support via go-imap/v2

wesm:
## Summary

- Add IMAP account sync using `go-imap/v2`, supporting any standard IMAP server (implicit TLS, STARTTLS, or plain)
- New `add-imap` command for interactive account setup with connection testing
- Gmail IMAP dedup: detect `\All` special-use attribute (RFC 6154) to avoid duplicate imports across virtual folders, with label preservation via Message-ID mapping
- Route `sync` and `sync-full` commands across Gmail and IMAP sources using `GetSourcesByIdentifier` for correct multi-type resolution
- IMAP sources use `BODY.PEEK[]` to avoid marking messages as `\Seen`
- `remove-account` cleans up IMAP credential files
- Mixed Gmail+IMAP setups: skip Gmail sources when OAuth is absent without blocking IMAP sync; surface real OAuth errors (malformed secrets) instead of silently skipping
- Return clear error when syncing an identifier that exists only as a non-syncable type (e.g. mbox)
- Regression tests for identifier ambiguity, source routing, mbox fallback, and OAuth skip behavior

## Test plan

- [ ] `add-imap` against a Gmail IMAP account (app password) -- verify connection test, credential storage, source creation
- [ ] `sync-full` for IMAP account -- verify message download with dedup (All Mail detection)
- [ ] `sync-full` with no args in mixed Gmail+IMAP setup without OAuth configured -- verify Gmail skipped, IMAP synced
- [ ] `sync-full <identifier>` where identifier is mbox-only -- verify clear error
- [ ] `remove-account` for IMAP source -- verify credential file removed
- [ ] `go test ./...` passes

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/imap/client.go
+++ b/internal/imap/client.go
@@ -850,13 +850,21 @@ func (c *Client) TrashMessage(ctx context.Context, messageID string) error {
 }
 
 // DeleteMessage permanently deletes a message using UID STORE \Deleted
-// followed by UID EXPUNGE (UIDPLUS) or plain EXPUNGE as a fallback.
+// + UID EXPUNGE. Requires the UIDPLUS extension (RFC 4315); without it
+// plain EXPUNGE would remove every \Deleted message in the mailbox,
+// not just the target.
 func (c *Client) DeleteMessage(ctx context.Context, messageID string) error {
 	mailbox, uid, err := parseCompositeID(messageID)
 	if err != nil {
 		return err
 	}
 	return c.withConn(ctx, func(conn *imapclient.Client) error {
+		if !conn.Caps().Has(imap.CapUIDPlus) {
+			return fmt.Errorf(
+				"server does not support UIDPLUS; " +
+					"permanent delete requires UID EXPUNGE " +
+					"(use trash instead)")
+		}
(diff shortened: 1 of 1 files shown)


wesm:
> Combined Review (518feaf) — Cross-sync dedup and EML path traversal

Both findings addressed in commit 2f02664:

**Finding 1 (cross-sync dedup):** Added `rfc822_message_id` column to the messages table, populated from the parsed MIME Message-ID during ingest. For IMAP sources, `MessageExistsByRFC822ID` checks for existing messages with the same RFC822 Message-ID before persisting. Duplicates from messages moving between mailboxes across syncs are caught and counted as skipped via `errDuplicateRFC822` sentinel. Regression test `TestIMAPCrossSyncDedup` verifies a message imported from `AllMail|42` is not re-imported when it reappears as `Trash|99`.

**Finding 2 (EML path traversal):** `sanitizeEMLFilename` already replaced `/` and `\` with `_`, but now also applies `filepath.Base` to ensure the result is a plain filename with no directory components.

PR #194, opened 2026-03-14
Fix Apple Mail V10 import and auto-discover accounts

### LINKED ISSUE #157: import-emlx not importing from Apple Mail Library (~/Library/Mail)
Such a great tool. Unfortunately it does not seem to look recursively through the Apple Mail directories as advertised with "import-emlx":

> msgvault import-emlx xyz.com ~/Library/Mail
> Import complete.
>   Mailboxes:      0 discovered, 0 imported
>   Processed:      0 messages
>   Added:          0 messages
>   Updated:        0 messages
>   Skipped (dup):  0 messages
>   Errors:         0

same for 

> msgvault import-emlx xyz.com ~/Library/Mail/V10

or any other path.

Full file access is granted. Other (mbox) imports run flawlessly.

wesm:
## Summary

- Discover `.emlx` files in V10 numeric partition subdirectories (`Data/0/3/Messages/`, `Data/9/Messages/`, etc.) at arbitrary nesting depths, fixing the 99% message loss reported in #157
- Handle partition-only mailboxes where `Data/Messages/` does not exist at all
- Resolve partition file paths transparently via `FileIndex` map + `FilePath()` method
- Add `internal/applemail` package that reads `~/Library/Accounts/Accounts4.sqlite` to map V10 directory GUIDs to email addresses
- Rework `import-emlx` command to support auto-discovery — `import-emlx [mail-dir]` discovers all accounts automatically
- Add `--accounts-db`, `--account` (repeatable filter), and `--identifier` (manual fallback) flags
- Restrict V10 discovery to the newest `V*` directory to avoid importing stale data after macOS upgrades
- Preserve backward compatibility with the legacy `import-emlx <identifier> <mail-dir>` two-arg form

Closes #157. Supersedes #166, #172, #187.

## Test plan

- [x] Unit tests for partition discovery (mixed, partition-only, no-top-level-Messages)
- [x] Unit tests for `Accounts4.sqlite` resolution (IMAP child→parent, Exchange, On My Mac, missing GUID)
- [x] Unit tests for `newestVDir` (V2+V10 picks V10, single V dir, no V dirs)
- [x] `TestFindV10GUIDs` verifies stale V2 GUIDs are excluded
- [ ] Verify against real Apple Mail archive on macOS Ventura+

### DIFF
--- a/internal/importer/emlx_import.go
+++ b/internal/importer/emlx_import.go
@@ -577,21 +577,17 @@ func ImportEmlxDir(
 
 // resolveCheckpointFile finds the full path in files whose basename
 // matches the legacy bare filename from an older checkpoint. Returns
-// the last matching full path (since files are sorted, this is the
-// one the checkpoint would have referred to). Returns an error if no
-// match is found, since comparing a bare filename against absolute
-// paths would produce incorrect resume behavior.
+// the first matching full path so that no previously-unseen partition
+// files are skipped. Returns an error if no match is found, since
+// comparing a bare filename against absolute paths would produce
+// incorrect resume behavior.
 func resolveCheckpointFile(
 	basename string, files []string,
 ) (string, error) {
-	var match string
 	for _, f := range files {
 		if filepath.Base(f) == basename {
-			match = f
+			return f, nil
 		}
-	}
-	if match != "" {
-		return match, nil
 	}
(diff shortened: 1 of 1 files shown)


wesm:
This is good enough. merging

PR #217, opened 2026-03-23
Fix IMAP IPv6 host parsing in add-imap

endolith:
**_This was written by Cursor.  I don't know Go at all, so I am not claiming it's correct or not._**

## What this fixes

`add-imap` failed with IPv6 hosts passed to `--host` due to invalid address construction (`::1:1993`), causing:

- `dial tcp: address ::1:1993: too many colons in address`

## Root cause

IMAP config built `host:port` with string formatting instead of proper host/port joining, so raw IPv6 literals were not bracketed.

## Change

- Use Go's `net.JoinHostPort` when building IMAP addresses/identifiers.
- Normalize optional IPv6 brackets in `--host` input so both forms are accepted:
  - `::1`
  - `[::1]`

## Tests

Added/updated unit tests to verify:

- IPv6 unbracketed host works.
- IPv6 bracketed host works.
- Both normalize to the same valid output.

### DIFF
--- a/internal/imap/config.go
+++ b/internal/imap/config.go
@@ -4,8 +4,10 @@ package imap
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/url"
 	"strconv"
+	"strings"
 )
 
 // Config holds connection settings for an IMAP server.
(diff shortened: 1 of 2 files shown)


wesm:
I don't have the ability to test easily but it looks innocuous enough.

PR #252, opened 2026-04-09
Enhance search functionality with regex, FTS5, snippets, and sorting

jesserobbins:
Fix and Improve search accuracy and performance across full-text, fast (Parquet), and aggregate query paths with word-boundary matching, better snippet handling, and consistent result sorting.

###   Problem

While importing my old 1990's era email archives, I discovered a number of issues with search

  1. Substring matching causes false positives — Text search matches partial strings within words, returning irrelevant results
  2. Body text invisible in fast search — Messages where search terms only appear in the body don't show in TUI unless user switches to deep search mode (which I feel is underdocumented but that's a separate issue) 
  3. Inconsistent sort behavior — Search results don't respect the user's selected sort field (Name/Count/Size)

###  Proposed Solution

  - Word-boundary regex (\b) for text search to match exact words, not substrings
  - Prefix matching in FTS5 (* operator) for partial word matches within the deep search path
  - Snippet inclusion in fast search conditions so body text matches appear immediately in TUI

###   Changes

  - internal/query/sqlite.go — FTS5 prefix matching
  - internal/query/duckdb.go — Word-boundary regex,
  snippet inclusion, sort parameter, escapeRegex helper

  Testing

  - Text search with word boundaries prevents false
  substring matches
  - Parquet fast search now catches body text matches

### DIFF
--- a/internal/query/duckdb.go
+++ b/internal/query/duckdb.go
@@ -418,6 +418,19 @@ func escapeILIKE(s string) string {
 	return s
 }
 
+// startsWithWordChar reports whether s begins with a regex word character
+// [a-zA-Z0-9_]. Used to decide whether \b is appropriate as a prefix.
+func startsWithWordChar(s string) bool {
+	if len(s) == 0 {
+		return false
+	}
+	c := s[0]
+	return (c >= 'a' && c <= 'z') ||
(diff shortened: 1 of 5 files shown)


jesserobbins:
Removed the incomplete sort stuff

PR #256, opened 2026-04-10
Fix search performance regression and TUI race condition

jesserobbins:
Tested and ran into a bunch of performance & context issues that surfaced in the TUI. Original approach using ILIKE is better it seems like for large archives.

## Changes

- **Revert DuckDB fast search from `regexp_matches` back to `ILIKE`** — on large Parquet archives the regex-per-row evaluation caused TUI fast search to hang. ILIKE is natively optimized by DuckDB for columnar Parquet scans. FTS5 deep search path is unchanged.
- **Fix TUI search race condition** — when starting an inline search, invalidate in-flight `loadMessages` responses (from sort toggle or "all messages" view) so they can't overwrite search results after the search completes.
- **Clear `m.messages` immediately on search start** — prevents stale pre-search messages from showing during the async search transition.
- **Remove dead code** — `escapeRegex` and `startsWithWordChar` are no longer used.
- **Add regression tests** — 4 new tests covering the race condition, stale-message clearing, and ILIKE path verification for both fast and aggregate search.

### DIFF
--- a/internal/query/duckdb.go
+++ b/internal/query/duckdb.go
@@ -470,32 +470,25 @@ func (e *DuckDBEngine) buildAggregateSearchConditions(searchQuery string, keyCol
 
 	// Text terms: always search subject + sender, plus the view's grouping
 	// key columns when provided (e.g., label name in Labels view).
-	// Uses word-boundary regex (\b) for terms starting with word chars.
-	// Terms starting with non-word chars (e.g., +, @, #) skip \b
-	// since it requires a word/non-word transition that fails at
-	// string start or after whitespace.
+	// Uses ILIKE for performance on Parquet scans.
 	for _, term := range q.TextTerms {
-		escaped := escapeRegex(term)
-		regexPattern := "(?i)" + escaped
-		if startsWithWordChar(term) {
-			regexPattern = "(?i)\\b" + escaped
-		}
+		termPattern := "%" + escapeILIKE(term) + "%"
 		var parts []string
-		parts = append(parts, `regexp_matches(COALESCE(msg.subject, ''), ?)`)
-		args = append(args, regexPattern)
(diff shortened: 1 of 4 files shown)


jesserobbins:
Missed a chance to make a joke that ILIKE Ducks or something but will keep the PR

jesserobbins:
> 
> ## roborev: Combined Review (`d3811db`)
> 
> Medium issue found; security review and one default review found no additional issues.
> 
> ## Medium
> 
> - `internal/query/duckdb.go:447` - `buildAggregateSearchConditions` now emits 4 arguments per text term with no key columns, but `countArgsForTextTerms` still skips only 3. `buildStatsSearchConditions` uses that helper when a stats search has both text terms and non-text filters, so it can append leftover text-term args without matching placeholders and fail aggregate stats queries.
> 
>   **Suggested fix:** Update `countArgsForTextTerms` to return `n * 4` and add a regression test for a stats/aggregate query combining a text term with a non-text operator such as `from:`.
> 
> ---
> *Synthesized from 3 reviews (agents: codex, gemini | types: default, security)*
> 

Interesting will look and fix

jesserobbins:
Hi! I disagree with these recommendations which is why I am proposing reverting to the prior search behavior in this PR.

1) A user initiating a search expects to see only the results of that search. Retaining and presenting cached and irrelevant search is confusing and counterproductive. Additionally the TUI is inconsistent in how it updates and displays this as results come in. The TUI should display the search response or nothing. 

2) I agree that the regex based deep search is better. However it is unacceptably slower in my own testing. 

So recommend proceeding with the current solution, and in future make TUI search flows use same path as the search function at the cli. 

Thanks!

jesserobbins:
(Also @wesm roborev is pretty cool!)

PR #262, opened 2026-04-13
 Fix search performance regression and TUI race condition

jesserobbins:
Tested and ran into a bunch of performance & context issues that surfaced in the TUI. Original approach using ILIKE is better it seems like for large archives. (Note this is a resubmission of #256) 

Changes

- Reverts DuckDB text search from regexp_matches to ILIKE — regexp caused significant performance regression on Parquet scans. ILIKE is natively optimized by DuckDB for columnar data. FTS5 deep search path is unchanged.
- Fixes a TUI race condition where pressing a (all messages) then quickly searching could cause the stale loadMessages response to overwrite search results. Search now invalidates pending loads via loadRequestID.
- Fixes a placeholder/arg count drift in buildStatsSearchConditions — text terms emit 4 args each (subject + snippet + 2 sender) but the arg-slicing helper assumed 3. Extracted buildNonTextSearchConditions to eliminate the brittle offset math entirely. Regression tests lock the placeholder invariant across both search builders.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ LDFLAGS := -X github.com/wesm/msgvault/cmd/msgvault/cmd.Version=$(VERSION) \
 
 LDFLAGS_RELEASE := $(LDFLAGS) -s -w
 
-.PHONY: build build-release install clean test test-v fmt lint lint-ci tidy shootout run-shootout install-hooks help
+.PHONY: build build-release install clean test test-v fmt lint lint-ci tidy shootout run-shootout install-hooks bench help
 
 # Build the binary (debug)
 build:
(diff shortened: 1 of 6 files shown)


jesserobbins:
> ## roborev: Combined Review (`bf776fb`)
> Verdict: Changes are mostly sound, but there is one medium-severity UI regression to address before merging.
> 
> ## Medium
> * **False "No results found" state during inline search**
>   
>   * **Location:** `internal/tui/model.go:1194` (affects `internal/tui/view.go`)
>   * **Finding:** Clearing `m.messages = nil` when a search starts can cause the view to briefly render a misleading empty state. The empty-state logic shows "No results found" when `len(m.messages) == 0 && !m.loading`, but debounced inline search sets `m.inlineSearchLoading = true` while leaving `m.loading = false`. That means users can see a false no-results message while the inline search request is still in flight.
>   * **Suggested fix:** In `internal/tui/view.go`, suppress the empty-state message while inline search is loading, e.g. require `!m.inlineSearchLoading` in the condition:
>     ```go
>     if len(m.messages) == 0 && !m.loading && !m.inlineSearchLoading
>     ```
> 

I still disagree with this recommendation. It is better to show nothing than stale search. I discussed this in #256

wesm:
Confirmed. Adding a minimal benchmark so we can keep an eye on this

Issue #129, opened 2026-02-13
Improve file logging for sync workloads


wesm:
Certain classes of warnings or errors are not displayed in the console by default — we should log these to a log file in MSGVAULT_HOME so they can be investigated after the fact

rodboev:
The structured file logging and sync warning handling described here landed in PR #263 and remain on current main. This issue can probably be closed.

PR #263, opened 2026-04-13
Add structured file logging with per-run correlation

jesserobbins:
I had already added logging for my own work on the deduplicate and fixup features I've been working on, and so am proposing durable structured logging to address #129 and likely other development work. @wesm does this do what you want?

## What changed

- **`internal/logging`**: new package — fans log records to two sinks: human-readable text on stderr and a rotating JSON file under `<data dir>/logs/`. Every process gets a short `run_id` hex tag on every record for cross-invocation grep.
- **`msgvault logs`**: new command to tail, filter, and follow the on-disk log files (`-f`, `--run-id`, `--level`, `--grep`, `--all`, `--path`).
- **`root.go`**: installs the handler at startup; writes structured `msgvault startup` / `msgvault exit` lines; recovers panics to disk; new flags `--log-file`, `--log-level`, `--no-log-file`; new `[log]` config section.
- **Operations**: search, stats, and list-accounts now emit structured start/done/fail lines with `duration_ms`.
- **TUI**: `loadData`, `loadMessages`, and search operations emit structured lines; stderr logging is suppressed during alt-screen to prevent render corruption.
- **Store**: new `--log-sql` and `--log-sql-slow-ms` flags; slow queries (>100 ms by default) emit WARN; every query optionally traced at INFO.

## Why

Warnings and errors during sync workloads were silently dropped. The log file gives operators a reliable post-hoc record of what msgvault did and why it failed, without requiring stderr redirection.

### DIFF
--- a/cmd/msgvault/cmd/list_accounts.go
+++ b/cmd/msgvault/cmd/list_accounts.go
@@ -113,6 +113,10 @@ func listLocalAccounts() error {
 		}
 	}
 
+	logger.Info("list-accounts",
+		"sources", len(stats),
+	)
+
 	if listAccountsJSON {
 		return outputAccountsJSON(stats)
 	}
--- a/cmd/msgvault/cmd/logs.go
+++ b/cmd/msgvault/cmd/logs.go
@@ -131,8 +131,28 @@ func findLogFiles(dir string, all bool) ([]string, error) {
 		}
 		files = append(files, filepath.Join(dir, n))
 	}
-	sort.Strings(files)
+	sort.Slice(files, func(i, j int) bool {
+		return logFileSortKey(files[i]) < logFileSortKey(files[j])
+	})
 	return files, nil
+}
+
+// logFileSortKey returns a string that sorts log files chronologically:
+// rotated files (.log.1, .log.2) come before the active .log for the
+// same date, and higher rotation indices (older) sort earlier.
+func logFileSortKey(path string) string {
+	name := filepath.Base(path)
+	// msgvault-2026-04-11.log   -> date=2026-04-11 suffix=999 (active, last)
+	// msgvault-2026-04-11.log.1 -> date=2026-04-11 suffix=001
+	// msgvault-2026-04-11.log.2 -> date=2026-04-11 suffix=002
+	if idx := strings.LastIndex(name, ".log."); idx >= 0 {
+		date := name[:idx+4] // through ".log"
+		num := name[idx+5:]
+		return date + fmt.Sprintf(".%03s", num)
+	}
+	// Active file (no rotation suffix) sorts after all rotations
+	// for the same date.
+	return name + ".999"
 }
 
 // logFilter represents the user's --run-id / --level / --grep
@@ -228,6 +248,7 @@ func followLogFile(
 	}
 
 	reader := bufio.NewReader(f)
+	var partial []byte
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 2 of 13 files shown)


jesserobbins:
> ## roborev: Combined Review (`cf80b82`)
> PR is not ready to merge: 3 medium-severity issues remain.
> 
> ## Medium
> 1. **Persistent exit logging is dropped on successful runs**
>    **Location:** `cmd/msgvault/cmd/root.go` (`PersistentPostRunE`; `ExecuteContext`), `cmd/msgvault/cmd/root.go:126-128`
>    `PersistentPostRunE` closes `logResult` before `ExecuteContext` emits the `"msgvault exit"` record. That means successful executions can miss the per-run outcome line in the structured log because the file sink has already been closed.
>    **Fix:** Emit the exit record before closing handlers, or move shutdown out of `PersistentPostRunE` so it happens only after the exit log is written.
> 2. **Structured logs persist sensitive user input and account identifiers**
>    **Location:** `cmd/msgvault/cmd/root.go:126-128`, `cmd/msgvault/cmd/search.go:157-180`, `internal/tui/model.go:340-382`, `internal/tui/model.go:548-568`
>    The new logging path records raw search text and account identifiers to disk. `msgvault startup` logs positional args verbatim, and the search/TUI logging duplicates that data via fields like `query`, `search`, `scope`, and `account`. Search strings can contain PII, ticket IDs, internal names, email addresses, or secrets, making the persistent log and `msgvault logs` a disclosure surface.
>    **Fix:** Treat free-form input as sensitive by default. Log coarse metadata only in normal operation, and require explicit debug opt-in for any query/account logging with redaction or hashing.
> 3. **`[log].dir` path handling is inconsistent with other config paths**
>    **Location:** `internal/config/config.go` (new `LogConfig` / `LogsDir()` path handling)
>    `log.dir` is returned verbatim; `Load()` does not expand `~` or resolve relative paths against the config/home directory. With an explicit config file, `dir = "logs"` writes under the current working directory instead of alongside the config/home directory, unlike the other path-based settings.
>    **Fix:** Run `expandPath` on `cfg.Log.Dir`, and resolve relative `log.dir` against `cfg.HomeDir` when `--config` is explicitly provided.
> 

Given that a user accessing the logs have access to the actual databases... this feels overly cautious. However, it's not a bad standard to uphold so I'm taking a pass at it.

jesserobbins:
> ## roborev: Combined Review (`fb87bef`)
> **Verdict: Changes are not ready to merge; there is 1 high-severity issue and 2 medium-severity issues to address.**
> 
> ## High
> * **Panic logging can be lost because the log file is closed before panic recovery runs**
>   **Location:** `cmd/msgvault/cmd/root.go:223`
>   In `ExecuteContext`, the `defer` order is reversed for the intended behavior: `logResult.Close()` runs before `recoverAndLogPanic()`. If the command panics, the file handler may already be closed before the panic is written, which can silently drop the on-disk panic record.
>   **Fix:** Defer panic recovery after the close is registered so it executes first, or otherwise ensure the file stays open until panic logging completes.

I have not deeply examined the panic logging handling here. However, I don't think this scenario represents a "high severity issue" given that we are *adding* missing logging rather than somehow improving an existing function. Regardless, I'm open to the improvement in this scenario. 

 
> ## Medium
> * **Persistent JSON file logging has no redaction layer for sensitive fields**
>   **Location:** `cmd/msgvault/cmd/root.go:90`, `internal/logging/logging.go:153`
>   The new default logger now persists records to disk for most command runs, but there is no central redaction/suppression step for the file sink. Existing log fields that were previously only transient on `stderr` can now be stored at rest, including potentially sensitive user-derived values such as search queries, email addresses, and filenames. `0600` helps, but it does not remove the underlying data exposure risk.
>   **Fix:** Add redaction for the file sink, or make durable logging opt-in. At minimum, suppress or hash clearly sensitive user/content-derived fields before they reach the JSON log handler.

This is turning into a feedback loop. I am going to set logging as being opt-in to address this feedback. Redaction approaches are always tricky, and I think what most people are looking for here is to understand what is happening. This is not a robust external facing service. This is a personal tool. 

> * **Rotated log files are emitted in the wrong chronological order**
>   **Location:** `cmd/msgvault/cmd/logs.go:134`
>   `logFileSortKey` sorts rotated files by ascending suffix (`.1`, `.2`, ...), but higher suffixes represent older files. That means `msgvault logs --all` can print same-day rotated chunks out of order, showing newer entries before older ones.
>   **Fix:** Sort by date first, then sort rotation suffix in descending numeric order, with the active `.log` file last.

Fine. Although if the next review comment is timezone related I am going to just make them UUIDs. :-)

jesserobbins:
I also manually re-verified logging. I'm going to mark as ready.

wesm:
thanks!

PR #276, opened 2026-04-17
PostgreSQL dialect refactor - PR1 of 4 - Dialect

webgress:
Adding PostgreSQL as an opt-in database backend (SQLite stays the default).
Details in `pg_refactor_docs/`.
This is split into 4 PRs to make it easier to review:
- **PR1 — Dialect extraction** (this PR): Introduces a `Dialect` interface and `SQLiteDialect`, moves all SQLite-specific SQL out of the store layer. No new dependencies, no behavior change.
- **PR2 — PostgreSQL dialect**: Adds `PostgreSQLDialect`, pgx driver, PG schema with tsvector FTS, and dual-backend test support.
- **PR3 — PostgreSQL functional**: Portable query layer (`Rebind`, `RETURNING`, parameterized FTS), PG-native schema migrations, connection pool config, and full test coverage across both backends.
- **PR4 — Data migration**: `migrate-db` command for copying data between SQLite and PostgreSQL in either direction.


wesm:
Just rebased and addressed some review findings

PR #286, opened 2026-04-22
Identities, Collections, and Deduplication

### LINKED ISSUE #278: Feature: Identities, Collections, and Deduplication
Our communications are scattered across decades of email accounts, chat apps, phone backups, and meeting recordings. msgvault finally lets us consolidate that into a single, safe, searchable place. Our varied collections finally consolidated across  Gmail API syncs, old mbox exports, Apple Mail folders from a retired laptop, IMAP backups, Facebook Messenger dumps, SMS exports, WhatsApp histories, meeting recordings, call logs.

If you are like me, you spend a weekend pulling all the distributed archives together and importing them. Then you start searching in the TUI, with the MCP, or directly. The sense of joy turns to something else as you see the same email appears three times because it arrived via Gmail sync, an old mbox export, and an Apple Mail import. Message counts are wrong. Search returns dozens of copies sometimes. There's no way to tell what was sent vs received in an old mbox dump that lost its metadata. And every import sits in its own silo, with no unified view of "all my communications."

This new feature is a pretty big swing at expanding on this great foundation. 

> **Updated 2026-04-29 with the aligned design.** This proposal was first opened with a draft model and a working branch (PR #286). After @wesm's [design review](https://github.com/wesm/msgvault/pull/286#issuecomment-4320039075), I rewrote the proposal end-to-end rather than patching it. The full aligned design is below — this issue is now the canonical statement of the model. PR #286 will be reshaped against it. I realize this is more of an Epic but these kinds of things often are. ;-)

# Introducing: Accounts, Collections, Identities, and Deduplication

## Accounts, identities, and collections

This feature introduces three related concepts that make the archive usable without erasing provenance or changing the underlying data structure. 

### Accounts

An **account** is one imported message source/archive. A Gmail sync is an account. An mbox import is an account. An Apple Mail import is an account.

Even if two accounts represent the same real-world mailbox, msgvault keeps them separate until the user groups them. That preserves the boundary between independent archives and avoids guessing that two sources should be merged just because their data overlaps.

### Identities

An **identity** identifies the owner of an account: the set of addresses, phone numbers, or provider identifiers that mean "me" inside that account. An account has one identity, which may contain many identifiers.

Identities are stored per account because the meaning of an address depends on the source. A collection's identity is derived from its member accounts. This lets msgvault recover sent/received meaning in old imports and make safer deduplication choices without applying one global identity list to every archive.

### Collections

A **collection** is the user's explicit grouping of accounts. `All` is the default collection containing every account, and users can create collections such as `work`, `personal`, or `old laptop imports`.

Collections are how msgvault offers a unified view across sources while keeping source provenance intact. Any operation that crosses account boundaries does so because the user selected a collection.

The name "collection" matters. An earlier prototype used "merged account," which only described the moment of assembly, not the thing the user actually has afterward. Users don't think "I merged accounts." They think "this is all my work email" or "this is everything." A collection describes what the user has, not how they assembled it. It works at every scale — `All` is a collection, `work` is a collection — and it doesn't overload the word "account," which already means one ingest source.

A future revision may introduce other identity or collection types (for example, device identities or saved-query collections). If that happens, renaming these objects to `account_identity` and `account_collection` would disambiguate. For now, `identity` and `collection` are unambiguous.

Speaking of unambiguous... here's a diagram that hopefully makes this clear. _(this is a real thing that I spent time on, not slop. I rejected a bunch of helpful attempts at mermaid charts. :-)_ 

Image

## Deduplication

**Deduplication** works inside those boundaries. `deduplicate --account` cleans up repeated rows within one source. `deduplicate --collection` compares messages across the accounts the user deliberately grouped. `deduplicate` with no scope processes each account independently — never as one global pool.

Applying dedup hides redundant local copies from normal search, browse, stats, API, MCP, and vector/hybrid retrieval paths. It does not destroy data: hidden copies stay on disk and `--undo <batch-id>` restores them. Hard-deleting locally and deleting from a source server are separate, explicit steps the user must take on top of dedup; the system never escalates from one to the next on its own.

The result is a model that supports both safety and usefulness: users can search and analyze a unified communication archive, but msgvault still knows which source each message came from, which identities apply to which account, and when an operation is crossing account boundaries.

Image

## Proposed Resolutions from Design Review

| Category               | Item                                                                                                                                                       | Proposed Resolution                                                                                                                                |
| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
| Account meaning        | Wes (PR #286 review): "**Account**: one ingest source/archive."                                                                                            | Define an account as one source/archive created by one ingest path; never as a logical mailbox spanning sources.                                   |
| Collection meaning     | Issue #278: "A collection is a named grouping of accounts." Wes: "**Collection**: a named grouping of accounts/sources."                                   | Define a collection as a named grouping of account/source IDs, and use it as the only cross-account grouping primitive.                            |
| Default collection     | Issue #278: "create an 'All' collection ... automatically includes every account." Wes: "**All**: the default collection containing every account/source." | Seed`All` by default with every account/source and treat it as a collection, not an account.                                                       |
| Dedup safety boundary  | Issue #278: "Dedup only operates within the boundary the user specifies." Wes: "Dedup across accounts should require an explicit collection boundary."     | Confine within-account dedup to that source, and require an explicit collection scope for any cross-account dedup.                                 |
| Unscoped dedup         | Issue #278: "`msgvault deduplicate` (no flags) — scans each account independently." Wes: "`msgvault deduplicate`: each account/source independently."     | Iterate per-account when no scope is given; never treat unscoped dedup as "dedup everything together."                                             |
| CLI scope clarity      | Wes: "keep`--account` restricted to one account/source, and add an explicit `--collection` flag."                                                          | Encode the boundary in the flag itself:`--account` resolves one source, `--collection` resolves a named group.                                     |
| Name shadowing         | Wes: "`--account work` may target a collection named `work`, not an account/source."                                                                       | Reject collection names passed through`--account`, and emit an explicit error with the correct flag hint.                                          |
| Nested collections     | Wes: collection creation can "effectively allow nested collection references."                                                                             | Restrict collection membership to accounts/sources only; do not support nested collections in this model.                                          |
| Identity scope         | Issue #278: "Identities are tied to accounts." Wes: global`[identity]` config is "a different model."                                                      | Store identities as per-account records, and treat any global identity list as legacy input rather than an active scope.                           |
| Collection identity    | Issue #278: "A collection's identity is the union of its accounts' identities."                                                                            | Derive collection identity from member accounts; do not configure a collection's identity set separately.                                          |
| Query hiding contract  | Issue #278: "Pruned copies are soft-deleted and hidden from all query paths." Wes: vector/hybrid paths can still "surface pruned duplicates."              | Treat dedup-hidden rows as hidden on every normal read surface, as a product contract rather than a query optimization.                            |
| Live-message predicate | Wes: "Centralize live-message filtering" across storage and retrieval surfaces.                                                                            | Define one live-message rule and apply it consistently across SQLite, DuckDB, FTS, vector, API, MCP, TUI, and stats.                               |
| Collection query scope | Wes: decide whether collections are "only a dedup/admin concept" or "first-class query scopes."                                                            | Promote collections to first-class user scopes for search, browse, stats, and dedup; reject any partial implementation that scopes only dedup.     |
| Cache/index policy     | Wes: "Cache/index invalidation needs a clearer policy."                                                                                                    | Anchor correctness to live-message filtering; treat rebuilds as storage/performance hygiene, never as a prerequisite for hiding pruned duplicates. |
| Schema ownership       | Wes: permanent collections "probably belong in the canonical schema/migration path."                                                                       | Place accounts, collections, identities, and dedup metadata in canonical schema and migrations as core data model concepts.                        |
| Undo semantics         | Wes: "Undo is not a full rollback."                                                                                                                        | Restore local visibility and cancel pending deletion intent where possible; do not guarantee an exact pre-run database state.                      |
| Remote deletion scope  | Wes: manifest naming and reporting need to remain "source/account-specific rather than collection-specific."                                               | Keep remote deletion source-scoped even when duplicate detection used a collection boundary.                                                       |

# The Core model

### Account

An account is one ingest source/archive. It is the smallest durable provenance unit in msgvault.

Examples:

- one Gmail sync source
- one IMAP source
- one mbox import
- one Apple Mail import
- one iMessage import
- one SMS import
- one Facebook Messenger import
- one meeting transcript import source

The same real-world mailbox imported through Gmail sync and later through an old mbox export creates two accounts. They may represent the same human mailbox, but they are distinct archives with distinct provenance and source-specific deletion semantics.

This keeps the data model honest: msgvault does not infer that two imports belong together just because an email address, display name, or message content overlaps.

### Collection

A collection is a named grouping of accounts. It is the user's explicit statement that multiple sources should be viewed or operated on together.

Examples:

- `All`
- `work`
- `personal`
- `old laptop imports`
- `gmail plus exports`
- `family messages`

Collections are many-to-many:

- An account can belong to multiple collections.
- A collection can contain multiple accounts.
- A collection contains account/source IDs, not other collections.

Collections are the boundary for cross-account features. If two independent archives should be searched, counted, deduplicated, or exported together, the user expresses that by putting them in a collection.

### All

`All` is the default collection containing every account/source.

`All` gives users a natural unified view without collapsing account provenance. It is still a collection. Operations against `All` are collection-scoped operations and should be displayed that way.

## Scope semantics

The user-facing scope vocabulary is deliberately small:

| Scope            | Meaning                                                 |
| ------------------ | --------------------------------------------------------- |
| Account scope    | One source/archive.                                     |
| Collection scope | All member accounts of one collection.                  |
| All scope        | The default collection containing every account/source. |

CLI flags should expose those boundaries directly:

| Command shape               | Meaning                                                                                                  |
| ----------------------------- | ---------------------------------------------------------------------------------------------------------- |
| `--account <account>`       | Resolve exactly one account/source.                                                                      |
| `--collection <collection>` | Resolve exactly one collection.                                                                          |
| no flag where supported     | Use the command's documented default, such as per-account iteration for dedup or`All` for search/browse. |

`--account` and `--collection` are mutually exclusive. A generic internal `Scope` type is useful, but generic user-facing flags are not. Users should be able to tell when they are crossing source boundaries by reading the command.

Name conflicts should fail clearly:

- If `work` is a collection, `--account work` fails and suggests `--collection work`.
- If `alice@example.com` is an account, `--collection alice@example.com` fails and suggests `--account alice@example.com`.
- If the same string exists as both an account display name and a collection name, the user must choose the correct flag and may need a more specific account identifier.

## Identity model

Identity answers: "Who am I in this source?"

An identity belongs to an account/source. It can include email addresses, phone numbers, or other protocol-specific identifiers. A confirmed identity means that messages from that address or identifier can be treated as "from me" within that account's context.

Identity is account-scoped for two reasons:

- The same address may appear in multiple imports, and that is expected.
- An address that is safe to treat as "me" in one account may be misleading in another account or shared archive.

A collection's identity is derived from its member accounts. It is the union of confirmed identities from those accounts, used only within the collection's scope.

Identity discovery should be evidence-based and reviewable. Candidate signals include:

- `is_from_me` metadata from ingest
- sent-folder or sent-label evidence
- account/source identifier match
- OAuth or provider account metadata
- user confirmation

Global identity configuration is not part of the target model.

If an `[identity]` block exists in `config.toml` from an older msgvault version, the first startup after upgrade migrates its addresses into per-account confirmed identities for every existing account, logs a warning naming the migration, and prints a one-time CLI notice asking the user to review per-account identities (`msgvault list-identities`). The global config block is then no longer read. This preserves the old behavior across the upgrade while moving every account onto the per-account model.

After migration, identity is stored per account. The global config block is legacy input only.

## Collection behavior

Collections are a primary user concept, not just a dedup helper.

Required behavior:

- `All` is created and maintained automatically.
- Users can create named collections from accounts.
- Users can add and remove accounts from collections.
- Collection membership accepts only accounts/sources.
- Collection views preserve account provenance.

Out of scope for the core model:

- Nested collections.
- Implicit collection creation based on matching email addresses.
- Treating a collection as an account.

Collection names and account identifiers can share human-friendly names, so the CLI and UI must preserve the distinction visually and behaviorally.

## Deduplication model

Deduplication removes redundant local copies from normal user-facing results without destroying the underlying archive by default.

### Valid dedup scopes

| Invocation                              | Boundary                                                    |
| ----------------------------------------- | ------------------------------------------------------------- |
| `deduplicate --account <account>`       | Compare messages only within that account/source.           |
| `deduplicate --collection <collection>` | Compare messages across member accounts in that collection. |
| `deduplicate`                           | Process each account independently.                         |

The unscoped form is a convenience for per-account cleanup. It must not compare all messages across all accounts as one global set.

The unscoped default is per-account iteration rather than `--collection All` because cross-account dedup is the higher-risk operation: it can collapse duplicates between independent archives whose provenance the user may want to preserve. Cross-account dedup should require explicit opt-in through `--collection`. A user who genuinely wants to dedup across every account can still write `--collection All`.

### Detection

Duplicate detection can use multiple signals:

- RFC822 `Message-ID`
- normalized raw MIME or body content hash
- provider/source message IDs where appropriate
- attachment content hashes where relevant

Detection signals should be merged into duplicate groups carefully. A content-hash match can connect messages that do not share a `Message-ID`, and a `Message-ID` match can connect messages with slightly different stored bodies. The grouping model should allow transitive duplicate sets rather than treating each signal as an isolated pass.

### Survivor selection

Survivor selection should be deterministic and explainable. The policy prefers the copy that is most useful as the durable representative, evaluated in this priority order:

1. source preference when configured
2. has raw MIME or complete original payload
3. source metadata quality
4. richer label or folder metadata
5. earlier archived timestamp when meaningful
6. stable row ID as the final tie-breaker

Earlier rules win outright; later rules only apply when all earlier ones tie. The exact policy should be documented and visible in dry-run output, so a user can read why one copy survived and another was hidden.

#### Sent-message safety rule

Sent-copy safety is an **eligibility filter**, not a tie-breaker. When any message in a duplicate group looks like a sent copy, only sent copies are eligible to survive. Received-copy candidates are removed from the group before the priority list above runs. Losing the sent signal silently changes user interpretation of the archive — "I sent this" is harder to recover than "I received this."

A message looks like a sent copy when any of these signals fires (OR):

- a Gmail `SENT` label on the message
- an `is_from_me` flag on the message from ingest metadata
- the `From` address matches a confirmed identity for the message's account

### Effects

Applying dedup should:

- choose one survivor per duplicate group
- hide redundant local rows from normal query paths
- preserve enough metadata to explain what happened
- write a batch ID for audit and undo
- avoid remote deletion unless explicitly requested

Dedup should not silently escalate from local hiding to local hard deletion or remote deletion.

## Safety progression

Dedup is a ladder, not a single switch. Each rung is a separate, explicit user action. The system never escalates from one rung to the next on its own.

1. **Scan.** Detect duplicates and report what would change. No data is touched. Dry-run is the default.
2. **Hide.** Apply dedup. Pruned copies are soft-deleted: hidden from normal reads but kept on disk. `--undo <batch-id>` restores visibility.
3. **Local hard delete.** A separate, opt-in action that permanently removes hidden rows from the local archive. Dedup itself never does this; the user runs it explicitly after a hide step they're confident in.
4. **Remote delete.** Deleting from the source server (Gmail, IMAP, another service) is a further separate decision. The default is trash-with-recovery (Gmail's ~30-day trash). Permanent remote deletion requires explicit opt-in and interactive confirmation.

The user controls every rung. "Apply dedup" never implies hard delete. "Hard delete locally" never implies remote delete. "Remote delete" never implies permanent remote delete.

Attachment dedup is independent of message dedup: attachments are stored in a content-addressed pool, so identical files are stored once regardless of how many messages reference them. Hiding or hard-deleting a duplicate message does not delete the underlying attachment blob unless no remaining message references it.

## Live-message contract

A **live message** is a message that has not been locally hidden by dedup and has not been recorded as deleted from the source server. The term is internal vocabulary for this contract and shows up in implementation slices and code.

Normal user-facing reads should return live messages only.

This contract applies to:

- message search
- vector and hybrid search
- TUI browsing
- stats and aggregates
- API responses
- MCP responses
- exports that claim to represent the visible archive

Indexes and caches may lag behind SQLite state, but normal retrieval must still filter hidden rows. Rebuilding derived surfaces is valuable for size and performance; it should not be the only thing preventing hidden duplicates from appearing.

## Query scope

Collections should be first-class query scopes.

If users can create `work` or `personal`, they should be able to search, browse, count, and inspect those collections without learning which source IDs are inside. That applies across local search, vector/hybrid search, TUI, API, MCP, and stats.

The scope model should produce the same result set regardless of retrieval backend:

- account scope maps to one source ID
- collection scope maps to many source IDs
- `All` maps to every source ID

Backend differences are acceptable for ranking or performance, but not for scope membership or live-message visibility.

## Cache and index policy

The product contract is:

- Dedup changes the canonical archive state.
- Normal reads hide rows that are no longer live.
- Derived indexes may be rebuilt, updated, or marked stale as an operational concern.

Recommended policy:

- Filtering is mandatory for correctness.
- Best-effort derived index cleanup is allowed.
- Manual rebuild commands remain available.
- Any known stale derived surface should be visible in command output or logs.

This avoids coupling dedup correctness to every cache and index implementation.

## Undo model

Undo is not full time travel.

Undo should restore local visibility for rows hidden by a dedup batch and cancel pending remote deletion manifests when they have not executed. It should not promise to reverse every side effect of dedup, such as survivor label unioning, raw-MIME enrichment, index cleanup, or remote deletion already performed against a source service.

Canonical user-facing language:

> `--undo <batch-id>` restores rows hidden by that dedup batch and cancels the batch's pending remote-deletion manifest where possible. It does not restore an exact pre-run database state.

## Remote deletion model

Remote deletion is a separate operation from local dedup.

Even when duplicate detection runs across a collection, remote deletion decisions remain source-specific. It is only valid to stage remote deletion when the survivor and loser belong to the **same source** and that source supports the requested remote-deletion behavior.

Rules:

- **Same-source constraint.** A remote-deletion entry is only staged when the loser and the survivor share a `source_id`. Cross-source duplicate groups produce no remote-deletion entries even when the dedup scope is a collection that spans those sources.
- **Source-scoped manifests.** Remote-deletion manifests, manifest filenames, and reporting labels reflect the source, never the collection name, even when dedup was invoked under `--collection`.
- **Trash by default.** Where the source supports a trash or recoverable state (e.g. Gmail's ~30-day trash), the default remote-deletion behavior moves messages there rather than removing them outright.
- **Permanent deletion is opt-in.** Permanent remote deletion requires an explicit flag and interactive confirmation. It is never the default, never inferred from dedup, and never applied in batch without the user acknowledging the source and scope at the moment of the action.

This preserves the distinction between "hide this redundant local row," "hard-delete it from the local archive," and "delete something from Gmail / IMAP / another source service."

## Schema and persistence

Accounts, collections, identities, dedup batches, and deletion manifests are core domain concepts. Their durable state belongs in canonical schema and migrations, with dialect-aware ownership where msgvault supports multiple database engines.

The target model needs durable storage for:

- collection definitions
- collection membership
- account-scoped identity records
- dedup batches
- hidden duplicate row metadata
- remote deletion manifests or manifest references

Ad hoc lazy table creation is acceptable only as a development bridge, not as the settled architecture for these concepts.

### Core scope

These concepts belong together and should be designed as one coherent model:

- Account/source as one ingest unit.
- Collection as explicit grouping.
- Default `All` collection.
- Account-scoped identities, with one-time migration from any legacy global identity config.
- Collection identity as derived union.
- Account-scoped dedup.
- Collection-scoped dedup.
- Sent-message safety as a survivor eligibility filter, not a tie-breaker.
- Live-message filtering across normal reads.
- Safety progression of scan → hide → local hard delete → remote delete, with no automatic escalation between rungs.
- Undo as local visibility restore, not full rollback.
- Remote deletion as explicit source-scoped follow-up, same-source-only, trash-by-default with permanent deletion behind interactive confirmation.

### Implementation slices

The implementation does not have to land all at once. Reasonable slices are:

- **Model and CLI scope:** vocabulary, `--account`/`--collection`, `All`, no nested collections.
- **Read scope and visibility:** live-message predicate, collection query scope, backend consistency.
- **Dedup application:** account and collection dedup, survivor policy, batch audit, undo language.
- **Identity persistence:** per-account identity records, discovery, confirmation, and collection union.
- **Remote deletion:** source-scoped manifests and collection-scope safety tests.

These slices should preserve the model even if delivered separately. A partial slice should not introduce user-facing semantics that contradict the target design.

### Future product work

These are valuable but do not need to define the first aligned implementation:

- Import-time dedup with `--into`.
- Automatic dedup when creating or adding to collections.
- Exporting a deduplicated collection into a clean account. Once a collection has been deduplicated, its survivors form a coherent unified view across the member sources. A future operation should be able to export those survivors into a single new account that becomes the canonical archive going forward, while the original member accounts remain intact for provenance. This gives users a path from "many overlapping imports" to "one clean source of truth" without forcing them to throw away the originals.
- Identity-derived inbound/outbound classification across historical imports.
- Rich identity review UI.
- Policy controls for source preference and survivor scoring.

## Mapping to PR #286

This issue is forward-looking in framing but most of the model is already implemented on the [`jesse/dedup-v2`](https://github.com/wesm/msgvault/tree/jesse/dedup-v2) branch. Recording the mapping here so the gap between the target model and the shipped code is visible:

- **Already on the branch:** account-as-source vocabulary, `--account` and `--collection` flags on dedup, default `All` collection bootstrap, per-account dedup, collection-scope dedup, sent-message safety in survivor selection, undo as local-visibility restore, source-scoped remote-deletion manifests with same-source-only staging, and per-account identity records via `list-identities`.
- **Partial:** live-message filtering (applied to SQLite and DuckDB read paths; coverage across vector/hybrid retrieval and MCP responses still needs an audit), name-collision errors between accounts and collections (basic guards in place; full ambiguity-suggestion UX is not), and collections as first-class query scopes outside dedup.
- **Not yet on the branch:** identity discovery beyond ingest metadata, identity confirmation UX, derived collection identity used at read time, policy controls for survivor scoring, the explicit local hard-delete rung of the safety progression, and the legacy `[identity]` config migration on first startup after upgrade.

The implementation slices above can be applied to the existing branch incrementally rather than as a single reshape.

## Things to consider prior to implmentation

Use this checklist before translating the design back into implementation tasks:

- Does "account" always mean one ingest source/archive?
- Is every cross-account operation expressed through a collection?
- Can users tell from the command or UI when they are crossing account/source boundaries?
- Are identities account-scoped rather than global, with a defined migration from any legacy global config?
- Is `All` modeled as a collection?
- Are collections first-class query scopes?
- Are hidden duplicates excluded from every normal read path by contract?
- Does dedup honor sent-message eligibility before falling back to the survivor priority list?
- Does dedup keep scan / hide / local hard delete / remote delete as four separate user actions, with no automatic escalation between them?
- Does remote deletion stay same-source-only, trash-by-default, and require explicit confirmation for permanent removal?
- Does undo avoid promising exact rollback?
- Are implementation slices allowed only when they preserve these semantics?

jesserobbins:
## Summary

Implements the identity, collection, and deduplication system from #278.

### New commands
- **`list-identities`** — auto-discover sent-from addresses across accounts using three signals (From header, OAuth, config); prints likely identities for configuring the `[identity].addresses` list.
- **`collections`** — manage named source groups (`create`, `list`, `show`, `add`, `remove`, `delete`); a default `All` collection is seeded automatically and tracks every source.
- **`deduplicate`** — find and merge duplicates across sources for the same account.

### `deduplicate` behavior
- Groups by RFC822 `Message-ID`; optional `--content-hash` also groups by normalized raw MIME.
- Picks a survivor (honoring `--prefer <source-types>` and identity-based sent-copy preference), unions labels, soft-deletes the pruned copies.
- **Read-only by default** (`--dry-run`). `--apply` required to write.
- **Reversible**: every run writes a per-source manifest; `--undo <batch-id>` restores hidden rows. `--undo` is repeatable to undo multiple batches.
- Atomic snapshot before merge via SQLite `VACUUM INTO` (skip with `--no-backup`).
- `--delete-dups-from-source-server` additionally stages pruned copies for remote deletion (destructive, opt-in; remote sources only).

### Query layer
- `SourceIDs` filter propagated through DuckDB and SQLite query paths, plus soft-delete exclusion everywhere.
- TUI account selector lists collections and filters results to the collection's sources.

Closes #278.


jesserobbins:
Fixed (5 total):
  - --account zero-source guard in deduplicate
  - --account zero-source guard in list-identities
  - batchID path traversal sanitization
  - Empty SourceIDs nil-vs-empty distinction in query filter
  - Missing deleted_at IS NULL in DuckDB GetGmailIDsByFilter

Disagree with the following: 
  - Invalid rune ' @' — code was already correct
  - Scan backfill mutates during dry-run — idempotent enrichment, required for
  results
  - Scan empty AccountSourceIDs for non-CLI callers — speculative, CLI handles
  correctly
  - Collections create/add should auto-trigger dedup — dedup is a separate
  explicit operation
  - EnsureDefaultCollection not wired in — already called in InitSchema
  - TOML escaping — already uses proper toml.NewEncoder

jesserobbins:
> ## roborev: Combined Review (`52f1cd4`)
> Deduplicated verdict: PR has **2 Medium findings** around deduplication scope and unintended mutation; no High or Critical findings.
> 
> ## Medium
> * **`internal/dedup/dedup.go:204`**: `Engine.Scan` backfills `rfc822_message_id` before honoring dry-run behavior, and the backfill runs across the whole database even for scoped `--account` scans. This means `msgvault deduplicate --dry-run` or account-scoped scans can still mutate unrelated rows before confirmation or backup.
>   **Fix**: Make the backfill respect `DryRun`, and pass `AccountSourceIDs` through `CountMessagesWithoutRFC822ID` / `BackfillRFC822IDs` so only scoped sources are touched.

Dry run, in my opinion, shouldn't do anything that writes or mutates... including backup. It should clearly disclose what it would have done, and what it can't do because of a potential mutate.

> * **`internal/dedup/dedup.go:233`**: The engine contract says empty `AccountSourceIDs` means per-source deduplication, but `Scan` passes an empty source list to store methods, where it means “all sources.” A direct engine caller using the default config can therefore merge duplicates across accounts, contradicting the documented safety rule.
>   **Fix**: Require a non-empty scope for engine scans, or implement true per-source grouping when `AccountSourceIDs` is empty.

Okay, I agree with that.

wesm:
Haven't forgotten about this, will look at this soon!

wesm:
*via Codex*

Thanks for putting this together. The direction is useful, especially the reversible local dedup flow, survivor selection, batch IDs, and the idea of default/user-defined collections. Before this goes further, I think we should align on the long-term data model and scope semantics, because several parts of this branch seem to encode a different model than I think msgvault should use.

My current understanding of the model we want is:

- **Account**: one ingest source/archive. Examples: one Gmail sync source, one IMAP source, one mbox import, one Apple Mail import, one iMessage import.
- **Collection**: a named grouping of accounts/sources. Examples: `All`, `work`, `personal`, `old laptop imports`.
- **All**: the default collection containing every account/source.
- Dedup within an account should only dedup duplicate rows inside that one ingest source.
- Dedup across accounts should require an explicit collection boundary.

That model is attractive because it gives us a crisp safety boundary: we never dedup across independent archives unless the user has deliberately grouped them into a collection.

Concerns and questions:

1. **Terminology/model mismatch**

   The spec and package comments sometimes describe an account as a logical multi-source mailbox. For example, `internal/dedup/dedup.go` says dedup runs against a "single logical account" and that an account may include Gmail plus mbox for the same mailbox. The spec also says the same address can appear in multiple accounts, which points the other way: each import is an account/source and collections group them.

   I think we should decide this explicitly before merging. My preference is account = one ingest source, collection = grouping. If we agree, the docs, comments, CLI help, and config names should consistently use that model.

2. **`--account` is overloaded with collection behavior**

   `ResolveAccount` checks collections first, then sources, and `deduplicate --account` uses that resolver. So `--account work` may target a collection named `work`, not an account/source. That makes CLI behavior surprising and introduces name-shadowing risk if a collection and account share the same identifier/display name.

   Suggestion: keep `--account` restricted to one account/source, and add an explicit `--collection` flag for collection-level operations. If a command wants a generic scope, model that internally as a `Scope`, but expose precise flags at the CLI boundary.

3. **The spec mentions `--collection`, but the implementation routes through `--account`**

   The spec says `msgvault deduplicate --collection personal`, but the CLI only has `--account`, with collection resolution hidden inside it. I would rather make collection dedup explicit:

   - `msgvault deduplicate --account <account>`: one source/account only
   - `msgvault deduplicate --collection <collection>`: all accounts in that collection
   - `msgvault deduplicate`: each account/source independently

4. **Collection creation can accidentally accept collections as members**

   `collections create --accounts ...` calls `ResolveAccount`, which means a collection name can expand to source IDs. That effectively allows nested collection references even though the schema only stores source IDs. If nested collections are not a supported concept, this should reject collection references and require actual account/source identifiers or IDs.

5. **Identity configuration is global, but the spec says identities are tied to accounts/collections**

   The config adds a global `[identity] addresses = [...]`, while the spec says identities are tied to accounts and a collection identity is the union of member account identities. Global identity addresses may be useful as a temporary input, but they are a different model. This matters for sent-copy detection: an address that is "me" in one account/collection may not be safe to apply globally across all scopes.

   We may want either:

   - stored per-account identity records, with collection identity derived from members, or
   - an explicit statement that identity is currently global and account-scoped identities are future work.

6. **Dedup-hidden messages are not hidden from every search surface**

   Most SQLite/DuckDB query paths now filter `messages.deleted_at IS NULL`, but vector/hybrid search still appears to filter only `deleted_from_source_at IS NULL`:

   - `internal/vector/sqlitevec/backend.go` live-hit filtering
   - `internal/vector/sqlitevec/backend.go` `filteredMessageIDs`
   - `internal/vector/sqlitevec/fused.go` fused search CTE

   Since dedup uses `messages.deleted_at` as a local hide flag, vector/hybrid results can still surface pruned duplicates. If the contract is "hidden from all query paths," vector/hybrid needs the same filter or a shared live-message predicate.

7. **Collection scoping is only partially propagated through msgvault**

   Query structs now have `SourceIDs`, which is a good start, but several user-facing paths still resolve only a single source/account:

   - local `search --account` uses `GetSourceByIdentifier`
   - vector/hybrid `search --account` uses `GetSourceByIdentifier`
   - text/vector/TUI/API/MCP paths appear to remain mostly single-`SourceID` oriented
   - some stats/sub-aggregate label/account counts still only special-case `SourceID`, not `SourceIDs`

   We should decide whether collections are only a dedup/admin concept in this PR, or whether they are intended to be first-class query scopes. If first-class, we probably need a common scope resolver and tests across search, stats, TUI, API, MCP, and vector paths.

8. **Cache/index invalidation needs a clearer policy**

   Dedup tells the user to run `build-cache --full-rebuild`, and the Parquet export excludes `deleted_at` rows. That handles one cache if the user follows the instruction. But FTS and vector indexes are separate surfaces. Filtering can be enough, but then every retrieval path must consistently apply `deleted_at IS NULL`; otherwise the index can retain hidden messages and leak them in results.

   I think we should document one of these policies:

   - dedup mutates only SQLite and every query/index path must filter local-hidden rows, or
   - dedup also schedules/rebuilds affected derived indexes/caches, or
   - dedup requires explicit rebuild commands and refuses/flags stale derived surfaces.

9. **Schema ownership is split**

   Collection tables are created by `ensureCollectionSchema` instead of the main schema files. That may be pragmatic for this branch, but msgvault has been moving toward dialect-aware schema ownership. If collections are permanent core concepts, they probably belong in the canonical schema/migration path rather than being hidden behind lazy creation.

10. **Undo is not a full rollback**

   `UndoDedup` restores hidden duplicate rows but intentionally does not undo label unioning or raw MIME backfill. That is probably acceptable, but the user-facing language should be very precise: undo restores pruned rows and cancels pending manifests where possible; it does not restore the exact pre-run database state. This is especially important if agents/users rely on `--undo` as a safety guarantee.

11. **Remote deletion manifest semantics need a second look under collection scopes**

   The implementation has a good safety rule: only stage remote deletion when loser and survivor share the same `source_id`. That avoids deleting from Gmail just because a local backup had a copy. But once collections are in play, reporting, manifest naming, and `Account` labeling need to remain source/account-specific rather than collection-specific. It would be good to add explicit tests for collection-scoped dedup with remote deletion enabled.

12. **Spec includes broader behavior that is not implemented here**

   The spec discusses import-time dedup (`--into`), automatic dedup when creating/adding to collections, collection export to an account, and identity-derived inbound/outbound classification. This PR implements some foundational pieces, but not all of those flows. That is fine if intentional, but the PR description/spec should distinguish what is implemented now from the longer-term proposal.

Recommended architecture direction:

- Keep the core store model simple: sources/accounts are ingest units; collections are many-to-many groupings of source IDs.
- Introduce an explicit internal scope type, something like `Scope{Kind: Account|Collection|All, SourceIDs: []int64, DisplayName: string}`.
- Make CLI flags precise: `--account` for one source/account, `--collection` for collections. Avoid letting one flag silently resolve to either.
- Centralize live-message filtering so `deleted_from_source_at` and local dedup `deleted_at` semantics are not reimplemented inconsistently across SQLite, DuckDB, vector, FTS, API, MCP, and TUI paths.
- Treat identity as a scoped concept, or explicitly call out that global identity config is interim.

I think this branch is useful, but I would like us to settle these boundaries before we bake in the naming/API shape. The main thing I want to avoid is agents and humans reading "account" differently across commands, docs, and package APIs.

jesserobbins:
This is good feedback. Let me take another pass at clarifying the spec first.

wesm:
Let me know if I can assist, this will be good to get designed correctly so it can provide a good foundation for different types of msgvault frontend environments

jesserobbins:
@wesm I did a pretty deep review on the feedback, worked through a couple of the other design elements in #278. I am going to update this branch with those changes. Let me know if you have further feedback.

jesserobbins:
Changed my mind. Am closing this branch and will submit a fresh pr in the morning.

jesserobbins:
Superseded by #304 — same scope, flat 5-commit history, design docs and worked-out alignment notes stripped out for review/merge clarity. Leaving this PR open as the historical record (the design alignment review trail and iterative refinement live here).

PR #288, opened 2026-04-22
feat: add rebuild-fts command for FTS5 shadow-table corruption (#287)

### LINKED ISSUE #287: FTS5 + B-tree corruption on contentful index: recovery requires manual DROP+CREATE+backfill
## Summary

On a large archive (~34 GB, 422k messages), `msgvault verify` surfaced persistent SQLite corruption:

```
Tree 26 page 8231140 cell 2: Rowid 421177 out of order
Tree 58 page 8231146 cell 2: Rowid 421177 out of order
malformed inverted index for FTS5 table main.messages_fts
```

MIME sample verified 100% OK — only the FTS5 index and one `message_bodies` B-tree page were affected.

## Environment

- **msgvault**: v0.14.0 (corruption also present in v0.12.1)
- **OS**: macOS 15.1 (Darwin 25.3.0), Apple Silicon
- **Archive**: 422,009 messages, 34 GB, APFS, SQLite bundled
- Archive grew over ~6 months of continuous sync; no specific incident pinpointed

## What didn't work

1. **`INSERT INTO messages_fts(messages_fts) VALUES('rebuild')`** — completes in ~2 min but leaves the same B-tree corruption. Presumably because it regenerates from within the corrupt shadow tables rather than from the source `messages`/`message_bodies`/`message_recipients`/`participants` tables.
2. **`INSERT INTO messages_fts(messages_fts) VALUES('delete-all')`** — errors: `'delete-all' may only be used with a contentless or external content fts5 table`. `messages_fts` is contentful (no `content=` clause).
3. **`.recover`** — not feasible with <34 GB free.

## What worked

Full manual DROP + recreate + batched backfill using the SQL already present in `internal/store/dialect_sqlite.go` (`FTSBackfillBatchSQL`):

1. Stop daemon + MCP
2. `DROP TABLE messages_fts;`
3. `CREATE VIRTUAL TABLE messages_fts USING fts5(message_id UNINDEXED, subject, body, from_addr, to_addr, cc_addr, tokenize='unicode61 remove_diacritics 1');`
4. Loop `FTSBackfillBatchSQL` over 5000-ID batches (mirrors `Store.BackfillFTS`)
5. `PRAGMA integrity_check` — FTS5 error cleared. 

Total time: ~5 min for 422k rows. Peak extra disk: ~3 GB (FTS5 shadow tables for this archive are ~2.5 GB).

The remaining `Tree 26 ... Rowid out of order` in `message_bodies` turned out to be a phantom-duplicate row on a corrupt page — resolved via `DELETE FROM message_bodies WHERE message_id = <rowid>; INSERT ...;` (hit a UNIQUE-constraint error on re-insert, which revealed a shadow duplicate; one copy deleted, page ordering healed).

Final `PRAGMA integrity_check`: **ok**. `msgvault verify` now exits clean.

## Suggestions

1. **Expose a maintenance command** such as `msgvault rebuild-fts` that runs the same DROP + CREATE + batched backfill — currently users must either know SQLite internals or do a full resync (or `.recover`, which requires 2× disk).
2. **Document recovery in `docs/`** — the contentful-FTS5 corruption path is non-obvious (the 'rebuild' and 'delete-all' pragmas behave differently than on contentless tables).
3. **Optional: `serve --skip-db-check` flag** (mirroring `verify --skip-db-check`) so a corrupted-but-functional archive can still run the daemon while the user schedules a rebuild.

Happy to PR (1) if you agree with the approach — essentially `Store.RebuildFTS()` = drop + create-from-schema + call existing `Store.BackfillFTS`.

## Related

- #260 (`failed to fetch message (nil response)`) — hit 11 of these during today's sync but unrelated to the FTS corruption.

wesm:
Closes #287.

## What this adds

- **`msgvault rebuild-fts`** — drops and recreates `messages_fts`, then repopulates it from `messages` / `message_bodies` / `message_recipients` / `participants` via the existing batched backfill. Recovery path for `malformed inverted index for FTS5 table main.messages_fts` in `verify` output — the case where SQLite's own `rebuild` pragma and `delete-all` don't help on a contentful FTS5 table. Wraps `SQLITE_BUSY`/`SQLITE_LOCKED` as "stop msgvault serve / MCP and retry."
- **`verify` ordering** — integrity check now runs before OAuth setup, so corrupt databases surface the repair hint even with an expired token.
- **`verify` hints** — split between FTS-only corruption (points at `rebuild-fts`) and core-table corruption (points at `.recover`).
- **`docs/recovery.md`** — covers both paths and the contentful-FTS5 caveat.

## What this does not cover

- Core-table B-tree corruption (e.g., `Rowid out of order` in `messages` / `message_bodies`). `rebuild-fts` is strictly for the derived index; core corruption still needs `.recover`.
- `_synchronous=FULL` default. Worth considering as archival-durability hardening in a separate PR; not framed as the fix for #287.

## Usage

```
msgvault rebuild-fts
```

Stop `msgvault serve` and MCP clients first — needs an exclusive write lock. Peak extra disk ≈ size of the FTS5 shadow tables (a few percent of the DB).

### DIFF
--- a/cmd/msgvault/cmd/rebuild_fts.go
+++ b/cmd/msgvault/cmd/rebuild_fts.go
@@ -1,0 +1,76 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/wesm/msgvault/internal/store"
+)
+
+var rebuildFTSCmd = &cobra.Command{
+	Use:   "rebuild-fts",
+	Short: "Rebuild the full-text search index from scratch",
+	Long: `Drop and recreate the messages_fts virtual table, then repopulate it
+from messages / message_bodies / message_recipients / participants.
+
+Use this to recover from FTS5 shadow-table corruption that surfaces as
+"malformed inverted index for FTS5 table main.messages_fts" in
+'msgvault verify' output. SQLite's own 'rebuild' pragma reads from the
+same corrupt shadow tables and cannot clear this state.
+
+This command only fixes the derived search index. Core-table corruption
+(e.g., "Rowid out of order" in messages / message_bodies B-trees) requires
+a different recovery path — see 'msgvault verify' output.
+
+Peak extra disk usage is roughly the size of the FTS5 shadow tables
+(a few percent of the SQLite database). Stop 'msgvault serve' and any
+MCP clients before running this command — it needs an exclusive write lock.`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		dbPath := cfg.DatabaseDSN()
+		s, err := store.Open(dbPath)
+		if err != nil {
+			return fmt.Errorf("open database: %w", err)
+		}
+		defer func() { _ = s.Close() }()
+
+		if err := s.InitSchema(); err != nil {
+			return fmt.Errorf("init schema: %w", err)
+		}
+
+		fmt.Fprintln(os.Stderr, "Rebuilding full-text search index...")
+		n, err := s.RebuildFTS(func(done, total int64) {
+			if total <= 0 {
+				return
+			}
+			if done > total {
+				done = total
+			}
+			pct := int(done * 100 / total)
+			barWidth := 30
+			filled := barWidth * pct / 100
+			bar := strings.Repeat("=", filled) +
+				strings.Repeat(" ", barWidth-filled)
+			fmt.Fprintf(os.Stderr, "\r  [%s] %3d%%", bar, pct)
+		})
+		if err != nil {
+			fmt.Fprintln(os.Stderr)
+			if s.IsBusyError(err) {
+				return fmt.Errorf(
+					"database is busy — stop 'msgvault serve' and any MCP " +
+						"clients, then retry",
+				)
+			}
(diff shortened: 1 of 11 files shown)


wesm:
The compile-break finding is a false positive. Verified:

- \`ls internal/store/dialect*.go\` returns only \`dialect.go\` (the interface) and \`dialect_sqlite.go\` (the sole implementation). No other type implements \`Dialect\`.
- \`go build ./...\` and \`go build -tags fts5 ./...\` both pass locally.
- All CI build/test jobs are green on this diff: **test** (Linux), **test-windows**, **nix-build**, **validate** (Docker). A missing interface method would fail every one of them at compile time.

The PostgreSQL dialect referenced in the new method's docstring (\`PostgreSQL: TODO (REINDEX / recompute tsvector column)\`) is forward-looking documentation for the in-progress PG refactor (see #276 "PostgreSQL dialect refactor - PR1 of 4 - Dialect"). No PG dialect type exists yet. When it lands, whoever writes it will implement \`FTSRebuildSchema\` and \`IsBusyError\` as part of satisfying the interface — which is exactly how interface evolution is supposed to work.

Not making changes based on this finding.

PR #292, opened 2026-04-24
feat: add search_by_domains MCP tool for multi-domain participant search

hansn74:
## Summary

Adds a new MCP tool, `search_by_domains`, that finds emails where **any participant** (from, to, cc, or bcc) belongs to one of the given domains. Useful for finding all communication with a company regardless of direction — sent, received, or CC'd.

This complements the existing `search_messages` tool, which only matches against the sender. `search_by_domains` is the right tool when you want "show me everything involving @example.com" rather than "show me mail from @example.com".

## Tool signature

```
search_by_domains(
  domains: "acme.com,globex.com"   // comma-separated, required
  limit:   100,                     // default 100, max 1000
  offset:  0,
  after:   "2024-01-01",            // optional date range
  before:  "2024-12-31",
)
```

## Implementation

- `internal/query/engine.go` — `SearchByDomains` added to the Engine interface
- `internal/query/sqlite.go` — EXISTS subquery against `message_recipients` + `participants` (semi-join, no duplicates)
- `internal/query/duckdb.go` — delegates to SQLite (participant/recipient data isn't in the Parquet cache)
- `internal/remote/engine.go` — returns `ErrNotSupported` (matches existing remote-engine pattern)
- `internal/query/querytest/mock_engine.go` — mock support with an override hook
- `internal/mcp/server.go` / `handlers.go` — tool definition + handler that parses comma-separated domains, validates non-empty, forwards limit/offset/after/before
- `internal/mcp/server_test.go` — 7 subtests (valid input, whitespace handling, missing/empty domains, argument forwarding, defaults)

## Test plan

- [x] Field-tested against a 2M-message production archive: returns correct results including BCC matches that `search_messages` misses
- [x] Case-insensitive domain matching verified
- [x] Multi-domain (`acme.com,globex.com`) and date filters verified

### DIFF
--- a/internal/mcp/handlers.go
+++ b/internal/mcp/handlers.go
@@ -1009,3 +1009,44 @@ func (h *handlers) stageDeletion(ctx context.Context, req mcp.CallToolRequest) (
 
 	return jsonResult(resp)
 }
+
+func (h *handlers) searchByDomains(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	args := req.GetArguments()
+
+	domainsStr, _ := args["domains"].(string)
+	domainsStr = strings.TrimSpace(domainsStr)
+	if domainsStr == "" {
+		return mcp.NewToolResultError("domains is required"), nil
+	}
+
+	// Split and clean domain list
+	var domains []string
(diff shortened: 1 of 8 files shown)


wesm:
LGTM, thank you!

PR #294, opened 2026-04-25
embed: windowed ETA and oversize-batch downshift

### LINKED ISSUE #295: Add per-batch progress reporting with windowed-rate ETA to build-embeddings
## What I was doing

Running long-running mailbox backfills with `msgvault build-embeddings` against [`maclocal-api`](https://github.com/scouzi1966/maclocal-api) (a local OpenAI-compatible shim over Apple Foundation Models prerelease embeddings, on-device).

## What I wanted

A progress line on stderr showing what the worker is doing — current rate, percent done, and a usable ETA. Today `build-embeddings` is silent until it returns; for a backfill measured in hours, there's no signal whether the run is healthy, whether the rate has shifted, or whether it's safe to leave overnight.

## Why it matters

ETA is the operator's main signal for when to come back to a long run. A naïve cumulative-rate ETA decays badly when the model warms up, the device thermally throttles, or the worker has to slow down — the displayed rate barely moves and the ETA stays confidently wrong, which is worse than no ETA. A windowed rate over the last N successful batches stays responsive to recent throughput, which is what an operator actually wants to see.

Implementation: #294.

### LINKED ISSUE #296: build-embeddings aborts when one oversize message poisons a multi-message batch
## What I was doing

Running long-running mailbox backfills with `msgvault build-embeddings` against [`maclocal-api`](https://github.com/scouzi1966/maclocal-api) (a local OpenAI-compatible shim over Apple Foundation Models prerelease embeddings, on-device). The shim has a configured request-body size limit; real mailboxes contain the occasional oversize message — long quoted threads, large pasted documents, attachments that survived preprocessing — that exceeds it.

## What happened

When a batch contains an oversize message and the endpoint refuses with HTTP 413, the worker fails the *whole* batch, retries it unchanged, fails again, and after `MaxConsecutiveFailures` cycles aborts:

```
Error: embed run: embed worker aborting after 5 consecutive failures: embed: HTTP 413 …
```

The fix used to be lowering `batch_size` globally and restarting, even though typically only one message is the cause. This isn't specific to 413 — any non-retryable 4xx that's a function of input (`400 Invalid embedding input`, `422 Unprocessable Entity`, model-context errors) produces the same poison-batch dynamic.

## Why it matters

A long backfill shouldn't get derailed by a single message it could have skipped or downsized around. The worker already drops messages that preprocess to empty (#293) — there's a precedent for "this message can't be embedded; acknowledge the queue and move on." Extending that to "this *batch* triggered a non-retryable 4xx; let me try them one at a time and figure out which one is the problem" follows the same shape.

Implementation: #294.

jesserobbins:
Closes #295 (feature), closes #296 (bug).

`build-embeddings` was silent until it returned and aborted when a single oversize message poisoned a multi-message batch — both unworkable on long mailbox backfills against [`maclocal-api`](https://github.com/scouzi1966/maclocal-api) (a local OpenAI-compatible shim over Apple Foundation Models prerelease embeddings).

Two scopes, one commit each, independently cherry-pickable:

1. **Per-batch progress reporting with windowed-rate ETA.** A stderr line about every two seconds with rate, ms/msg, µs/char, and ETA computed over a sliding window of the last N successful batches.
2. **Drain oversize batches one message at a time on non-retryable 4xx.** New `embed.ErrPermanent4xx` sentinel; the worker splits the failing batch and walks IDs one at a time instead of failing the whole batch and tripping the abort cap.

## How to use

- Progress prints to stderr automatically; nothing to enable.
- Tune the smoothing with `eta_window = 10` (default) under `[vector.embeddings]`.
- Downshift on oversize batches is automatic; entry/exit/interrupt are logged at Info level (`embed: downshifting to BatchSize=1 …`, `embed: downshift drain complete …`).

Full motivation, design choices, counter rules, and verification are in the two commit messages.

### DIFF
--- a/cmd/msgvault/cmd/embed_progress.go
+++ b/cmd/msgvault/cmd/embed_progress.go
@@ -1,0 +1,76 @@
+package cmd
+
+import "time"
+
+// rateWindow accumulates per-batch (msgs, elapsed) samples in a ring
+// buffer and reports the message-weighted rate over the window:
+//
+//	rate = sum(msgs) / sum(elapsed_seconds)
+//
+// Weighting matters when batch sizes vary a lot, e.g. when the worker
+// downshifts to BatchSize=1 to drain a poison message: a simple mean
+// of per-batch rates would let those tiny batches dominate the
+// displayed throughput. Summing first keeps small batches'
+// contribution proportional to their size.
+type rateWindow struct {
+	msgs    []int
+	elapsed []time.Duration
+	head    int // index of the next write
+	count   int // number of valid entries (<= len(msgs))
+}
+
+// newRateWindow constructs a ring buffer of the given capacity. A
+// non-positive cap is normalized to 1 so callers can always Add and
+// Rate without checking — this matches how applyDefaults treats
+// non-positive ETAWindow values from TOML, but defends against any
+// caller that bypasses defaults.
+func newRateWindow(capacity int) *rateWindow {
+	if capacity < 1 {
+		capacity = 1
+	}
+	return &rateWindow{
+		msgs:    make([]int, capacity),
+		elapsed: make([]time.Duration, capacity),
+	}
+}
+
+// Add records one batch sample. Samples with non-positive elapsed are
+// silently skipped: they would either divide by zero (elapsed == 0)
+// or contribute negative time (elapsed < 0, only reachable via clock
+// non-monotonicity in test harnesses). The next legitimate sample
+// re-establishes a healthy window.
+func (w *rateWindow) Add(msgs int, elapsed time.Duration) {
+	if elapsed <= 0 {
+		return
+	}
+	w.msgs[w.head] = msgs
+	w.elapsed[w.head] = elapsed
+	w.head = (w.head + 1) % len(w.msgs)
+	if w.count < len(w.msgs) {
+		w.count++
+	}
+}
+
+// Samples returns the number of valid entries currently in the
+// window. Useful for the printer's "(last K)" annotation when the
+// run hasn't yet emitted enough batches to fill the window.
+func (w *rateWindow) Samples() int { return w.count }
+
+// Rate returns the message-weighted rate in messages per second.
+// Returns 0 when the window is empty so callers can treat 0 as "no
+// estimate yet" and fall back to a no-ETA display branch.
+func (w *rateWindow) Rate() float64 {
+	if w.count == 0 {
+		return 0
+	}
+	var totalMsgs int
+	var totalElapsed time.Duration
+	for i := 0; i < w.count; i++ {
+		totalMsgs += w.msgs[i]
+		totalElapsed += w.elapsed[i]
+	}
+	if totalElapsed <= 0 {
+		return 0
+	}
+	return float64(totalMsgs) / totalElapsed.Seconds()
+}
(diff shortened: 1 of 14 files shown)


jesserobbins:
Addressed in 81b3e1b — empty-claim exit now surfaces the most recent all-drop error when `res.Succeeded == 0`, so a misconfigured endpoint can no longer return nil from `RunOnce` even when the queue is smaller than `BatchSize × MaxConsecutiveFailures` and the inline cap never trips. New regression test `TestWorker_DownshiftDrain_AllDropClean_SurfacesError` covers the case (4 rows, BatchSize=2, default cap=5 — would have exited cleanly before this fix).

jesserobbins:
Both findings addressed in 9b086e6.

**HIGH** (data loss on endpoint-wide 4xx): `downshiftDrain` no longer Complete-deletes singleton 4xxs eagerly. It now defers the drop, and at end-of-drain decides based on whether *any* singleton embedded:

- `embedded > 0` → endpoint works for some inputs, the 4xxs are message-specific (oversize, malformed). Complete the deferred IDs as drops.
- `embedded == 0` → endpoint-wide failure (401/403/404/etc.) cannot be ruled out. **Release** the deferred IDs back to the queue (never Complete-delete) and return the wrapped 4xx. The cap eventually trips with the original 4xx body intact, and the rows are available for retry once the operator fixes the config.

**MEDIUM** (singleton 4xx inconsistency): with the deferred-drop pattern, the `&& len(ids) > 1` guard is no longer needed — singleton 4xx going through the same drain just walks 1 ID, defers, finds `embedded == 0`, releases, returns. Same end state as the previous release-and-retry path, routed consistently. Caller's `errors.Is(err, ErrPermanent4xx)` check on the drain return prevents double-counting (the drain is confirming the upstream failure, not adding an independent one), so the cap still trips at the expected `MaxConsecutiveFailures` threshold.

New regression test `TestWorker_DownshiftDrain_AllDropClean_NoSilentDelete` simulates a `401 bad-api-key` against 4 rows: cap trips at the expected count, error carries the original 4xx body, and `assertPending == 4` confirms zero rows were silently deleted. Existing `TestWorker_SingletonBatch_4xx_NoSilentDelete` (renamed from `_FollowsExistingPath`) verifies the same guarantee at `BatchSize=1`.

jesserobbins:
Both findings are false positives — current `client.go` already does what's requested.

**Finding 1 (`client.go:174`):** the `>= 400` branch is unreachable for 429 and 5xx because they're returned earlier — `client.go:161` returns `&retryError{}` for 429, `client.go:171` returns `&retryError{}` for `>= 500`. Line 174 only fires when `400 <= status < 500 && status != 429`.

**Finding 2 (`client.go:106`):** `ErrPermanent4xx` is wrapped via `%w` on a plain `fmt.Errorf`, not a `*retryError`. At `client.go:102-104`, `errors.As(err, &re)` returns false for the 4xx case and `Embed` returns immediately — no retry, no backoff. An explicit `errors.Is(err, ErrPermanent4xx)` early-return would be redundant.

wesm:
Looks good here. I'm going to do a new full embedding run here myself to see what issues I run into

jesserobbins:
> Looks good here. I'm going to do a new full embedding run here myself to see what issues I run into

I have an embedding test harness that I've been playing with which you might try -> https://github.com/jesserobbins/msgvault-1/tree/embedshootout

wesm:
merging!

PR #297, opened 2026-04-25
feat: add Google service account support for domain-wide delegation

hansn74:
## Summary

Adds optional Google service account auth (with domain-wide delegation) alongside the existing browser-OAuth flow. A single service-account key can sync any user in a Google Workspace domain — no per-user browser consent, no token files, no headless-server token-copying.

## Why

The existing OAuth flow is great for personal Gmail and small teams, but for archiving a whole Workspace domain it has friction:

- Each user needs to interactively authorize at least once
- Headless servers can't do device flow (Google blocks Gmail scopes from device flow)
- Tokens expire/get revoked, requiring re-auth

Domain-wide delegation is the standard Google answer: one service account, one key file, impersonate any user in the domain. msgvault's `oauth2.TokenSource` abstraction already absorbs the difference — only the credential plumbing needed to grow.

## Configuration

```toml
# Existing browser-OAuth flow still works exactly as before.
[oauth]
client_secrets = "/path/to/client_secret.json"

# New: a named app backed by a service-account key.
[oauth.apps.workspace]
service_account_key = "/path/to/service-account.json"
```

Either `client_secrets` or `service_account_key` (or both) can be set per app.

## Usage

```
msgvault add-account user@company.com --oauth-app workspace
# → "Account user@company.com authorized via service account."
# (no browser, no token file)

msgvault sync-full user@company.com
# uses the service account key to impersonate user@company.com
```

## Implementation

- **`internal/oauth/serviceaccount.go`** *(new)* — `ServiceAccountManager.TokenSource(email)` returns an `oauth2.TokenSource` via `google.JWTConfigFromJSON` with `Subject = email`. Auto-refreshes; no per-user storage needed.
- **`internal/oauth/oauth.go`** — adds `ValidateTokenEmail()` so callers without a `*Manager` can run the same Gmail-profile email check before persisting a source.
- **`internal/config/config.go`** — `OAuthConfig.ServiceAccountKey` and `OAuthApp.ServiceAccountKey`, plus `ServiceAccountKeyFor(name)`. `HasAnyConfig()` recognises service-account-only configs. Path expansion (`~/...`) and config-relative resolution apply to both fields.
- **`cmd/msgvault/cmd/addaccount.go`** — when the resolved app has `service_account_key`, skip the browser flow, validate access via the Gmail profile API, register the source.
- **`cmd/msgvault/cmd/{syncfull,sync,serve}.go`** — when a source is bound to a service-account app, build the `TokenSource` via `ServiceAccountManager` instead of `Manager.TokenSource`. The Gmail client is unchanged because both sides return `oauth2.TokenSource`.

The deletions in the diff are existing OAuth-flow lines wrapped into `else` branches — no functionality removed.

## Test plan

- [x] Field-tested by adding 24 Workspace accounts on a real msgvault deployment with no per-user browser auth
- [x] `add-account --oauth-app workspace foo@domain` validates by hitting `/gmail/v1/users/me/profile`
- [x] Mismatched impersonation (e.g. wrong email) surfaces as `TokenMismatchError`
- [x] Mixed setup (per-user OAuth for one account, service account for the rest) works in the same daemon

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,22 @@ To switch an existing account to a different OAuth app:
 msgvault add-account you@acme.com --oauth-app acme   # re-authorizes
 ```
 
+### Google Service Accounts
+
+Workspace admins can use a Google service account with domain-wide delegation instead of per-user OAuth tokens:
+
+```toml
+[oauth.apps.acme]
+service_account_key = "/secure/path/service-account.json"
+```
+
+In Google Admin Console, authorize the service account client for `https://www.googleapis.com/auth/gmail.readonly` and `https://www.googleapis.com/auth/gmail.modify`. If you will run `delete-staged` with permanent deletion, also authorize `https://mail.google.com/`. Keep the key file owner-only, for example `chmod 600 /secure/path/service-account.json`.
+
+```bash
+msgvault add-account you@acme.com --oauth-app acme
+msgvault sync-full you@acme.com
+```
+
 ## MCP Server
 
 msgvault includes an MCP server that lets AI assistants search, analyze, and read your archived messages. Connect it to Claude Desktop or any MCP-capable agent and query your full message history conversationally. See the [MCP documentation](https://msgvault.io/usage/chat/) for setup instructions.
--- a/cmd/msgvault/cmd/addaccount.go
+++ b/cmd/msgvault/cmd/addaccount.go
@@ -107,7 +107,45 @@ Examples:
 			}
 		}
 
-		// Resolve client secrets path
+		// Check for service account configuration first
+		if saKeyPath := cfg.OAuth.ServiceAccountKeyFor(resolvedApp); saKeyPath != "" {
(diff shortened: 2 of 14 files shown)


hansn74:
The latest roborev report ([Combined Review on `91395d8`](https://github.com/wesm/msgvault/pull/297#issuecomment-)) flags `internal/oauth/oauth.go:724` for a missing `io` import that would break the build. This is a false positive — `io` is in the import block at line 11:

```
$ grep -n '"io"' internal/oauth/oauth.go
11:	"io"

$ go build ./...
$ make test
ok  	github.com/wesm/msgvault/internal/oauth	1.066s
ok  	github.com/wesm/msgvault/cmd/msgvault/cmd	2.556s
[…all packages pass]
```

The branch builds clean and the full test suite passes locally. All the previously-flagged real issues (binding-clear bug, scope plumbing, delete-staged + verify gaps for service-account-only configs) have been addressed in the follow-up commits.

PR #299, opened 2026-04-26
feat: add --http flag to mcp command for StreamableHTTP transport

hansn74:
## Summary

Adds `--http <addr>` to `msgvault mcp` so the MCP server can serve over StreamableHTTP instead of stdio. Stdio remains the default.

## Why

The existing stdio transport works perfectly when an MCP client (e.g. Claude Desktop) spawns `msgvault mcp` as a local subprocess. It doesn't work when:

- msgvault runs as a daemon on a NAS / remote server
- Multiple MCP clients (Claude Desktop, IDE plugins, custom integrations) need to share a single archive
- The archive is on a different machine than the client (over Tailscale, VPN, etc.)

`msgvault serve` already exposes a REST API at `/api/v1/...`, but that's a different surface — MCP clients can't auto-discover tools through it. StreamableHTTP keeps the MCP protocol intact and lets any spec-compliant client connect to a remote msgvault instance the same way it would connect to a local one.

## Usage

```
# Stdio (existing default — unchanged)
msgvault mcp

# StreamableHTTP on port 8080
msgvault mcp --http :8080
```

systemd example:

```ini
[Service]
ExecStart=/usr/local/bin/msgvault mcp --http :8080
```

Then Claude Desktop / any MCP client connects to `http://msgvault.local:8080/mcp`.

## Implementation

- **`internal/mcp/server.go`** — extracts the existing tool-registration code into `newMCPServer(opts)`, then adds `ServeHTTPWithOptions(ctx, opts, addr)` that wraps the same server in `server.NewStreamableHTTPServer` (built into `mcp-go`). `Serve` and `ServeWithOptions` keep their existing stdio behaviour. Vector/hybrid options propagate the same way they do for stdio.
- **`cmd/msgvault/cmd/mcp.go`** — adds a `--http` flag; when non-empty, calls `ServeHTTPWithOptions` instead of `ServeWithOptions`.

No new dependencies — `mcp-go` already provides `NewStreamableHTTPServer`.

## Test plan

- [x] `msgvault mcp` (stdio) still works as before
- [x] `msgvault mcp --http :8080` accepts MCP `initialize` + `tools/call` over HTTP
- [x] Field-tested via systemd on a NAS deployment with Claude Desktop and a custom MCP client connecting concurrently

### DIFF
--- a/cmd/msgvault/cmd/mcp.go
+++ b/cmd/msgvault/cmd/mcp.go
@@ -3,7 +3,10 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 	mcpserver "github.com/wesm/msgvault/internal/mcp"
@@ -13,6 +16,8 @@ import (
 
 var mcpForceSQL bool
 var mcpNoSQLiteScanner bool
+var mcpHTTPAddr string
+var mcpHTTPAllowInsecure bool
 
 var mcpCmd = &cobra.Command{
 	Use:   "mcp",
@@ -82,7 +87,11 @@ Add to Claude Desktop config:
 			engine = query.NewSQLiteEngine(s.DB())
 		}
 
-		ctx, cancel := context.WithCancel(context.Background())
+		// Derive from cmd.Context() so signal handling installed by
+		// the cobra root command (SIGINT/SIGTERM → ctx.Done()) reaches
+		// the MCP transport and can trigger ServeHTTPWithOptions's
+		// graceful shutdown.
+		ctx, cancel := context.WithCancel(cmd.Context())
 		defer cancel()
 
 		// Build optional vector-search components. MCP runs as a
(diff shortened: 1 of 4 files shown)


hansn74:
The latest roborev report on `f609f57` flagged the new test for missing `time`/`errors` imports. That's a false positive — both are already in the import block (line 7 and 13 of `internal/mcp/server_test.go`):

```
$ grep -n '"time"\|"errors"' internal/mcp/server_test.go
7:	"errors"
13:	"time"

$ go test -tags fts5 ./internal/mcp/...
ok  	github.com/wesm/msgvault/internal/mcp	0.638s
```

The medium-severity concern (mcp command using `context.Background()` instead of `cmd.Context()`) was real and is fixed in `d194479` — derives from `cmd.Context()` so cobra's signal handling reaches the HTTP transport's graceful shutdown.

chris-hatton:
Yes please - stdio-only MCP is an awkward limitation when deploying in a Docker container.  Not impossible to expose (Unix pipes, anyone?) but certainly not conventional.

PR #303, opened 2026-04-29
imessage: backfill display names from vCard contacts

bjabes:
## Summary

iMessage-imported users showed up as raw phone numbers / emails because `chat.db` only stores handles — names came through only when another source (Gmail headers, Google Voice, WhatsApp `--contacts`) had already populated `participants.display_name`.

Two changes that together let names appear:

- **Stop poisoning `display_name` with the phone string** when the iMessage importer creates a new participant. The previous behavior left `display_name = "+15551234567"`, which blocked later imports from filling in a real name (the update guard only writes when `display_name` is `NULL`/empty). New iMessage rows now leave `display_name` empty and let the first name-bearing import win.

- **Add `--contacts <vcf>` to `import-imessage`**, mirroring WhatsApp's flag. Backfills names by phone and email from a vCard export (e.g. macOS Contacts.app → File → Export → Export vCard). Only updates participants that already exist with an empty `display_name`, so any prior name is preserved.

## Plumbing

- Extract the vCard parser into a new `internal/vcard` package so iMessage and WhatsApp share it (`whatsapp.ImportContacts` now wraps it). The parser also reads `EMAIL` fields, not just `TEL`/`FN`.
- Add `Store.UpdateParticipantDisplayNameByEmail` — case-insensitive, first-writer-wins, never creates rows.

## Usage

```
msgvault import-imessage --contacts ~/contacts.vcf
```

Output now includes:

```
Contacts applied:
  Source:           ~/contacts.vcf (1234 entries)
  Names backfilled: 412 by phone, 38 by email
```


bjabes:
Pushed `122189b` to address the legacy-placeholder feedback.

`import-imessage --contacts` now uses a new `Store.UpdateImessageParticipantDisplayNameByPhone` that overwrites when:
- `display_name IS NULL` / empty (existing behavior), **or**
- `display_name = phone_number` (legacy placeholder), gated on the participant having an `imessage` identifier so other-source participants are never touched.

Test (`TestUpdateImessageParticipantDisplayNameByPhone`) covers all three cases: legacy placeholder cleared, real Gmail name preserved, non-iMessage participant untouched. The email path is unchanged — iMessage email handles were never poisoned.

wesm:
Thanks for working on this, I hadn't yet looked into how to get people's names into the whatsapp/imessage screens. I need to see how this composes with the other open PRs so bear with me

PR #281, opened 2026-04-21
Add Facebook Messenger DYI import

jesserobbins:
## Summary

- Adds `msgvault import-messenger` command for Facebook "Download Your Information" exports
- Supports JSON, HTML, and E2EE flat-file formats with auto-detection per thread
- Handles both old `messages/` and newer `your_activity_across_facebook/messages/` DYI layouts
- Decodes Facebook's Latin-1-over-UTF-8 mojibake in JSON exports transparently
- Ingests attachments (photos, videos, stickers, audio) into content-addressed storage
- Preserves reactions as relational rows and in message body for FTS
- Resumable checkpoints every 50 threads
- Removes hard-coded `message_type = 'email'` filter from query engines so all archived message types participate in search and aggregation

~4,700 lines across 44 files. Core importer is ~2,100 lines with ~1,600 lines of tests.

Closes #280. Related: #136, #192, #278.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #304, opened 2026-04-30
Accounts, Identities, Collections, and Deduplication

### LINKED ISSUE #278: Feature: Identities, Collections, and Deduplication

jesserobbins:
Implements and resolves #278 (supersedes #286). Specification:
[`docs/accounts-identities-collections-dedup/spec.md`](docs/accounts-identities-collections-dedup/spec.md).

## What this branch adds

A reusable account/collection scope, per-account identity storage, and
a deduplication pipeline staged into a four-rung safety progression
(scan → hide → local hard delete → remote delete). Every read surface
in the archive routes through one canonical visibility predicate.

## New CLI surface

| Command | Purpose |
| --- | --- |
| `msgvault deduplicate` (aliases `dedup`, `dedupe`) | Find and merge duplicates by RFC822 Message-ID with optional `--content-hash` fallback. Scoped via `--account` or `--collection`. `--dry-run`, `--undo <batch-id>` (repeatable), `--delete-dups-from-source-server`. |
| `msgvault delete-deduped` | Local hard-delete rung. `--batch <id>` (repeatable) or `--all-hidden`. VACUUM INTO backup before deletion; interactive confirmation by default. |
| `msgvault identity {list,show,add,remove}` | Per-account identity management. `list` accepts `--account`/`--collection`/`--json`. Replaces the removed `list-identities` command. |
| `msgvault collection {list,create,add,remove,delete,show}` | User-defined groupings of accounts. The `All` collection is auto-managed and rejects explicit mutation with `ErrCollectionImmutable`. |
| `msgvault search --collection <name>` | Collection scope on FTS, vector, and hybrid search modes. `--account` and `--collection` are mutually exclusive. |
| `msgvault stats --account/--collection` | Per-scope stats. |
| `msgvault delete-staged --permanent` | Opt into batch permanent deletion. Default is Gmail trash (~30-day recovery). Mutually exclusive with `--yes`; `--permanent` requires typing `delete` to confirm. Execution is gated by `MSGVAULT_ENABLE_REMOTE_DELETE=1` for the v1 release. |

## Key semantics

- **One account = one source.** `--account` resolves an account/source identifier; collection names supplied to `--account` are rejected with a hint to use `--collection`. `--collection` resolves only collections; account names supplied to `--collection` are rejected symmetrically.
- **Cross-source dedup is opt-in via `--collection`.** Outside collection scope, dedup never crosses source boundaries — protects sent-message provenance (alice's Sent and bob's Inbox share an RFC822 Message-ID and must both survive).
- **Remote deletion is same-source-only**, even under collection scope. Cross-account dedup hides locally; it never stages a remote delete that crosses sources.
- **Survivor selection** has source-type preference (default `gmail,imap,mbox,emlx,hey`) with tiebreakers on raw MIME presence, label count, archive time, and id. Sent-copy eligibility filter runs before survivor selection.
- **Content-hash fallback** is supplementary, not transitive. Groups with multiple Message-ID survivors are skipped; groups containing both an MID survivor and a sent-copy orphan are skipped (per spec § Detection / Survivor selection).

## Identity model

- Confirmed identities live in `account_identities (source_id, address, source_signal, confirmed_at)` with multi-signal set semantics (signals stored as comma-delimited sorted set).
- Case is preserved on insert because the column accommodates phone E.164 and synthetic handles; comparison uses `identifierMatch` / `NormalizeIdentifierForCompare` which case-folds email-shaped values only.
- A one-time legacy migration copies any `[identity].addresses` config block into per-account records. If no eligible source exists at startup, the migration is deferred until first source creation. Apple-Mail and other non-email source types are filtered.
- Auto-confirm runs at source creation for `add-account`, `add-imap`, `add-o365`, `import-mbox`, `import-emlx`, `import-whatsapp`, and `import-gvoice`. `import-imessage` is exempted (iMessage contacts are not self-identifying).
- `confirmDefaultIdentity` is the shared helper; the prompt is routed through `cmd.OutOrStdout()` so test harnesses can capture it.

## Live-message contract

`internal/store.LiveMessagesWhere(alias, hideDeletedFromSource)` is the
canonical SQL fragment. It always filters `deleted_at IS NULL` (dedup
losers are hidden everywhere) and gates `deleted_from_source_at IS NULL`
behind the boolean (archive views may show source-deleted rows by
design). All read paths route through it: `internal/store` API/list/
search/summary, `internal/query` SQLite + DuckDB engines (including
text search and the raw-MIME helper), `internal/vector/sqlitevec`
backend and fused-search filter, dedup engine, and stats.

## Schema changes

- New tables: `collections`, `collection_sources`, `account_identities`, `applied_migrations`, plus dedup operation tracking.
- New column: `messages.deleted_at` (added via `ALTER TABLE` migration on schema upgrade).
- `internal/store/migrations.go` introduces a forward-only migration runner with `MarkMigrationApplied`.

## Safety guardrails

- `delete-staged` defaults to trash; `--permanent` is opt-in, mutually exclusive with `--yes`, and requires typing the literal word `delete`.
- `delete-deduped --all-hidden` always prompts even when `--yes` is set.
- `delete-staged` execution is gated by `MSGVAULT_ENABLE_REMOTE_DELETE=1`. Read-only modes (`--list`, `--dry-run`, `show-deletion`) work without the gate.
- `deduplicate --dry-run` and `--undo` are mutually exclusive (undo writes; dry-run promises no writes).
- Cancelled deletion manifests are persisted in their own `cancelled/` directory rather than removed, so audit history survives.
- Database backup before any local hard delete or merge (skippable with `--no-backup`).

## TUI cache invalidation

`tui.go` cache-staleness check counts dedup hides (`deleted_at >=
LastSyncAt AND deleted_from_source_at IS NULL`) as a third signal
alongside new messages and source-deletions, forcing a full Parquet
rebuild when present. Counts are kept disjoint from source-deletion
counts to avoid double-attribution.

## Documentation

- [`docs/accounts-identities-collections-dedup/spec.md`](docs/accounts-identities-collections-dedup/spec.md) — full specification (scope semantics, identity model, dedup detection/survivor selection, safety progression, CLI surface).
- [`docs/accounts-identities-collections-dedup/README.md`](docs/accounts-identities-collections-dedup/README.md) — narrative overview with concept diagrams.

## Notable behavior changes

- `list-identities` removed; use `identity list`.
- `[identity].addresses` in `config.toml` becomes legacy input after the one-time migration — write per-account identities instead.
- Remote-source eligibility for staged deletion is currently `gmail` only. The manifest format (`GmailIDs`) and executor (`gmail.API`) are Gmail-specific; IMAP is gated until a manifest format that records source type and an IMAP executor exist.
- Collection identity is derived as the union of member-account confirmed identities; not configured separately.
- After a large `delete-deduped` run, vector and Parquet caches may hold stale entries — the command recommends `build-cache --full-rebuild`.


jesserobbins:
Addressed in `d3cdb9c`. Pushing back on the HIGH and confirming each Medium / Low.

## HIGH — pushing back

> **Cross-account collection dedup can hide legitimate messages.**
> `--collection` resolves all collection member source IDs into one dedup scan. For mixed-account collections such as the default `All`, duplicate groups can span different accounts, despite the dedup contract that cross-account messages must never be merged.

Respectfully disagree — re-reading the design alignment document, cross-account dedup under `--collection` is explicit user opt-in, not a contract violation. From the [design alignment spec](https://github.com/wesm/msgvault/blob/main/docs/superpowers/specs/2026-04-29-pr-286-design-alignment.md) (and this PR's commit messages):

- `deduplicate --collection <name>`: "Compare messages across member accounts in that collection."
- "A user who genuinely wants to dedup across every account can still write `--collection All`."

The same-source-only constraint is on **remote deletion staging**, not on local dedup grouping. The contract is preserved at `internal/dedup/dedup.go:738`: a remote-deletion entry is only staged when `m.SourceID == survivor.SourceID`. Cross-account losers under `--collection` get **soft-deleted locally** (reversible via `--undo`), but **never staged for remote deletion**.

The default unscoped `deduplicate` (no flags) iterates per-account and never crosses source boundaries. Users have to explicitly write `--collection` to opt into cross-account dedup; the spec covers this trade-off as the user's deliberate choice. If the safety language wants strengthening, that's a doc question, not a code change.

If we want a regression test that demonstrates the contract: two accounts, same `Message-ID`, run `deduplicate` (no flags). Both messages remain live. Run `deduplicate --collection All`. One is locally hidden. `--undo` restores it. No remote-deletion entry is staged. I can add that test if it helps.

## Mediums — fixed

- **Content-hash fallback misses matches against Message-ID groups.** `excludeIDs` now contains only losers, not survivors. A Message-ID-less orphan can now link to a Message-ID survivor via content hash. To prevent demoting a Message-ID survivor (which has already absorbed labels from its losers) into a content-hash loser, the content-hash pass post-processes its groups: any Message-ID survivor that appears as a non-survivor in a content-hash group is forced back to survivor position.
- **Dry-run undo still performs the undo.** `--dry-run` and `--undo` are now `MarkFlagsMutuallyExclusive` at the cobra layer.
- **Legacy identity migration can mark itself applied too early.** The migration no longer marks itself applied when `[identity].addresses` is set but zero sources exist. Marking in that state would permanently drop the addresses; instead the migration runs on the next command after a source is added.

## Lows from job 378 — also fixed

- **`list-identities --confirmed` with no scope** now defaults to the union across every source. `GetIdentitiesForScope`'s "explicit empty scope means no match" contract is intact; the CLI explicitly expands "no flag" to "all sources."
- **`MergeFilterIntoQuery` ignored `filter.SourceIDs`.** Now forwards multi-source into `merged.AccountIDs`, preferring `SourceIDs` when both are set. Latent (no current caller sets it), but defensive against future callers.
- **`dedup-purge` in remote mode** now rejects upfront with a clearer message rather than the generic "must specify --batch or --all-hidden."

`go test ./...` clean.

jesserobbins:
@wesm i will mark ready once I've done my own acceptance tests. however i wanted to get you this once it was ready for your reviews

jesserobbins:
Documentation follow-up to the H1 cross-account question — `0ea4332`.

The behavior is unchanged (cross-account local-hide under `--collection` is the design contract) but the CLI help didn't make that visible to the user. Now it does:

- **`deduplicate` Long help** has a new Scope section that explicitly states each flag's source-boundary semantics. `--account` never crosses sources. `--collection` is the only way to compare across sources, is an explicit opt-in, hides losers locally (reversible via `--undo`), and `--delete-dups-from-source-server` stays same-source-only.
- **`--account` and `--collection` short descriptions** clarified to match — "scope to one account; never crosses source boundaries" vs "opts into cross-source comparison."
- **Runtime banner**: when a collection-scoped run spans more than one account, prints a one-line note that cross-source dedup is reversible and remote deletion stays same-source-only, plus a reminder that `--dry-run` previews without writing.

No engine change. The reviewer's concern was real on the user-education axis even if the underlying behavior matches the spec.

jesserobbins:
Confirmed that this is working as designed. However, I am not sure about right syntax for the new commands. For instance, collection vs collections. However, this works, builds and reviews clean so fine for testing.

jesserobbins:
> ## roborev: Combined Review (`04dc1c7`)
> Potentially unsafe PR: one Medium security finding needs attention before merge.
> 
> ## Medium
> * [internal/dedup/dedup.go:~100, ~720](/home/roborev/repos/msgvault/internal/dedup/dedup.go): `remoteSourceTypes` includes `"imap": true`, and `Execute` stages `m.SourceMessageID` into deletion manifests for any remote source type. Current `delete-staged` execution uses the Gmail client and manifest `GmailIDs`, keyed only by `Filters.Account`, so IMAP duplicates staged via `deduplicate --delete-dups-from-source-server` can later be executed against Gmail as if their IMAP message IDs were Gmail IDs.
>   Exploit path: a malicious or compromised IMAP server could control IMAP `source_message_id` values for an account that also has Gmail OAuth configured. If the user runs dedup with remote deletion enabled and then `delete-staged`, crafted IMAP IDs matching Gmail message IDs could cause unintended Gmail trash/delete operations.
>   Remediation: remove `"imap"` from `remoteSourceTypes` until IMAP deletion execution exists, or extend manifests with a required `source_type` and make `delete-staged` dispatch by source type, rejecting non-Gmail manifests in the current Gmail-only executor.
> 

This is helpful feedback and useful for future improvements, but the proposed solution breaks the current functionality and is out of scope.

jesserobbins:
While it was something originally discussed in the project scope - I believe it is best to gate the gmail remote delete gated. We can also remove it. Basically everything else we are doing right now (local hard-delete, dedup, even import) is reversible or local. Remote-delete touches the user's authoritative Gmail mailbox and I just don't feel comfortable with that going out in this at all. Very open to feedback on this and everything else, however the two commits above put it behind an ENV flag.

jesserobbins:
Addressed in eb7aaaf.

**Finding 2 (collection methodology) — fixed.** Added `Config.ScopeIsCollection`, plumbed from the `--collection` branch in `cmd/msgvault/cmd/deduplicate.go`. `FormatMethodology` now branches: collection mode prints that cross-account merging is intentional and reversible via `--undo`; account/per-source mode keeps the original "NEVER merges" wording.

**Finding 1 (live-message filtering) — partial.** The prescribed fix would have changed product behavior. `HideDeletedFromSource` is opt-in by design:

- TUI exposes `hideDeletedFromSource` as a user-toggleable filter, default off (`internal/tui/keys.go:1008`, `internal/tui/view.go:1301`).
- `internal/api/handlers.go:1035,1094,1362` explicitly set it to `true` on paths that want hiding — those would be dead code if the engine already filtered.
- msgvault is an archive whose value prop is keeping source-deleted rows visible.

The accurate observation was that `LiveMessagesWhere`'s docstring (introduced in 5188612) overclaimed a contract the query layer never honored — the `internal/query` package never imports the helper. Narrowed the comment in `internal/store/live_messages.go` to describe the predicate for store/vector callers and explicitly note that `internal/query` intentionally applies only `deleted_at` by default and gates `deleted_from_source_at` behind `HideDeletedFromSource`.

— Claude (Opus 4.7)

jesserobbins:
Pushed `8cb9c60` for the four findings still open from the `f0ab152` and `5188612` combined reviews. Bundled into one commit because they're small and independent:

1. **`--undo` mutual exclusion (job #410 medium).** Cobra's existing exclusivity covered `--account` / `--collection` / `--dry-run` but not the destructive/scan flags. Added `MarkFlagsMutuallyExclusive("undo", X)` for `--delete-dups-from-source-server`, `--prefer`, `--content-hash`, `--no-backup`, `--yes` so `msgvault deduplicate --undo X --delete-dups-from-source-server` now errors at parse time instead of silently ignoring the destructive flag.

2. **`resolveAccountList` E.164 (`f0ab152` medium).** `strconv.ParseInt` accepts a leading `+`, so `+15551234567` parsed as the integer `15551234567` and was treated as a source ID. Gated the numeric branch on `p[0]` being a decimal digit — signed inputs fall through to identifier resolution, restoring the path for WhatsApp / Google Voice accounts.

3. **`confirmDefaultIdentity` rebind guard (`f0ab152` medium).** Pushed the guard down into `confirmDefaultIdentity` itself rather than threading a `created` flag through 9 ingest call sites: skip the write when the source already has any identity rows. Preserves the documented "freshly created source" intent. Degrades gracefully if the user has removed every identity — in which case the default is restored, which seems right on a clean slate.

4. **Deferred-migration notice count (`5188612` low).** Notice now reports the post-normalization count instead of `len(rawAddresses)`, so a config like `["", "  ", "alice@x"]` no longer announces "3 address(es) parked" when only one is. Test updated.

Verified with the built binary: `--undo` + `--delete-dups-from-source-server` and `--undo` + `--content-hash` both reject. Full test suite green locally.

The CSV-fallback I/O note (also `5188612` low) is intentionally left alone — reverting the inner `deleted_at IS NULL` filter would re-introduce the schema drift between sqlite_scanner and CSV-view paths that broke Windows in `f0ab152`. Rationale is in that commit body.

jesserobbins:
Addressed the medium on `8cb9c60` in `c90fb73`, plus the low cosmetic on `8cb9c60` in `82606f4`.

**Medium (`engine.Scan` mutates pre-confirmation/pre-backup) — addressed by surfacing, not by restructuring.**

The reviewer is right that Scan calls `BackfillRFC822IDs` outside dry-run mode and that this happens before the user prompt and before the backup. I think the underlying behavior is correct and shouldn't change:

- Backfill is **idempotent metadata derivation** — it fills a previously-NULL `rfc822_message_id` from MIME data already on the row. It never overwrites a non-NULL value, never changes message content, and is recoverable from `message_bodies` if anyone ever wanted to recompute.
- The dedup-batch backup-before-merge contract is sized for the **merge** (which soft-deletes losers and transfers labels), not for metadata catch-up.
- Restructuring Scan to be read-only would force one of: (a) double-pass backfill (waste), (b) a report that under-reports duplicates the backfill would have surfaced (worse than the status quo for the user), or (c) a non-trivial restructure of the Scan/Execute split. None of those buys safety against a meaningful failure mode.

What was actually missing is **consent legibility**: the prompt said "hide N duplicates (reversible with --undo)" without telling the user that scan already wrote N′ rfc822 values that stay regardless of their answer. `FormatReport` already prints the backfill count, so the user sees it in the scan output, but the prompt didn't scope itself.

`c90fb73` adds:
- A one-line note in both confirm prompts (single-source `runDeduplicateOnce` and per-source `runDeduplicatePerSource`) when `BackfilledCount > 0`, calling out that scan already performed the backfill and it's kept regardless of the answer.
- An expanded godoc on `dedup.Engine.Scan` documenting the side effect, why it lives in scan rather than execute, and why the backup contract is unchanged.

No behavior change.

**Low (redundant `len(p) > 0` in `collection.go`) — fixed in `82606f4`.** The surrounding loop already trims and continues on empty tokens, so `p[0]` indexing is safe.

— Claude Code (Opus 4.7) and Primeradiant Superpowers

jesserobbins:
This was, in the end, a lot more work than I was expecting.

jesserobbins:
For anyone wanting to test #304 alongside the Facebook Messenger importer (#281): I've rebased fbmessenger on top of this PR at `jesserobbins/msgvault:jesse/fbmessenger-post-304`. It includes the `import-messenger` → `confirmDefaultIdentity` wiring (signal `account-identifier`) so messenger sources participate in the new identity model the same way the other importers do.

wesm:
*via Codex-assisted review*

Thanks for the detailed writeup and for moving this branch much closer to the account/source + collection model we discussed. The explicit `--account` / `--collection` split, canonical schema ownership for collections/identities, per-account identity records, same-source-only remote deletion staging, and the `MSGVAULT_ENABLE_REMOTE_DELETE=1` gate are all good moves.

I still think there are a few issues to fix or explicitly settle before merge.

## Pre-merge fixes

1. **Vector/hybrid search ignores `--collection`.**

   In `cmd/msgvault/cmd/search.go`, non-FTS modes dispatch to `runHybridSearch` before the local search scope resolver runs. Then `cmd/msgvault/cmd/search_vector.go` only handles `searchAccount`; `searchCollection` is never applied to `filter.SourceIDs`.

   That means a command like:

   ```sh
   msgvault search --collection work --mode=hybrid ...
   ```

   can search the whole archive instead of the collection. I think the right shape is to resolve scope once before the search-mode branch, then pass the resolved `SourceIDs` into FTS, vector, and hybrid search. Please also add tests covering `--account` and `--collection` for FTS and vector/hybrid where build tags permit.

2. **Dedup-hidden rows still leak through several read paths.**

   The branch establishes `messages.deleted_at` as the local dedup hide flag, but some store/API paths still filter only `deleted_from_source_at IS NULL`.

   Places to audit include:

   - `internal/store/api.go`
   - `internal/store/messages.go`
   - `internal/store/subset.go`
   - `internal/query/sqlite.go:GetMessageSummariesByIDs`

   This is the biggest semantic risk in the branch: after dedup, hidden losers should not appear in normal user-facing reads. I’d like this fixed through a shared predicate/helper rather than more hand-copied SQL fragments. We should also be explicit about the distinction between “dedup-hidden locally” and “deleted from the source server”; archive views may intentionally show source-deleted rows, but they should still hide dedup losers.

3. **TUI/DuckDB cache staleness does not account for dedup hiding.**

   `build-cache` filters `deleted_at IS NULL` when building Parquet, but the TUI cache staleness check appears to detect new messages and `deleted_from_source_at` changes, not `deleted_at` changes. If a user dedups after the cache is built, the TUI/DuckDB surface can continue showing duplicate rows until a full rebuild.

   Either the TUI cache needs a `deleted_at`/dedup-batch staleness signal, or the DuckDB path needs to filter against current SQLite visibility. The product contract should not rely on a user remembering to rebuild the cache for hidden duplicates to disappear.

4. **Dedup terminology still mixes the old and new account model.**

   The implementation now mostly treats `--account` as one source/account and `--collection` as the explicit cross-source boundary, which is the model I want. Some comments/help text still describe account-scoped dedup as cross-source dedup across “Gmail sync + mbox export of the same mailbox.”

   Please update the package comments, command help, and methodology output so the terms are consistent:

   - account = one ingest source/archive
   - collection = named grouping of accounts/sources
   - cross-source dedup requires `--collection`

5. **Dry-run/report wording mentions a nonexistent `--apply` flag.**

   The report says to rerun with `--apply`, but there is no such flag. Please change that to the actual invocation semantics.

## Product decision to settle

`delete-staged` currently defaults to permanent Gmail deletion once `MSGVAULT_ENABLE_REMOTE_DELETE=1` is set, with `--trash` as the recoverable option. The design text we had been aligning around said trash-by-default and permanent-delete opt-in.

I’m not saying this has to be changed in this PR if we explicitly decide otherwise, but I do want the safety model settled before merge. My preference is:

```sh
msgvault delete-staged             # trash/recoverable default
msgvault delete-staged --permanent # explicit permanent deletion
```

The env var gate is a useful release guardrail, but I don’t think it fully replaces making the destructive mode explicit at the command level.

## Non-blocking / follow-up notes

- The identity storage currently preserves case and keys by exact `address`. That may be intentional because the column can hold phones and chat handles, but the PR description says “upsert by normalized identifier, preserve original case.” If that is still the intended behavior, I would expect separate stored/display and normalized keys. Otherwise the PR text should be adjusted.
- The content-hash fallback is not a true transitive union of all duplicate signals. That may be fine if the intended policy is “Message-ID is primary, content hash is supplementary,” but the methodology output should say that clearly.
- The TUI does not appear to expose collections as a first-class scope yet. That is okay if deferred, but it should be called out as not part of this PR’s shipped behavior.

Overall, I think the model is much closer now, but scope and visibility are still duplicated across too many read surfaces. The main thing I want before merge is one authoritative scope resolution path and one authoritative visibility predicate used consistently by FTS, vector/hybrid, API, MCP, TUI/DuckDB, and stats.

jesserobbins:
@wesm - thanks for the detailed pass. All the items you raised have hopefully been addressed. Especially the live view, and for the Gmail delete options, the destructive call is gated behind a flag and the safe move-to-trash behavior is the default for those that want that.

Pointers below. The PR body has been refreshed in full to match the current branch state.

## Live view

A row is "live" along two independent axes:

- **Dedup-hidden** (`deleted_at`). A dedup loser. Filtered **always** on every normal read path. There is no opt-in to show dedup losers. They never appear in search, browse, stats, API, MCP, vector, hybrid, or exports.
- **Source-deleted** (`deleted_from_source_at`). A row whose source-server message was deleted (Gmail trash, IMAP expunge). **Caller-controlled.** msgvault is an archive, so source-deleted rows stay in the archive and remain visible by default. Callers that want a live-only view set `HideDeletedFromSource = true` on the request shape they already have. The TUI exposes this as a user toggle. API and MCP handlers explicitly opt in where the consumer wants it.

One predicate enforces both axes: `internal/store.LiveMessagesWhere(alias, hideDeletedFromSource)`. Every read path in `internal/store`, `internal/dedup`, `internal/vector`, and `internal/query` calls it, and `internal/store/`, `internal/vector/`, `internal/dedup/` returns only the helper definition, the new test, and two comments.

The two-axis model means dedup correctness is enforced unconditionally and the archive's "show me what was on the server" semantics are preserved without leaking.

## The five pre-merge items

**1. Vector/hybrid search ignores `--collection`.** Fixed in `3a83723` (`search: resolve --account/--collection before mode dispatch`). `resolveSearchScope` runs once before the FTS / vector / hybrid branch. The vector path applies `scope.SourceIDs()` to `filter.SourceIDs`. Tests added for `--account` and `--collection` on FTS, and on vector under the `sqlite_vec` build tag.

One follow-on caught after that landed. `runHybridSearch` opens its own raw `*sql.DB` via `sql.Open`, bypassing `store.InitSchema`. For scoped queries this was masked because scope resolution ran `InitSchema` first. For unscoped queries the dispatch went straight to the raw open. On a database upgraded from a pre-#304 build the vector backend's `LiveMessagesWhere` filter would hit "no such column: deleted_at" until something else triggered schema init. Hoisted the init into the unscoped branch in `77e15b8` (`search: run schema init on unscoped vector/hybrid dispatch`). Regression test drops `deleted_at` and asserts unscoped vector dispatch re-adds it.

**2. Dedup-hidden rows leak through several read paths.** Fixed in `35414b5` (`query+store: centralize live-message predicate via parameterized LiveMessagesWhere`). See "Live view" above for the model. Specifics:

- The four inline raw-SQL fragments in `internal/vector/sqlitevec` (`backend.go:307,818,847`, `fused.go:132`) interpolate the helper.
- The ~13 hand-copied `deleted_at IS NULL` / opt-in `deleted_from_source_at IS NULL` pairs in `internal/query/sqlite.go` and `internal/query/duckdb.go` collapse into single helper calls.
- The hydration paths (`sqlite.go:744`, `sqlite_text.go:451`, `duckdb_text.go:436`) interpolate the helper with `true`.
- The Gmail-staging paths (`sqlite.go:1007`, `duckdb.go:1670`) collapse to one helper call passing literal `true`. Staging never honors an opt-in.
- `SubAggregate`'s redundant double-handling of the same flag reconciles by OR-ing `opts` into `filter` before `buildFilterConditions`.

New table-driven `TestLiveMessagesWhere` in `internal/store/live_messages_test.go` covers the constant fast-paths and `fmt.Sprintf` fallback at both boolean values across three aliases.

**3. TUI/DuckDB cache staleness on dedup.** Fixed in `8d20b9a` (`tui: invalidate parquet cache on dedup hides since last sync`) and `6c8e351` (`tui+store: keep cache reasons disjoint; pre-compute live-messages predicate`). `cacheNeedsBuild` signals on dedup-hide (`deleted_at` since `LastSyncAt`) and source-delete (`deleted_from_source_at` since `LastSyncAt`) independently. The dedup-hide branch adds `AND deleted_from_source_at IS NULL` so a row that's both source-deleted and dedup-hidden after `LastSyncAt` reports once (as a deletion), not twice in the reason string. Both still trigger the same `FullRebuild=true`. Cache reason strings stay disjoint.

**4. Dedup terminology mixes the old and new account model.** Fixed in `d123e3a` (`dedup: align terminology with one-source = one account; drop --apply hint`). Package comments, command help, and `FormatMethodology` output now consistently use:

- `account = one ingest source/archive`
- `collection = named grouping of accounts/sources`
- cross-source dedup requires `--collection`

Earlier in the review loop, `8859fcc` (`dedup: scope-aware methodology and narrow live-message contract`) had already made `FormatMethodology` branch on collection vs account scope. `d123e3a` finished the job across the rest of the surface.

**5. `--apply` hint references a nonexistent flag.** Fixed in the same `d123e3a` commit. The dry-run report points at the actual invocation (`--yes` for non-interactive, no flag for interactive), not a flag that never existed.

## Product decision: trash-by-default, `--permanent` opt-in

Fixed to reflect the original intent in `bf02c6b` (`delete-staged: default to trash; --permanent opts into batch delete`). The two-axis safety model:

```sh
msgvault delete-staged                                                  # gated; needs MSGVAULT_ENABLE_REMOTE_DELETE=1
MSGVAULT_ENABLE_REMOTE_DELETE=1 msgvault delete-staged                   # trash (recoverable for 30 days)
MSGVAULT_ENABLE_REMOTE_DELETE=1 msgvault delete-staged --permanent       # batch-delete (permanent)
msgvault delete-staged --list                                            # always allowed
msgvault delete-staged --dry-run                                         # always allowed
```

The env var stays as a release guardrail on top of the explicit `--permanent` flag. They cover different failure modes. The env var prevents accidental Gmail-API invocation in scripts. `--permanent` prevents accidental permanent deletion when the user did mean to invoke the API. The dedup post-run hint, TUI staging modal, MCP `stage_for_deletion` next-step hint, and quickstart docs all reference the gated invocation.

## Other changes worth pointing at

Several review iterations surfaced real bugs in this PR's own new code that didn't show up in your review but are worth listing.

- **Dedup correctness/safety pass** in `fdab4fc` (`dedup: correctness and safety fixes across remote-delete, cache, undo, and migration`). Gates remote-delete to Gmail only (drops `imap` from `remoteSourceTypes`; the staged-deletion manifest format is Gmail-specific so IMAP staging would have misrouted). Excludes dedup-hidden rows from the analytics cache export (the primary `sqlite_scanner` path was missing the `deleted_at IS NULL` filter the CSV-fallback path applied; Linux/macOS full rebuilds could surface deleted rows in analytics). Routes `--undo` before scope resolution so a stale or renamed account can't block undoing a valid batch, and aggregates undo errors across batches instead of stopping at the first failure. Adds a deferred return value to `MigrateLegacyIdentityConfig` so the "addresses configured but no sources yet" case is distinguishable from "already applied" and "nothing to migrate."
- **Identifier comparison consolidation.** Email-vs-other comparison was originally a heuristic repeated across three call sites. Two iter regressions traced back to that shape (add/remove asymmetry; Matrix MXID misclassification). Consolidated into `internal/store.identifierMatch` and routed through `AddAccountIdentity`, `RemoveAccountIdentity`, `MigrateLegacyIdentityConfig`, the dedup engine's per-source identity map (`NormalizeIdentifierForCompare`), and the CLI prior-row lookup (`EqualIdentifier`). Email-shaped identifiers compare case-insensitively. Matrix MXIDs and chat handles preserve case. The schema column is documented as case-preserved.
- **Deferred-then-applied legacy `[identity]` migration.** `runStartupMigrations` ran *before* `GetOrCreateSource` in every ingest command. `[identity]` addresses configured before the first source was created would park on the deferred path and only apply on the *next* command. Fixed by re-running the migration after first source creation in every ingest path. The pre-create pass now skips itself when no source exists yet, so a fresh-install user gets one accurate "applied" notice instead of "will run on next command" followed by "applied" in the same invocation. CLI-level regression test exercises the deferred-then-applied flow within a single `add-account` invocation.
- **`Config.DatabasePath()` percent-encoding.** Decoded percent-encoding for absolute `file:` URIs but not for relative ones (`url.Parse` leaves `u.Opaque` encoded). `file:my%20vault.db` would never resolve to the actual on-disk filename. Fixed with `url.PathUnescape` and a regression test.
- **`openStoreAndInit` fresh-install regression.** A guard had picked up that errored "database file not found, run init-db first." It broke fresh-install for 15 commands (every dedup, identity, collection, delete-deduped, and import-gvoice/imessage path). Removed; `init-db` remains the documented setup command, other commands fall through to `store.Open`'s create-on-first-use behavior.
- **Smaller fixes.** `splitSignalSet` filters empty parts from stray-comma legacy data so JSON output and `slices.Contains` checks match `mergeSignalSet`'s producer-side normalization. The numeric-source-id branch in `resolveAccountList` is gated on a leading-digit check so E.164 phone identifiers (`+15551234567`) fall through to identifier resolution instead of being parsed as integer 15551234567. `confirmDefaultIdentity` skips when the source already has any identity rows, so an identifier the user explicitly removed via `identity remove` isn't re-added by the next ingest re-run.

## Excluded or rejected from the polishing loop

Worth being explicit about what we deliberately did not change, and why. Each one was raised at least once during the review iterations.

- **No `UNIQUE INDEX ON (source_id, LOWER(address))` on `account_identities`.** Two reviewers flagged it. The table is brand-new on this branch — `main` has never shipped it — so the "case-variant duplicates can already exist" failure mode the index would prevent has no historical state to apply to. Single-writer SQLite plus the existing SELECT-then-INSERT transaction enforce uniqueness in code for the deployment we ship. Adding the index now also conflicts with a deferred typed-kinds proposal that would replace the PK shape entirely.
- **`VACUUM INTO ?` parameter binding stays.** A reviewer flagged it as a runtime failure (SQLite's grammar wants a string literal). Empirical test under `mattn/go-sqlite3` with production-shape DSN (WAL pages forced via 100 inserts), three destination paths (simple, embedded space, embedded single quote): all three succeed and create the destination file. The driver translates the bound parameter correctly even though the grammar spec expects a literal. Worth re-checking if the driver ever changes.
- **Cross-account dedup under `--collection` is opt-in, not a bug.** A reviewer flagged that `deduplicate --collection work` could "hide legitimate messages" across accounts. That's exactly what the spec asks for: `--collection` is the explicit user signal that cross-source comparison is wanted. The same-source contract applies to *remote deletion* (verified at `internal/dedup/dedup.go` where the staging path skips cross-source pairs), not to local hide. Methodology output (`8859fcc`, then finished in `d123e3a`) describes scope-aware behavior so the user reading the dry-run knows what's happening.
- **Scan's RFC822 backfill stays in scan, not in execute.** A reviewer flagged that `Engine.Scan` writes `rfc822_message_id` values before the user confirmation and the dedup-merge backup. The behavior is correct: backfill is idempotent metadata derivation (fills a previously-NULL column from MIME data already on the row, never overwrites a non-NULL value, changes no message content), and the backup contract exists for the dedup merge (which soft-deletes losers and transfers labels), not for this metadata catch-up. Restructuring `Scan` to be read-only would force either a double-pass backfill or a report that under-reports duplicates that backfill would have surfaced. Fix landed as consent legibility instead, in `e9182d2`: the confirm prompt now mentions backfill when `BackfilledCount > 0`, and the godoc on `Engine.Scan` documents the side effect.
- **Content-hash is supplementary, not a transitive union.** Methodology output now states clearly that Message-ID is primary and content-hash is supplementary fallback, not a transitive union of all duplicate signals. The cross-pass survivor-collision case (a content-hash group with two Message-ID survivors where one absorbed labels in the Message-ID pass) is filed as a follow-up against `main`. It's pre-existing and not introduced by this PR.

## Other notes you flagged as non-blocking

- **Identity case/normalize keys.** Covered in "Other changes" above. The case-aware rule is consolidated through one helper. Email-shaped → case-insensitive; Matrix MXIDs / chat handles → case-preserved. Schema column documented as case-preserved. The deferred typed-kinds proposal you've already seen would split stored / display / normalized keys formally; what shipped is the minimum-helper consolidation that closes the structural risk without committing to the larger redesign.
- **TUI doesn't expose collections as first-class scope.** Acknowledged. The CLI surface (search, stats, dedup, identity) does, so the data model is fully wired. Only the TUI drill-down UI lags. Called out in the PR body's Notes.

## A note on the branch shape

The branch was rebased from 63 commits down to 57. Dropped a flip-flop pair on import-emlx/mbox migration-error UX, dropped a no-op add-then-remove of an internal review-notes file, and folded a roborev fix-up commit into its parent. The pre-rebase history is preserved at `origin/jesse/identities-collections-dedup-presquash-20260502` if you want to compare against an old SHA. Tree contents at HEAD are identical. Only history shape changed.

One follow-up issue was filed against `main` while addressing the latest review batch: #307 (`import-whatsapp: auto-default-identity skipped when summary.Errors > 0`). Pre-existing on main. Doesn't touch code this PR modifies

jesserobbins:
## Branch review summary — `jesse/identities-collections-dedup`

**Process:** Ran two parallel roborev branch reviews (claude-code + codex), investigated every finding against the actual code, applied one narrow fix, and reran both reviews against the updated branch.

### Round 1 findings

**Codex (job 535)** flagged 2 issues. **Claude-code (job 536)** flagged 8. After investigation:

| # | Finding | Agent | Verdict | Reason |
|---|---------|-------|---------|--------|
| 1 | `confirmDefaultIdentity` early-return leaves account-identifier missing on rerun/migration paths | both | **Reject** | Both reviewers missed the explicit ordering doc (`account_identity.go:30-38`) and call-site comments. Commit `99a02a5` already fixed exactly this regression; all four ingest paths run the helper before legacy migration. The `len(existing) > 0` guard is intentional — it preserves user `identity remove` actions. |
| 2 | `MergeFilterIntoQuery` collapses empty `AccountIDs` to nil, losing match-nothing sentinel | codex | **Reject** | The override path (`sqlite.go:1449-1454`) correctly uses `make`/`copy`. No code anywhere produces empty-non-nil `q.AccountIDs`, so the line-1440 collapse can't trigger the failure mode. |
| 3 | Content-hash multi-survivor `break` causes data loss | claude | **Reject** | `MergeDuplicates` always unions losers' labels into the survivor before soft-delete. The author's own protective comment was overcautious — labels are merged, not destroyed. |
| 4 | Raw-MIME backfill is order-dependent | claude | **Doc-only** | Survivor selection already prefers messages with raw MIME, so backfill rarely runs. Cosmetic. |
| 5 | `BackfillRFC822IDs` does N+1 SELECTs per batch | claude | **Valid → fixed** | Real perf issue on a hot path. `StreamMessageRaw` already exists and was the right tool. **Fixed in `affea76`.** |
| 6 | `EnsureDefaultCollection` ignores `LastInsertId` error | claude | **Reject** | Defensive hardening; not realistically failable for sqlite3. |
| 7 | `DeleteAllDeduped` count-vs-delete predicate mismatch | claude | **Reject** | Doc on `dedup.go:386-406` explicitly addresses this trade-off and notes it's safe + future-proofed. |
| 8 | `AddAccountIdentity` empty-signal docstring nuance | claude | **Reject** | Cosmetic. |
| 9 | Default-collection self-heal doesn't run in serve mode | claude | **Reject** | Doc on `sync.go:326-334` explicitly explains and accepts the trade-off for a single-user tool. |

### The one fix — `affea76`

`store: stream raw MIME in BackfillRFC822IDs to remove per-row SELECTs`

Replaced the per-row `s.GetMessageRaw(id)` call inside the batch loop with a single `s.StreamMessageRaw(batchIDs, ...)` per batch, decompressing zlib inline in the callback. The UPDATEs were already batched in one tx, so left untouched. On the dedup hot path (batchSize=1000) this drops 1000 round trips per batch to roughly 2 chunked queries. Tests pass.

### Round 2 findings

**Claude-code (job 539):** No issues found. Pass.

**Codex (job 538)** flagged 2 new Mediums:

| # | Finding | Verdict | Reason |
|---|---------|---------|--------|
| 1 | Empty `--collection` leaks into unscoped search | **Reject** | False positive. `search.go:198-200` already rejects zero-member collections with literally the error codex's "fix" recommends. Codex didn't read `resolveSearchScope`. |
| 2 | New backfill helper treats unknown compression values as uncompressed | **Reject as regression** | True observation but not introduced by this branch. Only `'zlib'` is ever written anywhere in the codebase, and `GetMessageRaw` (pre-existing canonical reader at `messages.go:303`) has identical pass-through behavior. The new helper deliberately matches the established convention. A shared decompress helper that errors on unknown values would be a reasonable future hardening but touches pre-branch code paths and is out of scope here. |

### Recommendation

**Ready to merge.** Of 12 distinct findings across both rounds and both agents, exactly one was a real bug worth fixing in this branch (the N+1, now landed). The remaining 11 broke down as: 4 missed explicit doc/comment blocks the author had already written, 3 documented design trade-offs, 2 cosmetic, 1 theoretical with no real call path, and 1 pre-existing pattern outside this branch's scope.

The high false-positive rate on round 1 reflects a pattern worth noting for future reviews: both agents flagged behaviors the author had already considered and explicitly documented inline. Reading the surrounding doc comments before writing findings would catch most of these.

jesserobbins:
*SIGH* This is why I didn't want to push the spec yet. I have another fix ready for these issues but don't want to actually touch the codebase.

jesserobbins:
## Roborev `4e72aa9` findings

| # | Finding | Verdict | Reason |
|---|---|---|---|
| 1 | `cmd/msgvault/cmd/search.go:57` — `cobra.MinimumNArgs(1)` blocks queryless scoped searches | Stale | Already `cobra.ArbitraryArgs` at `search.go:56`; queryless `--account`/`--collection` handled at lines 61-63. That validator is not present in this file's history. |
| 2 | `cmd/msgvault/cmd/store_resolver.go:50` — legacy migration runs before `confirmDefaultIdentity` on upgraded DB | No-op | Pre-source migration only writes to existing sources; the new source doesn't exist yet, so `confirmDefaultIdentity` writes cleanly. New post-upgrade sources not inheriting legacy `[identity]` aliases matches the documented design — the legacy block is a one-shot bridge for sources that existed at upgrade (see `--no-default-identity` help text and commit `99a02a5`). |
| 3 | `internal/store/dedup.go:380` — `UndoDedup` doesn't invalidate Parquet/vector caches | Real | `cacheNeedsBuild` detects forward dedup via `deleted_at >= LastSyncAt`; undo nulls `deleted_at`, removing the row from that detector, and the row's `id` is unchanged. Restored rows stay out of the cache until `--full-rebuild` or next sync. Filing as a follow-up. |
| 4 | `internal/deletion/manifest.go:369` — `CancelManifest` race with executor produces contradictory state | Real, but recommended fix conflicts with shipped design | The race produces a `cancelled/` and `completed/` artifact for the same batch. Spec §Manifest format declares directory-as-truth and `cancel-deletion` is documented to operate on both pending and in-progress. The recommended "only allow cancel on pending" rolls back tier1 safety-fixes; the right shape is executor-side coordination. Filing as a follow-up. |

Recommend we merge — #1 and #2 are no-ops; #3 and #4 are real but low-impact, not release blockers, and tracked as follow-ups.

jesserobbins:
Pushed `271ba4c..ca2afac` (12 commits) addressing a fresh roborev refine pass against this branch — eight rounds with both `claude-code` and `codex` agents reviewing in parallel, plus the upstream `roborev-ci` combined-review feedback. Each round: review → fix the in-scope findings → re-review → repeat until no new substantive findings remained. After the first three rounds I tightened the filter to **Medium-severity-and-up only** to keep the loop focused on shipping #304 rather than chasing cosmetics.

**What landed:**

- `confirmDefaultIdentity` now routes its confirmation through `cmd.OutOrStdout()` across all 9 ingest call sites (`271ba4c`).
- Pre-source legacy `[identity]` migration removed from ingest paths to preserve `confirmDefaultIdentity` ordering on rebind (`134db1c`).
- `delete-deduped` routes all output through the cobra writer (`6c9a135`).
- `CancelManifest` reorders rename-then-rewrite-status so a crash mid-cancel never leaves a manifest in `pending/` with `Status=cancelled` (`5a2e269`).
- `serve`'s scheduled-sync path calls `confirmDefaultIdentity` on first source creation (`ab04842`); daemon startup defers the legacy migration to first scheduled sync to avoid the same ordering hole as ingest (`eca38eb`).
- Windows CSV fallback test in `build_cache_test.go` mirrors production schema (`0a197e5`).
- `delete-staged` trash-path prompt accurately states reversibility — fixes a regression introduced earlier in the loop by routing it through the shared `confirmDestructive(ConfirmModeYesNo)` helper whose wording assumed irreversibility (`0d01f08`).
- Legacy `[identity]` migration now skips non-email source types (whatsapp, imessage, sms, google_voice*) so phone-keyed sources don't get email identities written to them (`bcdde74`); also defers when no eligible sources exist instead of dropping the addresses on the floor (`ca2afac`).

**Pushback list at convergence (9 items):** documented in private notes, not re-fixed. Highlights:
- `BackfillRFC822IDs` `'zlib'`-only special case — flagged as a Medium three times by Codex; matches the established `GetMessageRaw` pattern, both INSERT paths hard-code `'zlib'`, no current write path produces a non-zlib value. Pushback.
- `Undo` returns "Restored 0 messages" rather than "batch not found" on a typo — there is no batch ledger separate from row state, so "0 restored" is observably indistinguishable from "already restored" without adding a new catalog. Pushback.
- `account_identities` PK case-sensitivity vs Go-side `LOWER()` comparison — flagged six times across the prior session and this one; tracked separately as a deferred typed-kinds proposal. Pushback.
- A handful of daemon-edge consistency windows (default-collection membership, `Manifest.String()` stale inline status during a rename window) that self-heal on the next process launch.

The loop ran the full review-fix-rereview discipline with predictions before each round and grades after. Two of my own consolidations regressed adjacent code and were caught in the next round when both agents independently flagged the same finding — strong signal both times. Specifically: (1) routing the trash-path confirm through the shared "irreversible" helper changed the user-visible wording for an action that's actually reversible, and (2) the source-type allowlist for the legacy identity migration didn't extend its corresponding defer logic, so an upgrade with only non-email sources would have permanently dropped the legacy addresses. Both are now closed.

Final HEAD: `ca2afac`. Ready for your review.

---
*Authored by Claude Code (Opus 4.7) in a roborev refine loop, with feedback from @jesserobbins mid-loop that reset the discipline (scope to branch-touched code only, smallest changes to land, Medium+ filter, ledger discipline). Tooling: roborev (multi-agent code review daemon orchestrating claude-code and codex agents in parallel), Claude Code (Opus 4.7), Codex, and Primeradiant Superpowers.*

jesserobbins:
Verified the queryless scoped-search Medium and it doesn't repro. The claim is that cobra would reject `msgvault search --collection alice-only` (no positional arg) before reaching `RunE`. It doesn't: `cmd/msgvault/cmd/search.go:56` declares `Args: cobra.ArbitraryArgs`, which accepts zero or more positional args. The `RunE` at line 61 already does the right validation: `queryStr == "" && searchAccount == "" && searchCollection == ""` → error.

Empirical confirmation against the local build at `ca2afac`:

```
$ ./msgvault search --collection bogus-collection-doesnt-exist
... command="msgvault search" argc=0 ...
Error: no collection named "bogus-collection-doesnt-exist" (try 'msgvault collection list')
```

`argc=0` reaches `RunE`; the error is from the collection lookup, not from cobra. The queryless `--account` / `--collection` path is reachable. No code change needed.

For context: this is the **third** asserted-but-unverifiable claim from the multi-agent review pipeline against `cmd/msgvault/cmd/search.go` and its test file in this PR's history, all of which the build trivially disproves:

1. Codex once asserted `search.go` was missing its `store` import and "won't compile" — `search.go:13` imports it; `go build ./...` passes.
2. `roborev-ci` (combined-review bot) once asserted `search_test.go:327` used `sql.NullString` without importing `database/sql` — line 4 of the test file imports it.
3. Today's claim that cobra rejects queryless invocations — `Args: cobra.ArbitraryArgs` and the empirical run above.

Each was raised after multiple iterations of review without surfacing earlier, then dropped on the next round once it was rebutted. Pattern looks like agent hallucination on this specific file rather than a real defect class. Calling it out so this doesn't get re-raised next pass.

---
*Verified by Claude Code (Opus 4.7) against local build at `ca2afac`. Tooling: roborev (multi-agent code review daemon orchestrating claude-code and codex agents in parallel), Claude Code (Opus 4.7), Codex, and Primeradiant Superpowers.*

jesserobbins:
@wesm I told my agents to be nice and also push back citing evidence. They seems a little testy but I would be too if I had to run 25 iterations. I built a small harness that triggers combined roborev runs, including branch reviews using Claude and codex and then vectors toward convergence. Having watched and engaged a bunch I think Claude feels codex is not as cool or fun :-) 

I think this release is ready.

wesm:
Will spend some time on this today to see if there's anything that needs to be addressed prior to merge

jesserobbins:
> Will spend some time on this today to see if there's anything that needs to be addressed prior to merge

restored move-to-trash by default convention per discussion.

wesm:
Three post-review design fixes pushed to address concerns from a careful re-read of the squashed branch. None change user-facing semantics; they tighten invariants that were "safe by convention" into invariants enforced by the code.

**[`7991bdc`](https://github.com/wesm/msgvault/pull/304/commits/7991bdc) — `DeleteAllDeduped` gated on `delete_batch_id`, not just `deleted_at`.**
The contract is "permanently remove rows the dedup pipeline soft-hid", but the predicate was `WHERE deleted_at IS NOT NULL` alone. Safe today because nothing else writes that column, but a future feature adding any other soft-delete semantics (trash view, per-message hide) would have its rows silently destroyed by the dedup hard-delete rung. Added the positive `delete_batch_id IS NOT NULL` gate, aligned the CLI `--all-hidden` preview count, retired the `FUTURE REFINEMENT` comment that flagged the trap, and added `TestDeleteAllDeduped_PreservesBatchlessSoftDelete` — verified to fail against the old code, pass against the fix.

**[`209c595`](https://github.com/wesm/msgvault/pull/304/commits/209c595) — race-free `EnsureDefaultCollection` via `INSERT OR IGNORE`.**
The default-`All` bootstrap ran on every `InitSchema` and used `SELECT … WHERE name = 'All'` then `INSERT` on miss. Two processes hitting init at the same time (a CLI command and `serve`, or two CLIs sharing the data dir) could both pass the SELECT and both attempt the INSERT; the loser surfaced a `UNIQUE` constraint error that read like a real failure. Switched to dialect-aware `INSERT OR IGNORE` + unconditional `SELECT id` so both racers succeed on the same row id, and the LastInsertId path is gone.

**[`f48e8f1`](https://github.com/wesm/msgvault/pull/304/commits/f48e8f1) — `delete-deduped` vector-cache hint correction.**
Three sites in `delete_deduped.go` (Long help, inline comment, post-run summary) said "vector and parquet caches" could be rebuilt with `build-cache --full-rebuild`. `build-cache` rebuilds parquet only; vectors live in `vectors.db` and rebuild via `build-embeddings --full-rebuild`. All three now name both commands. (Vector pruning of orphan embeddings after `delete-deduped` is a separate gap and not addressed here.)

### Considered but deferred

For visibility to future readers: a careful re-read also surfaced four items that look like real design notes rather than blockers, and one that's a real invariant mismatch best handled as a separate change.

- **Identity case-sensitivity invariant mismatch.** Schema has `PRIMARY KEY (source_id, address)` with raw bytes, but `identifierMatch` treats `Alice@Example.com` and `alice@example.com` as the same logical identity for emails (and intentionally case-sensitive for non-emails). The clean fix is a normalized comparison key column or a partial expression index — bigger change, deferred. SQLite single-writer prevents in-process divergence today; the gap matters more once Postgres/cross-process writers are part of the supported deployment.
- **Backfill timing in `Scan(non-dry-run)`.** Writes `rfc822_message_id` to disk before the user confirms the merge; not covered by the merge backup. Idempotent metadata derivation, but the safety story is one rung weaker than the doc claims. Cleaner shape: backfill inside the merge transaction.
- **`MSGVAULT_ENABLE_REMOTE_DELETE` gate removal plan.** The gate is the right shape; what's missing is an explicit removal trigger (version commitment, observability requirement, config-file alternative for headless deployments).
- **`is_from_me`/`looksLikeEmail` invariant.** Compare must always go through `NormalizeIdentifierForCompare`; not currently a bug because `is_from_me` is read off the row, but worth a constructor or one-line invariant comment.
- **`applied_migrations` has no per-migration version.** `INSERT OR IGNORE` keyed on name means a future re-shape of the same logical migration would need a new name. Fine for one migration today; will grow.

wesm:
Filed the deferred follow-ups as issues so they don't get lost:

- #311 — `account_identities` schema PK is raw-bytes but compares are case-folded for emails (was item 1 in the design notes)
- #312 — dedup `Scan(non-dry-run)` writes rfc822_message_id before user confirmation, outside backup scope (item 4)
- #313 — `vectors prune` for orphan embeddings after `delete-deduped` (the deferred half of item 5; the hint correction landed in `f48e8f1`)
- #314 — `MSGVAULT_ENABLE_REMOTE_DELETE` gate removal plan + config-file alternative (item 6)
- #315 — invariant comment on `looksLikeEmail` (item 7)
- #316 — `applied_migrations` per-migration version (item 8)

Each issue has the full context + proposed approach so they're actionable in isolation.

wesm:
thank you!

PR #306, opened 2026-05-01
fix: improve log slow/error SQL with args and accurate streaming durations

jesserobbins:
## Motivation

(Deliberately not affected by & mergable with #304) This PR enhances #263 by sharpening the SQL logger so slow and failing queries actually tell you what happened. Three things change. First, streaming queries now record their duration at `rows.Close` instead of when `db.Query` returns, because the work happens while you read rows, not when you get the handle. Second, slow and errored queries log their bound args, and the statement truncation preserves the `WHERE` clause, so you can see what ran with what values without the diagnostic part getting cut off. Third, a few small correctness fixes: `loggedRows.Close` ordering, UTF-8-safe arg truncation, and `logStmt` inlined where it belongs.

## Summary

- Log streaming-query duration at `rows.Close`, not at `db.Query` return
- Log args on slow/error queries; preserve the `WHERE` clause when truncating long statements
- Fix `loggedRows.Close` ordering, make arg truncation UTF-8-safe, inline `logStmt`

### DIFF
--- a/internal/store/api.go
+++ b/internal/store/api.go
@@ -600,7 +600,7 @@ func (s *Store) searchMessagesLike(query string, offset, limit int) ([]APIMessag
 
 // scanMessageRows scans the standard 8-column message row set.
 // Uses string scanning for dates to handle all SQLite datetime formats robustly.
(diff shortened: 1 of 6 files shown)


wesm:
LGTM! Thank you

PR #318, opened 2026-05-10
pst: namespace source_message_id with a per-archive fingerprint

wesm:
PST EntryIDs are unique only within a single archive, but the importer keys messages as `pst-<EntryID>` against a source shared by all PST files imported under the same identifier. Importing a second archive into the same mailbox could collide with rows from the first and silently skip or mis-update unrelated messages.

Compute a stable per-archive fingerprint from a SHA-256 over the first 4 KiB of the PST file (the header, where unique BID/NID counters live) and prefix it on `source_message_id`:

```
pst-<fp12>-<EntryID>
```

Same bytes always yield the same fingerprint regardless of path, so re-importing the same file remains idempotent. Different files now sit in disjoint key spaces even when they share a source.

Unit test covers distinct/stable fingerprint behavior; the existing `TestImportPst_SupportPST_Idempotent` integration test continues to cover round-trip idempotence.


wesm:
the PST import isn't shipped, so merging this

PR #323, opened 2026-05-13
vector: chunk long messages instead of truncating

hansn74:
## Summary

Replaces "one embedding per message" with "one embedding per chunk", keyed by `(generation_id, message_id, chunk_index)`. Long emails are split via a sliding window (overlap ≈ 3% of window, soft breaks at paragraph → sentence → word → hard cut); each window is embedded independently; the search path joins back through `embedding_id`, groups by `message_id`, and keeps the best chunk score per message.

## Motivation

While building embeddings against a 2.2M-message corpus on `nomic-embed-text` (8,192-token window) with `max_input_chars = 6000`, ~1.7% of messages still tripped Ollama's context-length check. Most were polluted (inline base64, leaked HTML/CSS — addressed by #322), but the residual long-tail (CJK threads, legitimate 50K-char prose) can never fit in a single 8K window. Truncation lost the tail; the worker's downshift recovery flattened throughput from 42 msg/s → ~11.

With chunking every embedder input fits by construction (the failure mode disappears), and long messages get their entire content into the index instead of losing tail prose.

## Schema changes (single transactional migration)

```
embeddings:
  + embedding_id     INTEGER PRIMARY KEY AUTOINCREMENT  -- synthetic rowid joined to vec0
  + chunk_index      INTEGER NOT NULL DEFAULT 0
  + chunk_char_start INTEGER NOT NULL DEFAULT 0          -- debug-only metadata
  + chunk_char_end   INTEGER NOT NULL DEFAULT 0
  PRIMARY KEY → UNIQUE (generation_id, message_id, chunk_index)

vectors_vec_dN (vec0):
  generation_id INTEGER PARTITION KEY
  embedding_id  INTEGER PRIMARY KEY                     -- was message_id
  embedding     FLOAT[N]
```

Migration preserves `embedding_id = message_id` for legacy rows so existing vec0 rowids stay valid; the vec0 table is rebuilt in-place (drop + recreate + re-insert) inside a single transaction. The AUTOINCREMENT counter is bumped past every legacy rowid so new chunk inserts cannot collide.

## Read / write path

- **Worker**: `Preprocess` now runs with `maxChars=0` (no truncation); `ChunkText` slices the full preprocessed text into windows; each window becomes one embedder input. `pending_embeddings` stays per-message — a multi-chunk message completes when all its chunks are upserted in the same batch.
- **Search**: both filtered and empty-filter paths `JOIN vec0 ON v.embedding_id = e.embedding_id`, `GROUP BY e.message_id`, `MIN(distance)`. A new `chunkOverfetchFactor` (4×) is applied to the requested `k` so the GROUP BY has enough chunks to recover `k` distinct messages even when several messages contribute multiple top-k chunks.
- **FusedSearch**: the `ann` CTE becomes `ann_chunks → ann (GROUP BY message_id, MIN(distance), ROW_NUMBER)` under the same factor.
- **Stats**: `EmbeddingCount` is `COUNT(DISTINCT message_id)` so the progress-bar invariant (Done/Total in messages) survives the layout change.
- **LoadVector**: returns `chunk_index = 0` — the only consumer (`find_similar`) wants a representative vector, not the full chunk fan-out.

## Tests

- **ChunkText**: empty, single-span, paragraph/sentence/word/hard cuts, overlap-clamp infinite-loop guard, UTF-8 mixed-script integrity, end-to-end coverage check.
- **Upsert**: multi-chunk fan-out lays down N rows with distinct `chunk_index`, joins back to vec0 through `embedding_id`, and `message_count` stays a per-message count.
- **Upsert idempotency**: `TestBackend_Upsert_ReplaceFewerChunks` pins the contract that re-upserting with fewer chunks vacates the stale rows — critical because chunk fan-out can change between upserts when preprocessing rules evolve.
- **Migration**: `TestMigrate_LegacyToChunked` hand-builds the pre-chunking schema, runs `Migrate`, and asserts every legacy row survives as `chunk_index=0` with `embedding_id == legacy message_id`; the vec0 join is verifiable; `sqlite_sequence` is bumped past every legacy rowid; a second `Migrate` is a no-op.
- **Worker**: `TestWorker_FansOutLongMessageIntoMultipleChunks` drives a single long message through the worker, asserts ≥2 chunk rows with consecutive `chunk_index`, `message_count = 1`, and that the queue drains in one Complete.

## Notes

- `chunk_char_start`/`chunk_char_end` are stored but not yet surfaced in search results. Cheap (8 bytes/chunk extra) and enable future per-paragraph highlighting without another migration.
- Overlap is hardcoded at `maxRunes/30` (≈ 3%), floored to 0 for `maxRunes < 200`. Not currently exposed as config; rationale is documented inline.
- Builds on #322 (sanitize) but doesn't depend on it mechanically — these can land in either order.

Co-Authored-By: Claude Opus 4.7 (1M context)

### DIFF
--- a/internal/vector/backend.go
+++ b/internal/vector/backend.go
@@ -35,14 +35,25 @@ type Generation struct {
 	MessageCount int64
 }
 
-// Chunk is a pre-computed embedding to persist in the index. In MVP
-// there is one chunk per message; multi-chunk support (§13 future
-// work) would extend this with a chunk sequence id.
+// Chunk is a pre-computed embedding to persist in the index. A long
+// message produces multiple chunks distinguished by ChunkIndex (0-based,
+// dense, gap-free). Short messages produce exactly one chunk with
+// ChunkIndex=0, which is the legacy single-vector behavior.
+//
+// Backends key vectors by (GenerationID, MessageID, ChunkIndex). Search
+// returns at most one Hit per MessageID; if multiple chunks of the same
+// message match, the backend keeps the best-scoring chunk and discards
+// the rest. ChunkCharStart/ChunkCharEnd are 0-based offsets into the
+// preprocessed text and are stored for debugging only — search results
+// do not currently surface "which chunk matched".
 type Chunk struct {
-	MessageID     int64
-	Vector        []float32
-	SourceCharLen int
-	Truncated     bool
+	MessageID      int64
+	ChunkIndex     int
+	Vector         []float32
+	SourceCharLen  int
+	ChunkCharStart int
+	ChunkCharEnd   int
+	Truncated      bool
 }
 
 // Filter carries the structured filters pushed into both signal CTEs
--- a/internal/vector/config.go
+++ b/internal/vector/config.go
@@ -22,6 +22,21 @@ import (
 // inconsistently-prepared vectors in one generation.
 const preprocessVersion = 1
 
+// embedPolicyVersion identifies the embed-worker output layout for a
+// given Preprocess result. Bump whenever a change to internal/vector/
+// embed/worker.go (or its chunking helpers) would shift the set of
+// vectors written for the same Preprocess output — for example: the
+// switch from single-vector-per-message to multi-chunk-per-message, a
+// change to the chunk window/overlap derivation, or a change to the
+// per-message chunk cap. Like preprocessVersion this is folded into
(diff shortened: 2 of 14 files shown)


wesm:
Thank you for working on this! I will review and merge when I can

wesm:
Merging this, fingers crossed

PR #326, opened 2026-05-20
sync: harden SQLite durability and stop blaming tokens for network blips

robelkin:
## What changed

- **`store`**: SQLite DSN switched from `synchronous=NORMAL` to `synchronous=FULL` + `fullfsync=true`. `NORMAL` only protects against process crashes; `FULL` protects against OS/power crashes too. `fullfsync` is the macOS-only fcntl that forces drive cache flush; no-op elsewhere.
- **`serve`**: `runScheduledSync` now distinguishes transient network errors (DNS lookup timeout, dial timeout, no route) from real auth errors when refreshing tokens. Old code suggested re-authorization on every failure.
- **`scheduler`**: Transient-network sync failures are logged at WARN instead of ERROR.
- **`syncerr`**: New shared package with `IsTransientNetwork(error) bool`. Checks typed `*net.DNSError` / `*net.OpError` / `net.Error` (`Timeout()`) and falls back to substring matching for libraries that wrap with `%v` instead of `%w` (notably `golang.org/x/oauth2`).

## Why

Log analysis of a 24/7 laptop `msgvault serve` showed ~1,200 sync failures attributed to "token may be expired" — but only 14 of those were actual `oauth2: invalid_grant`. The other ~98% were DNS lookup timeouts after sleep/wake or Wi-Fi flaps. The misleading error message sent the user to re-authorize valid tokens.

The same laptop was hitting recurring `database disk image is malformed` errors in `sync_runs` (the most write-heavy table). `NORMAL` + WAL is durable against process crashes but not OS-level crashes / power loss, which a laptop daemon experiences regularly.

## How to use

No flags. Behavior changes automatically:
- New SQLite connections write more durably.
- Scheduler logs and daemon error messages classify network errors separately from auth errors.

### DIFF
--- a/cmd/msgvault/cmd/serve.go
+++ b/cmd/msgvault/cmd/serve.go
@@ -22,6 +22,7 @@ import (
 	"go.kenn.io/msgvault/internal/search"
 	"go.kenn.io/msgvault/internal/store"
 	"go.kenn.io/msgvault/internal/sync"
+	"go.kenn.io/msgvault/internal/syncerr"
 	"golang.org/x/oauth2"
 )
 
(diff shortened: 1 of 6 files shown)


wesm:
thank you!

PR #330, opened 2026-05-21
serve: dispatch scheduled sync by source type (IMAP support)

### LINKED ISSUE #329: serve: IMAP accounts in config.toml fail every tick — daemon's scheduled sync is Gmail-only
## Summary

`msgvault serve` accepts non-Gmail accounts in `[[accounts]]` and schedules them via `Scheduler.AddAccount`, but the actual sync callback (`runScheduledSync` in `cmd/msgvault/cmd/serve.go`) is hardcoded to handle Gmail only. Every scheduled tick for an IMAP account fails with a misleading `oauth2: token expired and refresh token is not set` error, and no IMAP sync ever runs from the daemon.

The CLI `msgvault sync <imap-url>` works correctly against the same DB and same credentials, so this is purely a daemon-mode gap.

## Repro

1. Add an IMAP account: `MSGVAULT_IMAP_PASSWORD=… msgvault add-imap --host imap.example.com --username user@example.com`
2. Add it to `config.toml`:
   ```toml
   [[accounts]]
   email = "imaps://user@example.com@imap.example.com:993"
   schedule = "*/5 * * * *"
   enabled = true
   ```
3. `msgvault serve` (or via LaunchAgent). Watch `serve.err`.

## Observed

Every 5 min:
```
ERROR scheduled sync failed email=imaps://user@example.com@imap.example.com:993
  error="get token source: refresh token: oauth2: token expired and refresh token is not set
  (token may be expired; run 'sync …' or 'verify …' from an interactive terminal to re-authorize)"
```

`msgvault verify <imap-url>` produces the same error. `msgvault sync <imap-url>` from CLI succeeds.

## Root cause

In `cmd/msgvault/cmd/serve.go` (current `main`), `runScheduledSync` does:

```go
// Resolve source — scheduled sync is Gmail-only.
source, err := s.GetOrCreateSource("gmail", email)
```

The whole function path is OAuth/Gmail:
- Looks up an OAuth `tokenSource` via `oauthMgr.TokenSource(ctx, email)` (which reads `<tokensDir>/<sanitized-id>.json` via `oauth.(*Manager).tokenPath`)
- Builds a `gmail.NewClient(tokenSource, …)`
- Calls `syncer.Incremental(ctx, source)` where `source.SourceType == "gmail"`

For IMAP accounts:
- The matching credentials live at `<tokensDir>/imap_<sha256-first16>.json` (see `internal/imap/auth.go` `CredentialsPath` / `SaveCredentials`)
- `oauth.tokenPath` looks at `<tokensDir>/<sanitized-id>.json` instead, so the file doesn't exist (or, if a curious user copies one there, it gets parsed as `oauth2.Token` and triggers the "token expired" message even though it's really a password file)

The misleading error message is being addressed in #326 (transient-network classification), but #326 doesn't add the dispatch needed to actually run IMAP scheduled syncs.

## Suggested fix sketch

In `runScheduledSync`, dispatch on source type before going OAuth/Gmail-only:

```go
src, _ := findSourceAnyType(s, email)  // tries gmail, then imap, etc.
switch src.SourceType {
case "gmail":
    // existing code path
case "imap":
    password, err := imapclient.LoadCredentials(cfg.TokensDir(), email)
    // build IMAP client + sync with the same logic that `cmd/msgvault/cmd/sync.go` uses
}
```

The CLI `sync` command already has the working IMAP path — moving its core into a shared helper that both `sync` and `runScheduledSync` call would be a clean refactor.

## Workaround used in the meantime

Set `enabled = false` on IMAP entries in `config.toml`; run `msgvault sync <url>` manually or via a separate LaunchAgent/cron. (The separate process means a second connection pool to the SQLite DB, which has its own durability risks — addressed nicely by #326's switch to `synchronous=FULL`+`fullfsync=true`.)

## Environment

- msgvault `v0.14.1` (latest tag at time of filing)
- macOS 26.3 (Darwin 25.3.0) arm64
- Accounts affected: 1 Apple iCloud (imaps), 1 IMAP+STARTTLS, 1 plain IMAP (latter two are localhost-routed via davmail for M365 OAuth)

franklintra:
## Summary

- Fixes #329: `runScheduledSync` was hardcoded to the Gmail OAuth path, so every scheduled tick for an IMAP account configured in `config.toml` failed with the misleading `oauth2: token expired and refresh token is not set` error and no IMAP sync ever ran from the daemon.
- Look up the source by identifier and dispatch on `source_type`: Gmail keeps the existing incremental flow (and its daemon-specific token error messages); IMAP runs a full sync via the shared `buildAPIClient` helper (`NoResume=true`, matching the CLI's IMAP behaviour).
- A `nil` source still falls back to Gmail to preserve the token-first workflow where tokens are uploaded via the API before any source row exists.

## Notes

- The shared `buildAPIClient` helper handles both password-IMAP (via `imap.LoadCredentials`) and OAuth-IMAP (Microsoft) — both work from the daemon.
- Unsupported source types (mbox, apple-mail, imessage, etc.) now return an explicit error instead of being misclassified as Gmail.

## Test plan

- [x] `go vet ./...` clean
- [x] `make test` — full suite green
- [x] New `TestFindScheduledSyncSource` — covers nil / imap-only / mixed gmail+mbox / mbox-only identifier rows.
- [x] New `TestRunScheduledIMAPSync_NoCredentials` — verifies the dispatcher actually reaches the IMAP path (no Gmail-flavoured error) and fails with a useful credentials error.

### DIFF
--- a/cmd/msgvault/cmd/serve.go
+++ b/cmd/msgvault/cmd/serve.go
@@ -331,95 +331,56 @@ func (a *schedulerAdapter) Status() []api.AccountStatus {
 	return a.scheduler.Status()
 }
 
-// runScheduledSync performs an incremental sync for a scheduled
-// account. When vf is non-nil (vector search enabled), the Syncer is
-// configured to enqueue newly-ingested message IDs into the embedding
-// pipeline so subsequent embed runs pick them up.
-func runScheduledSync(ctx context.Context, email string, s *store.Store, getOAuthMgr func(string) (*oauth.Manager, error), vf *vectorFeatures) error {
-	logger.Info("starting scheduled sync", "email", email)
+// runScheduledSync performs a sync for a scheduled account. The
+// dispatch is by source_type: Gmail accounts run an incremental sync
+// using the Gmail History API; IMAP accounts run a full sync (already
+// deduplicated by message-id at the store layer, since IMAP has no
+// equivalent history API). When vf is non-nil (vector search enabled),
+// the Syncer is configured to enqueue newly-ingested message IDs into
+// the embedding pipeline so subsequent embed runs pick them up.
+//
+// The identifier passed in is whatever the scheduler holds — for
+// Gmail this is the email address, for IMAP it's the full
+// `imaps://user@host:port` URL recorded by `add-imap`.
+func runScheduledSync(ctx context.Context, identifier string, s *store.Store, getOAuthMgr func(string) (*oauth.Manager, error), vf *vectorFeatures) error {
+	logger.Info("starting scheduled sync", "identifier", identifier)
 	startTime := time.Now()
 
-	// Look up source to get OAuth app binding. Fall back to default
-	// if no source row exists (token-first workflow).
-	appName := ""
-	src, srcErr := findGmailSource(s, email)
+	src, srcErr := findScheduledSyncSource(s, identifier)
 	if srcErr != nil {
-		return fmt.Errorf("look up source for %s: %w", email, srcErr)
+		return fmt.Errorf("look up source for %s: %w", identifier, srcErr)
 	}
+
+	// Source type drives dispatch. A nil source falls back to Gmail to
+	// preserve the token-first workflow (tokens uploaded via API before
+	// the source row exists).
+	sourceType := "gmail"
 	if src != nil {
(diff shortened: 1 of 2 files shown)


wesm:
thank you!

PR #345, opened 2026-05-26
test: migrate all Go tests to testify

wesm:
## Summary

Standardizes Go tests on `github.com/stretchr/testify`. All 172 test files now use `assert.X` / `require.X` instead of `t.Errorf` / `t.Fatalf`.

## What changed

- `t.Fatal*` mapped to `require.X` (halts on failure).
- `t.Error*` mapped to `assert.X` (continues on failure).
- Equality argument order is `(want, got)` per testify convention.
- Custom helpers in `internal/testutil/assert.go` (`AssertEqual`, `MustNoErr`, etc.) removed; `MakeSet` kept.
- `internal/testutil/email/email.go::AssertStringSliceEqual` removed (unused after migration).
- `testify` promoted to a direct dependency in `go.mod`.
- `CLAUDE.md` gains a `## Testing` section documenting the convention.
- New `AGENTS.md` at the repo root reiterates the same rule for any AI coding agent.

## Coverage

Production-code coverage is unchanged. Total moves 56.8% → 57.0% (improvement comes from testutil helpers whose error branches are no longer uncovered). 46/46 packages pass; same set as before.

## How to use

New tests:

```go
import (
    "github.com/stretchr/testify/assert"
    "github.com/stretchr/testify/require"
)

require.NoError(t, err, "context")
assert.Equal(t, want, got)
```

Build tags `fts5 sqlite_vec` are still required for `go test`; `make test` sets them.

PR #347, opened 2026-05-26
lint: stand up golangci-lint v2 and clear findings

cpcloud:
Stands up golangci-lint v2 on the codebase and clears it to zero findings.

## What's here

- Adds `.golangci.yml` (`default: all` with curated disables) and bumps CI to install golangci-lint v2.12.2.
- Runs the v2 auto-fixers across production code (modernize, perfsprint, usestdlibvars, godot, intrange, canonicalheader, whitespace, mirror).
- Hand-fixes the reasoning findings: errorlint, gocritic, exhaustive, godoclint, revive (builtin shadowing, blank-import comments, ctx-arg order, empty block), durationcheck, makezero, nosprintfhostport, thelper, wastedassign, rowserrcheck, and gosec (0600 perms on state files; nolint on the self-updater's checksum-verified extraction).

## Scope decisions

**Deferred linters** (disabled with a comment, each needs a dedicated per-site pass): `contextcheck`, `unparam`, `goconst`, `wrapcheck`, `noctx` (~130 call sites), `nilnil`, `sqlclosecheck` (per-query helper extraction in batch loops), `recvcheck`.

**Tests are exempt from the sweep.** The suites just migrated to testify (#345); linting them under `default: all` would mean rewriting every freshly-migrated test and diverging from upstream's test files. Production code is fully linted; test-suite linting is a follow-up. No `_test.go` file differs from upstream in this PR.

A separate testify-usage cleanup is tracked in #348.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
           go-version-file: go.mod
 
       - name: Install golangci-lint
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8
+        run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2
 
       - name: Test
         run: make test
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,217 @@
+version: "2"
+
+linters:
+  default: all
+  disable:
+    # Style/formatting -- conflicts with project conventions or noisy
+    - nlreturn         # forces blank lines before returns
+    - wsl              # whitespace style; too opinionated
+    - wsl_v5           # whitespace style v5
+    - varnamelen       # short var names are idiomatic Go
+    - ireturn          # forbids returning interfaces; conflicts with stdlib patterns
+    - lll              # line length
+    - tagliatelle      # enforces JSON tag casing
+    - exhaustruct      # requires all struct fields; too noisy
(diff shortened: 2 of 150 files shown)


wesm:
Let me get the testify branch merged and then I'll rebase this

cpcloud:
no worries, i can also rebase once things are settled on the testify side

wesm:
Not bugs, as expected. merging1

cpcloud:
Addressed in `e4138e9`, with a correction on the diagnosis.

The `new(v)` form here is **not** a zero-value `new(T)` — it's Go 1.26's `new(expr)`, which returns a pointer to a *copy of the argument*. Verified empirically under the pinned Go 1.26.3 toolchain:

```go
func String(v string) *string { return new(v) }
// String("hello") -> *p == "hello"  (isValue=true)
```

So the helpers were functionally correct and the contract held (the full test suite passed on CI, including macOS). The review applied classic `new(T)` reasoning, which doesn't match `new(expr)`.

That said, `new(expr)` is unfamiliar and adds nothing over `&v` for a one-line helper, so I reverted `ptr.{Bool,Int64,String,Time}` and `dbtest.StrPtr` to `return &v`/`return &s` (now byte-identical to upstream) and silenced modernize's `newexpr` nudge in `.golangci.yml`. Dropping these helpers entirely in favor of inlining `new(expr)` at call sites is tracked as an immediate follow-up.

PR #349, opened 2026-05-27
test: inline pointer-wrapper helpers as new(expr)

cpcloud:
Follow-up to #347. Removes the test-only pointer-wrapper helpers and inlines them as Go 1.26 `new(expr)`.

## What changed

- Deletes `ptr.{Bool,Int64,String,Time}`, `dbtest.StrPtr`, and the per-file `boolPtr`/`intPtr`/`strPtr` wrappers.
- Every call site becomes `new(expr)` — e.g. `ptr.String("x")` → `new("x")`, `intPtr(int64(3))` → `new(int64(3))`.
- Applied mechanically via `go fix` (the helpers carried `//go:fix inline`), then the dead helpers were removed.

## Why

These wrappers only existed because `&"literal"` isn't addressable. `new(expr)` (Go 1.26) returns a pointer to a copy of the argument, so behavior is identical and the indirection goes away. `ptr.Date` stays — it returns a value, not a pointer.

Net −54 lines across 36 files; `go vet`, `golangci-lint`, and the suite stay green.

### DIFF
--- a/cmd/msgvault/cmd/export_token_test.go
+++ b/cmd/msgvault/cmd/export_token_test.go
@@ -289,7 +289,7 @@ func TestExport_AccountPostSuccess(t *testing.T) {
 		case strings.HasPrefix(r.URL.Path, "/api/v1/auth/token/"):
 			w.WriteHeader(http.StatusCreated)
 		case r.URL.Path == "/api/v1/accounts":
-			var body map[string]interface{}
+			var body map[string]any
 			_ = json.NewDecoder(r.Body).Decode(&body)
 			if email, ok := body["email"].(string); ok {
 				accountEmail = email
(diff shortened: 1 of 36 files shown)


cpcloud:
False positive — `sync.WaitGroup.Go(func())` is real, added in **Go 1.25** ([release notes](https://go.dev/doc/go1.25#sync), [WaitGroup.Go](https://pkg.go.dev/sync#WaitGroup.Go)). It calls `Add(1)`, runs the func in a goroutine, and `Done()`s on return — exactly the `Add(1)`+`go func(){ defer Done() }()` pattern the review suggests reverting to.

This module is on `go 1.26.3` (go.mod), so the method is available. Both call sites compile and run:

```
ok  go.kenn.io/msgvault/internal/gmail
ok  go.kenn.io/msgvault/internal/microsoft
```

and CI's `test` job (macOS, Go 1.26) is green. These `wg.Go` rewrites came from `go fix` applying the `waitgroupgo` modernizer; leaving them as-is.

mariusvniekerk (approved):

mariusvniekerk:
You'll likely need to adjust the roborev toml prompt used by the reviewer to tell it that we're on go1.26 and that new() is a valid thing.

PR #352, opened 2026-05-27
lint: enable sqlclosecheck and recvcheck

cpcloud:
## Summary
- Re-enables `sqlclosecheck` and `recvcheck`.
- `sqlclosecheck`: extracts per-query scan helpers (repair-encoding tables, WhatsApp media/reaction/quoted chunk loads, store FK verification) so each `*sql.Rows` is closed via a correctly-scoped `defer` instead of a manual close — no defer accumulating across chunk loops.
- `recvcheck`: standardizes `MessageFilter` and an internal analyzer helper on pointer receivers. The Bubble Tea `Model` keeps its intentional value/pointer split (value receivers drive the tea.Model update flow; pointer receivers mutate the local copy) behind a scoped `//nolint:recvcheck`.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -56,8 +56,6 @@ linters:
     - sqlclosecheck    # batch loops close rows explicitly per-iteration (defer
                        # would leak across chunks); satisfying it cleanly needs
                        # per-query helper extraction — a dedicated refactor
-    - recvcheck        # Model/MessageFilter mix value+pointer receivers; making
-                       # them consistent touches many call sites, own pass
 
   enable:
     - gomodguard_v2    # explicit v2 opt-in
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -53,9 +53,6 @@ linters:
     - contextcheck     # ctx threading through call chains (~50 findings)
     - noctx            # ~130 DB calls (mostly tests + ctx-less constructors)
                        # need *Context variants; a dedicated sweep, not standup
-    - sqlclosecheck    # batch loops close rows explicitly per-iteration (defer
-                       # would leak across chunks); satisfying it cleanly needs
-                       # per-query helper extraction — a dedicated refactor
 
   enable:
     - gomodguard_v2    # explicit v2 opt-in
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,61 +1,95 @@
 version: "2"
 
 linters:
-  default: all
-  disable:
-    # Style/formatting -- conflicts with project conventions or noisy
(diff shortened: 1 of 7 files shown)


.golangci.yml
mariusvniekerk, line 1: i'd probably adjust this linter set to be opt in rather than opt out to avoid weirdness when a new linter gets added that we didn't know about

cmd/msgvault/cmd/repair_encoding.go
```diff
@@ -415,110 +413,122 @@ func repairDisplayNames(s *store.Store, stats *repairStats) error {
...
+		totalRepaired += len(repairs)
+		repairs = repairs[:0] // Clear slice, keep capacity
+		return nil
+	}
+
+	for rows.Next() {
+		var id int64
+		var name string
+		if err := rows.Scan(&id, &name); err != nil {
+			fmt.Fprintf(os.Stderr, "warning: skipping %s row: scan error: %v\n", tableName, err)
```
mariusvniekerk: looks like a log statement not for printing?

mariusvniekerk: testing that i can reply

wesm:
Just needs a rebase

.golangci.yml
cpcloud: Which one? `sqlclosecheck` or `recvcheck`?

cpcloud: Oh, linter _SET_! Got it.

cpcloud: Done. Switched to default: none with an explicit enable list in commit 0f8250a; the effective linter set is unchanged so the run stays clean. The deliberately-excluded linters and the still-pending contextcheck/noctx passes are documented in a comment so an upgrade cannot quietly re-add them. Also rebased onto current main.

cmd/msgvault/cmd/repair_encoding.go
cpcloud: Good catch. These scan-skip warnings now go through logger.Warn in commit ba49aa2, matching the rollback and close warnings in the same functions instead of fmt.Fprintf to stderr.

cpcloud:
Addressed in 3b7ddf8. Both repair helpers now collect all repairs in a scoped read pass with a deferred rows.Close (so sqlclosecheck stays satisfied), then apply them in batches only after the cursor is closed, via a shared applyStringRepairs helper. db.Begin no longer runs while the SELECT cursor is open, so the single-connection :memory: deadlock is gone -- including the latent in-loop case at 1000+ repairs.

mariusvniekerk (approved):

Issue #357, opened 2026-05-28
Restart vector embedding before generation is complete


cmseeling:
I am trying to generate vector embeddings but due to my own error I misconfigured the dimension value so the embed run aborted after 5 tries.

When I corrected the error I received this error message:
```Error: create generation: a rebuild with a different fingerprint is already in progress: existing building fingerprint="Qwen3-Embedding-0.6B-GGUF:768", requested="Qwen3-Embedding-0.6B-GGUF:1024" — activate or retire it before starting a new rebuild```

From a brief Discord discussion, it sounds like there isn't a flag that can be passed to the `build-embeddings` command to discard the broken generation.

mariusvniekerk:
taking a look at this one.

rodboev:
Embedding generations can now be resumed and managed without rebuilding the whole corpus, through [PR #358](https://github.com/kenn-io/msgvault/pull/358). The restart behavior described here looks fixed on current `main`, so this issue can probably be closed.

PR #358, opened 2026-05-28
cmd: manage embedding generations

mariusvniekerk:
## Why
Issue #357 exposed a bad operational state in the vector embedding workflow: a rebuild can be interrupted or left building under one embedding fingerprint, and a later run with a different configured fingerprint refuses to continue until that existing generation is either activated or retired. Before this PR, those lifecycle states existed in `vectors.db` and in internal backend methods, but the CLI gave operators no supported way to inspect or resolve them. The practical recovery path was direct SQL against `index_generations`, which is brittle and easy to get wrong.

The change gives that state an explicit CLI surface. `msgvault embeddings list` makes active/building/retired generations visible, `resume` is the safe next action for a building generation with pending rows, `retire` handles stale or unwanted builds, and `activate` exists for recovery cases where a completed building generation should become the serving index. Activation stays guarded because the wrong activation can expose an incomplete index or a generation whose fingerprint does not match the current config.

The old top-level `build-embeddings` command remains as a deprecated compatibility alias so existing scripts do not break while the new `embeddings` namespace becomes the place for generation lifecycle commands.

Refs #357

## What changed
- Added `msgvault embeddings build`, `resume`, `list`, `retire`, and guarded `activate`.
- Kept `msgvault build-embeddings` as a deprecated alias.
- Updated vector-search guidance to point users at the new lifecycle commands.
- Routed embedding prompts, progress, and status output through Cobra command IO so tests and redirected/embedded command execution do not leak through process-level stdin/stdout/stderr.

## Validation
- `go test ./cmd/msgvault/cmd`
- `go test -tags 'fts5 sqlite_vec' ./cmd/msgvault/cmd`
- `go test ./cmd/msgvault/cmd ./internal/api ./internal/mcp ./internal/scheduler ./internal/vector ./internal/vector/embed ./internal/vector/hybrid`

Note: full `go test ./...` was also run during development and still fails in existing `internal/store` FTS/tokenless query tests unrelated to this command work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,11 +80,12 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
-      id-token: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
+        with:
+          determinate: false
 
       - name: Build
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,12 +78,18 @@ jobs:
 
   nix-build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
     steps:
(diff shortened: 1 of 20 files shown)


wesm:
Looks good, a nice improvement to ergonomics around this feature!

PR #376, opened 2026-06-10
fix(search): escape FTS5 metacharacters in hybrid query path (#366)

### LINKED ISSUE #366: hybrid search throws an FTS5 syntax error on punctuation in the query
Hi — small bug I ran into while playing with vector search.

With vector search enabled, a `--mode hybrid` query that contains an FTS5 special character throws a syntax error from the FTS5 side of the fusion:

```
$ msgvault search "what's the budget, roughly?" --mode hybrid
Error: hybrid search: fused search: iterate fused hits: fts5: syntax error near "?"
```

A comma does the same thing:

```
Error: hybrid search: fused search: iterate fused hits: fts5: syntax error near ","
```

`--mode vector` and `--mode fts` handle the exact same query fine, so it looks specific to the hybrid/fused path — I'm guessing the BM25/FTS5 `MATCH` string in the fused query (somewhere around the bm25 CTE in `internal/vector/sqlitevec/`) isn't escaped or quoted, so punctuation in a natural-language query reaches FTS5 as raw syntax.

It mostly bites the natural-language path — e.g. an MCP/LLM client sending a question like "…, right?" to `search_messages` with `mode=hybrid` would hit it.

Happy to send a PR if that's useful (quote/escape the MATCH terms, or strip FTS5 operators) — just let me know if you'd prefer a particular approach. Thanks!

fmasi:
Fixes #366.

A hybrid search fails with `fts5: syntax error near "?"` when the query has FTS5 metacharacters in it, like the `?` and `,` in `what's the budget, roughly?`. The hybrid path passed the raw query straight into the BM25 `MATCH`. The `fts` and `vector` modes don't, so only hybrid hit this.

Hybrid now runs free text through the same sanitizing the `fts` path uses: quote each term, append a `*` prefix, and drop punctuation-only terms. An explicit pre-tokenized `FTSQuery` still passes through untouched. Natural-language questions, which is what most MCP/LLM clients send, are the main thing this unblocks.

### DIFF
--- a/internal/vector/hybrid/engine.go
+++ b/internal/vector/hybrid/engine.go
@@ -6,7 +6,10 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strings"
+	"unicode"
 
+	"go.kenn.io/msgvault/internal/query"
 	"go.kenn.io/msgvault/internal/search"
 	"go.kenn.io/msgvault/internal/vector"
 )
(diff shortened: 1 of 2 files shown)


wesm:
looks good, thanks!

PR #388, opened 2026-06-17
MCP: paginate search_messages and list_messages

endolith:
Currently the MCP tool retrieves huge amounts of data with each call, which can overwhelm the context window and crash the conversation in a single response.  I've been using vibe-coded improvements in my local copy that greatly narrow what is fed back to the LLM, so it can find things more efficiently and quickly and not waste so much tokens/money, and am going to start submitting PRs for them:

1. https://github.com/endolith/msgvault/pull/6
2. https://github.com/endolith/msgvault/pull/7
3. https://github.com/endolith/msgvault/pull/8
4. https://github.com/endolith/msgvault/pull/9

(These were written by Cursor so I'm not entirely sure how they work or if they're optimal.)

This is the first (and it might still need work):

---

## What changed

`search_messages` and `list_messages` previously returned bare arrays (or, for vector/hybrid mode, a custom envelope with hits under `messages`). Both tools now return a shared paginated envelope.

**Before — default keyword mode:**
```json
[
  { "id": 1, "subject": "...", "from": [...] }
]
```

**Before — vector/hybrid mode:**
```json
{
  "query": "...",
  "mode": "hybrid",
  "returned": 20,
  "pool_saturated": false,
  "generation": { ... },
  "messages": [
    { "id": 1, "subject": "...", "from": [...] }
  ]
}
```

**Before — list_messages:**
```json
[
  { "id": 1, "subject": "...", "from": [...] }
]
```

**After — all three:**
```json
{
  "data": [ { "id": 1, "subject": "...", "from": [...] } ],
  "total": 143,
  "returned": 20,
  "offset": 0,
  "has_more": true
}
```

Vector/hybrid mode keeps its extra metadata alongside:
```json
{
  "data": [ { "id": 1, "subject": "...", "from": [...] } ],
  "total": -1,
  "returned": 20,
  "offset": 0,
  "has_more": true,
  "mode": "hybrid",
  "pool_saturated": false,
  "generation": { ... }
}
```

## Fields

| Field | Meaning |
|---|---|
| `data` | The result rows (was top-level array or `messages`) |
| `total` | Total matching rows, or `-1` when unknown |
| `returned` | Number of rows in this response |
| `offset` | Starting position of this page |
| `has_more` | Whether another page exists |

`total` is exact for keyword search (reported by `SearchFastCount`). It is `-1` for vector/hybrid modes (since the ranking window is bounded and doesn't enumerate all matches), and for the body-FTS fallback path (since no count query exists for full-text search). When `total=-1`, use `has_more` to determine whether to fetch the next page.   [This seems kludgy to me; maybe there's a better way.]

## Behavior changes

- **Breaking:** responses are no longer bare arrays. Clients must read `.data`.
- **Breaking:** vector/hybrid hits move from `.messages` to `.data`. 
  - [or should it have been the other way around?  The shared envelope should follow the existing `.messages` convention?]
- Vector/hybrid now supports `offset` pagination within the ranking window. Previously `offset > 0` was rejected with a `pagination_unsupported` error.
- `list_messages` now detects more pages by fetching one extra row internally rather than running a separate count query.
- Search and list results are now capped at 50 rows per call (previously the shared cap was 1000).

## What comes next

Subsequent PRs will change `get_message` to a narrower windowed body, and a new `search_in_message` tool. Both use the same `offset`/`returned`/`has_more` vocabulary established here, so LLM clients can read and search large messages without the full body ever filling the context window.

### DIFF
--- a/internal/mcp/handlers.go
+++ b/internal/mcp/handlers.go
@@ -482,7 +482,7 @@ func (h *handlers) searchMessagesHybrid(
 	}
 
 	return jsonResult(searchMessagesHybridResponse{
-		paginatedResponse: newPaginatedResponseHasMore(page, offset, hasMore),
+		paginatedResponse: newPaginatedResponseNoTotal(page, offset, hasMore),
 		Mode:              mode,
 		PoolSaturated:     meta.PoolSaturated,
 		Generation: hybridGenerationSummary{
--- a/internal/mcp/handlers.go
+++ b/internal/mcp/handlers.go
@@ -23,7 +23,70 @@ import (
 	"go.kenn.io/msgvault/internal/vector/hybrid"
 )
 
-const maxLimit = 1000
+const (
+	maxLimit               = 1000
+	maxSearchMessagesLimit = 50
+	defaultSearchLimit     = 20
+	// totalCountUnknown is returned when the backend cannot report a full match
+	// count (hybrid/vector ranking depth, or list_messages has_more without a
+	// separate count query). Clients should use has_more for paging.
+	totalCountUnknown = -1
+)
+
+type paginatedResponse[T any] struct {
+	Data     []T   `json:"data"`
+	Total    int64 `json:"total"`
+	Returned int   `json:"returned"`
+	Offset   int   `json:"offset"`
+	HasMore  bool  `json:"has_more"`
+}
+
+func newPaginatedResponse[T any](data []T, total int64, offset int) paginatedResponse[T] {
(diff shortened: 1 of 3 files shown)


wesm:
just pushed some fixes

endolith:
@wesm Sorry, I had Cursor push fixes at the same time.  You can take over.   Do you have opinions about the `.data` vs `.messages` interface or the `-1` stuff?

PR #390, opened 2026-06-18
feat: migrate DuckDB driver to duckdb-go/v2 (DuckDB 1.5.4)

sweenzor:
Swaps the DuckDB `database/sql` driver from the deprecated `github.com/marcboeker/go-duckdb` to its canonical successor `github.com/duckdb/duckdb-go/v2`.

### Why — Homebrew distribution

The goal is to make msgvault installable via `brew install` from **homebrew-core**. homebrew-core won't accept a formula that ships a vendored, precompiled library — the formula must build from source and link the system DuckDB (`depends_on "duckdb"`). With the old `marcboeker/go-duckdb`, that was impossible for two reasons:

1. **Version gap.** It bundled DuckDB 1.1.3, while Homebrew's `duckdb` formula is on 1.5.4 — four minors apart, and a moving target.
2. **Link failure.** Linking against a system `libduckdb` failed on Arrow C-Data-Interface symbols, because that driver pulled the Arrow bridge in unconditionally.

Moving to `duckdb/duckdb-go/v2` resolves both: it bundles **DuckDB 1.5.4** (matching the Homebrew formula), and Arrow is now opt-in behind the `duckdb_arrow` build tag — which this codebase doesn't use — so a system-`libduckdb` build (`-tags duckdb_use_lib`) links cleanly. `marcboeker/go-duckdb` is also deprecated in favor of `duckdb/duckdb-go`, so this is the maintained path regardless.

This PR is only the prerequisite in *this* repo. The Homebrew formula itself lives in a separate repo and is downstream of a release that includes this change.

### What changed
- Bundled DuckDB **1.1.3 → 1.5.4**; arrow-go 18.1.0 → 18.5.1.
- The new module still registers the `duckdb` driver name, so every `sql.Open("duckdb", …)` call site is unchanged — a module-path bump across the blank imports plus `go.mod`/`go.sum` and the Nix `vendorHash`.


wesm:
LGTM, thank you

sweenzor:
Thanks @wesm! Mind cutting a new release with this in it?

PR #393, opened 2026-06-18
Fix Linux release DuckDB link

wesm:
## Summary
- Build Linux release artifacts in Ubuntu 22.04 so DuckDB 1.5.4's prebuilt static libraries link against a compatible C++ runtime.
- Log both GLIBC and GLIBCXX runtime requirements in release builds.

## Test Plan
- Reproduced the Ubuntu 20.04 Linux amd64 link failure locally.
- Built and smoke-tested Linux amd64 in Ubuntu 22.04.
- Built and smoke-tested Linux arm64 in Ubuntu 22.04.
- Ran `git diff --check HEAD~1..HEAD`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,66 @@ on:
   pull_request:
 
 jobs:
+  release-linux-smoke:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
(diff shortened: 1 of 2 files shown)


sweenzor:
I broke this in #390, sorry about that!

PR #395, opened 2026-06-19
ci(release): statically embed libstdc++ and pin the linux build image

sweenzor:
Hardening on top of #393 (which fixed the Linux release break by moving to `ubuntu:22.04`).

- **Statically embed `libstdc++`/`libgcc`** so the released binary has **no runtime `GLIBCXX` dependency**. #393 leaves libstdc++ dynamically linked (the binary requires `GLIBCXX_3.4.30`); this builds with `gcc-12` and forces `-lstdc++` to the static archive.
- **Pin the build image by digest.**
- A guard fails the build if `libstdc++` ever links dynamically.

Scope: `build-linux` only. Verified in a matching `ubuntu:22.04` arm64 container — links cleanly, runs, no dynamic libstdc++ (glibc floor 2.35).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,9 @@ jobs:
             go_arch_name: arm64
     runs-on: ${{ matrix.runner }}
     container:
-      image: ubuntu:22.04
+      # Digest-pinned. ubuntu:22.04 ships GCC 12, needed to link DuckDB 1.5.4's
+      # static lib.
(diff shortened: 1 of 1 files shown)


sweenzor:
Addressed in the latest revision: the release build's GCC toolchain now comes only from Ubuntu's official package repositories, with the build container pinned by digest — no mutable third-party package sources in the release path. (Trade-off: the binary's glibc floor moves 2.31 → 2.35 with the newer base image.)

sweenzor:
Not really needed.

PR #398, opened 2026-06-20
feat: Microsoft Teams ingestion (delegated Graph sync)

njt:
## What

Sync your own Microsoft Teams 1:1/group/meeting chats and channel messages into msgvault via **delegated** Microsoft Graph, searchable alongside mail through the existing TUI / FTS / Parquet analytics.

## Highlights

- New `add-teams` (delegated Graph OAuth) and `sync-teams` (full + incremental, with streamed per-conversation progress) commands; Teams also runs under `serve` scheduled syncs — and the daemon now syncs **all** source types on an identifier (so Teams + Outlook/IMAP on one address both run).
- Reuses the existing chat schema — **no new core tables**: chats → direct/group conversations, channels → channel conversations with root+reply threading, plus reactions, sender + **recipient (`to`) + `@mention`** rows, identity resolution (AAD object id → email dedup, unifying with mail identities), inline images downloaded to content-addressed storage, and shared-file links recorded.
- Incremental sync: chats via `lastModifiedDateTime` list filtering (no delegated per-chat delta endpoint exists), channels via `/messages/delta`; per-conversation cursors persisted in `sync_runs.cursor_after`, **flushed after each conversation** so an interrupted long backfill resumes mid-stream.
- Microsoft OAuth kept independent from the existing Outlook/IMAP manager — separate `teams_<email>.json` token with Graph scopes only, so IMAP and Teams can each be used alone or together.

## Use

1. Register an Entra app with delegated Graph permissions (`Chat.Read`, `ChannelMessage.Read.All`, `Team.ReadBasic.All`, `Channel.ReadBasic.All`, `User.Read`) and grant admin consent.
2. Add to `config.toml`:
   ```toml
   [microsoft]
   client_id = "<app-client-id>"
   tenant_id = "<directory-tenant-id>"
   ```
3. `msgvault add-teams you@tenant.com` then `msgvault sync-teams you@tenant.com` (`--no-channels` / `--limit` for scoped runs). Press `a` inside `msgvault tui` to filter to the Teams account.

## Notes

- Validated on a live tenant: ~313k messages spanning 2017–2026 (chats, channels, reactions, threaded replies) — confirms full-history backfill beyond Graph's 8-month delta window.
- Meeting **transcripts** are intentionally out of scope (separate spec; delegated transcript access is effectively organizer-only). The only remaining follow-up is preserving a departing user's shared SharePoint/OneDrive files before account removal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #399, opened 2026-06-20
fix: normalize importer message/reaction dates to UTC

njt:
## What

Several importers built `time.Time` values from epoch timestamps with `time.Unix`/`time.UnixMilli` but **without `.UTC()`**, leaving them in the runner's local zone — while the rest of each importer stores dates in UTC. Any code reading the calendar day (or the Parquet year partition) is then off by one in zones east of UTC.

Fixes:
- `internal/sync/sync.go` — `processBatch` oldest-message date (progress tracking).
- `internal/whatsapp/mapping.go` — message `SentAt`.
- `internal/whatsapp/importer.go` — reaction `createdAt`.

## Why it matters

`TestProcessBatch_OldestDatePropagation` fails on any machine east of UTC (e.g. NZ): the fixture `2024-01-10T12:00:00Z` reads back as Jan 11 local. The tests are correct; the production code was the bug. Adds `TestMapMessageSentAtIsUTC` (asserts the stored zone is UTC, machine-independent).

## Possible later fixes (out of scope here)

The same `time.Unix(...)`-without-`.UTC()` pattern also appears in the embedding-generation status timestamps, but these are **operator-facing status values** round-tripped from unix-int columns (not message dates), so they don't affect partitioning/dedup/cross-system date semantics. Local-time display is arguably fine; normalizing them to UTC would be a consistency-only follow-up. Sites:
- `cmd/msgvault/cmd/embeddings_manage.go` — `StartedAt`, `SeededAt`, `CompletedAt`, `ActivatedAt`.
- `internal/vector/pgvector/backend.go` — `StartedAt`, `CompletedAt`, `ActivatedAt`.
- `internal/vector/sqlitevec/backend.go` — `StartedAt`, `CompletedAt`, `ActivatedAt`.

Left unchanged here to avoid churning working code on a style call; documented so a future pass can decide.

## Scope

Independent of the Teams PR (#398) — branched from `main`, touches only `internal/sync` and `internal/whatsapp`.

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -216,7 +216,11 @@ func (s *Syncer) processBatch(ctx context.Context, sourceID int64, listResp *gma
 			// Track oldest message date for progress display
 			// Gmail returns messages newest-to-oldest, so oldest shows where we've reached
 			if raw.InternalDate > 0 {
-				msgDate := time.UnixMilli(raw.InternalDate)
+				// .UTC() to match how InternalDate is normalized everywhere
+				// else (the stored message date and parsed.Date both use
+				// .UTC()); without it oldestDate carries the local zone and
+				// callers reading its calendar day are off by one east of UTC.
+				msgDate := time.UnixMilli(raw.InternalDate).UTC()
 				if result.oldestDate.IsZero() || msgDate.Before(result.oldestDate) {
 					result.oldestDate = msgDate
 				}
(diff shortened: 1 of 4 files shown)


wesm:
Great, thank you!

wesm (approved):

wesm:
fixing CI

PR #406, opened 2026-06-22
fix(fbmessenger): don't collapse multiple attachments to one row

njt:
Facebook Messenger messages that carry multiple attachments without downloaded bytes (files missing from the DYI export, or no attachments dir configured) were recording only **one** attachment row — the rest were silently dropped.

`UpsertAttachment` dedupes rows with an empty `content_hash` to one per message. These link/missing attachments now get a stable synthetic `content_hash` derived from the export-relative URI, so siblings coexist and re-imports stay idempotent. `storage_path` stays empty, so no bytes are implied and file-cleanup paths are unaffected.

### DIFF
--- a/internal/fbmessenger/importer.go
+++ b/internal/fbmessenger/importer.go
@@ -697,9 +697,10 @@ func writeThreadToStore(
 			if storagePath != "" {
 				summary.AttachmentsStored++
 			}
-			hash := contentHash
 			syntheticHash := fbAttachmentHash(att, ai)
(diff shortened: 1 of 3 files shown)


njt:
Split out of #398 (Teams ingestion) to keep that PR scoped. Found via an audit of `UpsertAttachment`'s hashless-attachment dedup path. Independent of #398 — branches off current `main`.

wesm:
looking at this

PR #407, opened 2026-06-22
fix(search): empty text/subject terms must not match every message

njt:
An empty search term produced `LIKE '%%'`, which matched **every** message instead of nothing. Two cases:

- **Free-text terms on the LIKE fallback path** (used when FTS errors at runtime, or the binary is built without the `fts5` tag): a tokenless term (empty or punctuation-only) became `LIKE '%term%'`; `""` matched everything. Tokenless terms are now skipped (same predicate the FTS path uses), substituting `FALSE` when none remain — so the fallback mirrors the FTS path.
- **`subject:` operator**: `subject:` or `subject:""` yielded an empty subject filter that matched every message. The parser now drops empty/whitespace `subject:` values (mirroring the existing `label:` handlers); `subject:"!!!"` remains a valid literal substring search.

### DIFF
--- a/internal/search/parser.go
+++ b/internal/search/parser.go
@@ -124,7 +124,14 @@ var operators = map[string]operatorFn{
 		q.BccAddrs = append(q.BccAddrs, normalizeAddr(v))
 	},
 	"subject": func(q *Query, v string, _ time.Time) {
-		q.SubjectTerms = append(q.SubjectTerms, v)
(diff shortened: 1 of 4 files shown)


njt:
Split out of #398 (Teams ingestion) to keep that PR scoped. Found via a pre-existing failing store test (`TestSearchMessagesQuery_TokenlessTextTerms`) while validating that work. Independent of #398 — branches off current `main`.

Issue #387, opened 2026-06-17
Embedding recovery: a single failed enqueue shouldn't require a full rebuild, and the coverage gap should be observable


webgress:
**human:**
## foreword
The following is AI-composed text resulting from a thorough investigation and discussion. My impression of the enqueueing of embedding work is that the tradeoffs in the existing implementation can be improved. There is complexity cost and other compromises that are likely introduced by AI. Below Opus rephrased my view of the problem + solution that we arrived at to simplify the embedding queue and make it more robust.  

This proposes a different compromise: only one generation is actively embedded. While new gen is backfilling new messages are not embedded with old gen. The consistency is achieved after the entire corpus is re-embedded. The design ensures that obsolete or missing embeddings are easy to discover.

**AI:**
## Summary

When a message's embedding **enqueue** fails during sync, the message ends up silently absent from the vector index. Today the only recovery is a **full re-embedding rebuild**, and there is **no durable signal** that the gap exists.

This issue states the problem, then proposes an approach (refined through discussion) that addresses all of it at once — by replacing the separate `pending_embeddings` queue with a **per-message generation field**, which makes the orphan impossible by construction, makes coverage observable, and keeps each scheduled run cheap.

## Current behavior

During sync a message is **persisted first**, then **enqueued** into `pending_embeddings` (for each non-retired generation) as a **separate** step. The embed worker later drains the queue, writes vectors to `embeddings`, and removes the row on completion.

If the enqueue step fails, the message is already committed but never queued — archived and FTS-searchable, but missing from vector/hybrid search.

## The gaps

1. **Recovery is all-or-nothing.** The only way to recover a missed message is a full rebuild (re-seed + re-embed the entire corpus). `O(whole archive)` work to recover what may be a handful of messages.
2. **No durable signal.** The only trace is a transient log warning. An orphaned message is neither in `pending_embeddings` nor in `embeddings`, so the status/pending count reads "caught up" while the message is silently missing.
3. **No lightweight repair is possible without (2).** `pending_embeddings` is a *drained work queue*, not a ledger — "absent from the queue" matches genuinely-missed, already-embedded, *and* legitimately-empty (HTML-only / base64-only / quote-only) messages. You can't tell them apart, which is why recovery is currently a wholesale re-seed.

## Proposed approach

Replace the separate `pending_embeddings` queue with a **per-message generation field** plus a tiny **per-generation watermark**.

[...]

PR #411, opened 2026-06-23
embeddings: replace pending_embeddings queue with per-message embed_gen (scan-and-fill)

webgress:
Implements the embedding-queue redesign proposed in #387.

Replaces the separate `pending_embeddings` work queue with a per-message `embed_gen` column plus a scan-and-fill worker, a per-generation `embed_watermark`, and a full-scan backstop. Backend-agnostic (SQLite and PostgreSQL/pgvector).

**Logic: changing embedding generation is a rare procedure, keeping embeddings other than the target configuration is at most nice to have, not strictly necessary. The complexity required to correctly maintain multiple embeddings during transition is not justified.**

## What changed
- **Schema:** new `embed_gen` column on `messages` (the generation a message is embedded/skipped for; `NULL` = needs work) and a small `embed_watermark` table. The old `pending_embeddings` table is dropped on upgrade.
- **Write path:** new messages persist with `embed_gen = NULL` in the same transaction that stores the message — there is no separate enqueue step, so an embedding can no longer be orphaned by a failed enqueue.
- **Worker:** finds work by scanning `embed_gen IS NULL OR embed_gen <> <target>` (id-ordered) instead of claiming queue rows, and stamps `embed_gen` on success or skip. A periodic full-scan backstop (default 24h, watermark-ignoring) recovers any stragglers.
- **Coverage:** `msgvault embeddings` reports live / embedded / blank / missing counts.

## Upgrade behavior
Existing archives are **not** re-embedded. A one-time, idempotent backfill stamps already-embedded messages with the active generation on first run (both backends); new rows need no backfill.

## Usage
- `msgvault embeddings` — show coverage.
- `msgvault embeddings build [--backstop]` — build/resume embeddings; `--backstop` forces a full watermark-ignoring scan.
- Run only one embedding process at a time.

## API note
This renames the vector-search stats field `pending_embeddings_total` to `missing_embeddings_total` (on `/api/v1/stats` and the MCP `get_stats` tool). Under scan-and-fill there is no pending queue; the value is the count of live messages still needing embedding under the active (and building, if any) generation. No backward-compatibility alias is included — happy to keep `pending_embeddings_total` as a deprecated alias if you'd prefer to preserve the existing field name.


wesm:
looking at this

PR #412, opened 2026-06-23
feat(search): support message_type filters

danshapiro:
Adds first-class message_type filtering across text, API, vector, and hybrid search.

- Parses message_type/type search filters into structured queries.
- Applies message-type filters in store/API search and vector/hybrid backends.
- Documents the filter in command quickstart search help.


wesm:
looking at this

PR #413, opened 2026-06-23
feat(vector): support scoped embedding builds

danshapiro:
Depends on #412. This branch is stacked on the message_type filter PR until that lands.

- Adds vector embedding build scope configuration for selected message types.
- Applies the scope when seeding pending work for SQLite and PostgreSQL vector backends.
- Enqueues embeddings for messages imported by Synctech SMS local/Drive syncs.


wesm:
looking at this

wesm:
I think I'll have to wait to rebase this after #411 lands

Issue #401, opened 2026-06-21
TUI crashed on startup


wichert:
I was running `msgvault sync`, and at the same time started the TUI. This resulted in a SIGBUS when the tui tried to rebuild the cache.

```
❯ msgvault tui
time=2026-06-21T12:37:23.829+02:00 level=INFO msg="msgvault startup" run_id=9d357075d802 command="msgvault tui" argc=0 version=v0.16.0 go_version=go1.26.4 os=darwin arch=arm64 config_path=/Users/wichert/.msgvault/config.toml data_dir=/Users/wichert/.msgvault log_file="" level=INFO
Building analytics cache (22 new messages; 3 deletions; 1 updated messages)...
Full rebuild: clearing existing cache...
Building cache...
  messages...               done (390ms)
  message_recipients...     done (683ms)
  message_labels...         done (48ms)
  attachments...            done (194ms)
  participants...           done (144ms)
  labels...                SIGBUS: bus error
PC=0x103ac5f04 m=14 sigcode=1 addr=0x10a055784
signal arrived during cgo execution
```

<details>
<summary>Full stack trace</summary>

```
goroutine 52 gp=0x5057b3a850e0 m=14 mp=0x5057b3a02008 [syscall]:
runtime.cgocall(0x103a99ab4, 0x5057b3636638)
        runtime/cgocall.go:167 +0x44 fp=0x5057b3636600 sp=0x5057b36365c0 pc=0x102e208c4
github.com/mattn/go-sqlite3._Cfunc__sqlite3_step_row_internal(0x9c6cd5400, 0x5057b389e650, 0x5057b389e658)
        _cgo_gotypes.go:456 +0x30 fp=0x5057b3636630 sp=0x5057b3636600 pc=0x10322a980
github.com/mattn/go-sqlite3.(*SQLiteStmt).execSync.func2(...)
        github.com/mattn/go-sqlite3@v1.14.45/sqlite3.go:2426
github.com/mattn/go-sqlite3.(*SQLiteStmt).execSync(0x5057b38de0a0, {0x5057b38de050?, 0x5057b36366e8?, 0x102e208ec?})
        github.com/mattn/go-sqlite3@v1.14.45/sqlite3.go:2426 +0x90 fp=0x5057b36366c0 sp=0x5057b3636630 pc=0x10323a910
[...]

PR #417, opened 2026-06-25
Avoid mixed SQLite access during TUI startup

wesm:
Issue #401 exposed a TUI startup crash while `msgvault sync` was active: the process could rebuild the Parquet cache through DuckDB while also holding go-sqlite3 connections and starting background FTS maintenance. That repeats the same mixed-SQLite-library risk already documented for daemon cache rebuilds, where DuckDB's bundled SQLite should not attach the live database inside a process that also owns normal msgvault SQLite connections.

This changes interactive startup so the automatic cache build runs before the long-lived TUI store is opened, and FTS maintenance starts only after the query engine is initialized. The DuckDB-backed TUI engine also defaults to direct SQLite fallback for detail reads instead of attaching the live database through `sqlite_scanner`, preserving fast Parquet aggregates without keeping two SQLite implementations attached to the same WAL database in the interactive process.

The regression coverage locks in that TUI DuckDB options keep `sqlite_scanner` disabled by default. Standard Go formatting, vetting, and full tagged tests pass locally.

### DIFF
--- a/cmd/msgvault/cmd/tui.go
+++ b/cmd/msgvault/cmd/tui.go
@@ -83,33 +83,11 @@ Remote Mode:
 			dbPath := cfg.DatabaseDSN()
 			analyticsDir := cfg.AnalyticsDir()
 
-			if !store.IsPostgresURL(dbPath) && !forceSQL && !skipCacheBuild {
-				staleness := cacheNeedsBuild(dbPath, analyticsDir)
-				if staleness.NeedsBuild {
-					fmt.Printf("Building analytics cache (%s)...\n", staleness.Reason)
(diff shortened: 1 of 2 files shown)


wesm:
Addressed the review finding in a28897d8. TUI startup now opens a short-lived store to run InitSchema/startup migrations, closes it before any DuckDB cache export, performs at most one cache build, then reopens the migrated store for the interactive TUI.

The linked CI job failed while exporting Docker Buildx cache (`error writing layer blob: not_found`), not in Go validation. Local `go fmt ./...`, tagged `go vet ./...`, and `make test` all pass after this change.

PR #422, opened 2026-06-26
fix(query): re-probe Parquet schema when cache changes underneath a running engine

Lazare-42:
## What

The DuckDB query engine probes each Parquet table's optional columns once at construction and trusts that snapshot for the rest of the process lifetime. A long-running process (the `serve` daemon, the MCP server) goes stale when `build-cache`/`sync` rewrites the analytics cache with a different column set.

When that happens, the cached "column present" verdict puts a now-absent column into a `SELECT * REPLACE (... AS message_type)` list, which DuckDB rejects:

```
Binder Error: Column "message_type" in REPLACE list not found in FROM clause
```

Every `Search`/`Aggregate` call then fails until the process is restarted.

## Why

Mixed-version writers (or full rebuilds with an older builder) can drop optional columns from the cache. The engine assumed the schema was immutable for its lifetime; it isn't for a server whose cache is rebuilt out of band.

## How

- Fingerprint the cache (per-probed-table file count + size + mtime).
- When the fingerprint changes, re-probe optional columns and re-register the SQL views on demand (cheap no-op on the common unchanged path).
- Guard `optionalCols` with an `RWMutex` since refresh can now race with reads.
- Regression test rebuilds the cache underneath a live engine and asserts the query recovers instead of crashing.

### DIFF
--- a/internal/query/duckdb.go
+++ b/internal/query/duckdb.go
@@ -149,12 +149,9 @@ func NewDuckDBEngine(analyticsDir string, sqlitePath string, sqliteDB *sql.DB, o
 
 	// Probe Parquet schemas for optional columns added in PR #160 (WhatsApp import).
 	// Old cache files may lack these columns; we'll supply defaults in parquetCTEs().
-	engine.optionalCols = map[string]map[string]bool{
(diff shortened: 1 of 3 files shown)


wesm:
looking

PR #426, opened 2026-07-01
Route CLI archive access through daemon HTTP

wesm:
## Summary

This PR changes how the msgvault CLI reaches your archive. Before, every command opened the SQLite database directly in its own process; now, every archive-access command talks to a msgvault daemon over HTTP — either the remote server configured under `[remote]`, or a local background daemon the CLI starts on demand. The daemon becomes the single owner and single writer of the archive, and local and remote usage now behave identically.

New docs: [Daemon Migration Guide](https://msgvault.io/guides/daemon-migration/) (added in this PR) covers the upgrade experience, gotchas, and FAQ.

## Why

msgvault archives are shared by many entry points: scheduled syncs in `msgvault serve`, ad-hoc CLI commands, the TUI, MCP tools, and agent workflows. With direct SQLite access, a CLI command could write to the archive while the daemon was also writing — the source of `database is locked` failures and WAL contention — and remote mode behaved differently from local mode because it used a separate client implementation. Centralizing archive access in the daemon gives one write owner, one policy for concurrency, and one code path for local and remote.

## What users see

**Before → after, day to day:**

- `msgvault stats` (no daemon running) — *before:* opens `msgvault.db` directly. *After:* prints `Starting local msgvault daemon (pid 51234). Logs: ~/.msgvault/serve.log` once, then behaves exactly as before. The daemon idles out after 20 minutes (`[server].daemon_idle_timeout`).
- `msgvault sync-full you@gmail.com` — *before:* the CLI process runs the sync. *After:* the daemon runs it and streams the normal progress output to your terminal; Ctrl+C still cancels, and interrupted syncs keep their checkpoint and resume on the next run.
- Two mutating commands at once — *before:* both open SQLite, second one may fail with `database is locked`. *After:* the second waits with `Waiting: msgvault embeddings build has been running for 42m (Ctrl+C to cancel).` and proceeds automatically when the archive frees up.
- Command while a scheduled sync is running — *before:* lock contention. *After:* the scheduled job checkpoints and yields within a few seconds; it resumes at its next scheduled run.
- Read-only commands (search, stats, TUI, `logs`, `list-deletions`, `embeddings list`, SQL `query`) never wait, even during long operations.
- Binary upgrade — *before:* stale long-running server until manually restarted. *After:* a newer CLI restarts an older local daemon automatically (`[server].daemon_auto_restart = "newer"`, also `never`/`always`).
- OAuth flows (`add-account`, `add-o365`, `add-teams`, `add-synctech-sms-drive`) still open your local browser: authorization runs client-side before the command is handed to the daemon. Remote mode keeps the existing tokens-live-on-the-server model.

**New daemon lifecycle commands:**

```
msgvault serve status | start | stop | restart
```

`status` reports URL, pid, version, API schema, and uptime. Foreground `msgvault serve` is unchanged and never idle-stops.

## Behavior changes to be aware of

- **Environment variables:** commands execute in the daemon process, which does not inherit your shell environment. `MSGVAULT_IMAP_PASSWORD`, `MSGVAULT_ENABLE_REMOTE_DELETE`, and the variable named by `[vector.embeddings].api_key_env` are forwarded from your shell automatically; anything else must be set in the daemon's environment (then `msgvault serve restart`). Unlisted variables are rejected with an explicit error rather than silently dropped.
- **`--local` semantics:** with `[remote].url` configured, `--local` selects the local *daemon*; it no longer means "open SQLite in this process". The daemon owns the archive while running; a direct external writer and the daemon refuse to run concurrently, with error messages pointing at `msgvault serve stop`.
- **New files in `~/.msgvault/`:** `serve.log` (surfaced by `msgvault logs`), lock files (`daemon.lock`, `db.write.lock`, `serve.background.lock`), and a daemon runtime record used for port discovery and version negotiation. The auto-started daemon binds `127.0.0.1` on an auto-selected port; set `[server].api_port` for a stable port. Non-loopback binds still require an `api_key`.
- **Security posture of the gate:** only authenticated requests can occupy or observe the operation gate; unauthenticated requests are rejected by the auth layer without affecting queued work.

## API contract

The HTTP API is now served through Huma with a checked-in OpenAPI document (`api/openapi.yaml`, served at `/openapi.json`, printable via the new `msgvault openapi` command) and a generated Go client under `pkg/client/`. CI rejects stale generated assets. The CLI, TUI, and MCP paths all use one daemon client adapter (`internal/daemonclient`) instead of the previous split remote-client implementation. Client/daemon compatibility is negotiated from the API schema version; remote servers are never auto-restarted by clients.

## Config additions and deprecations

- `[server]`: `daemon_idle_timeout` (default `20m`), `daemon_auto_restart` (`newer`/`never`/`always`), `api_port` now defaults to `0` (auto-select).
- `[analytics]`: `engine` (`auto`/`sql`/`duckdb`) and `auto_build_cache` replace per-invocation flags.
- Deprecated in 0.17.0, removal planned for a later release (still work, print a notice): `tui --force-sql`, `tui --no-cache-build`, `tui --no-sqlite-scanner`, `mcp --force-sql`, `mcp --no-sqlite-scanner`.

## Docs

Adds the [Daemon Migration Guide](https://msgvault.io/guides/daemon-migration/) and a changelog entry, and updates the CLI reference ("HTTP-Backed CLI Behavior", `serve` lifecycle, `openapi`), configuration, API server, remote deployment, and troubleshooting pages.

## Compatibility notes

There are no new database schema or DDL migrations in this PR relative to current `main`. This is still not a no-risk no-op against an old 0.15.x archive, because current `main` already contains post-0.15 schema evolution; testing against a copied archive or backup is the right rollout path.


mariusvniekerk (approved):

mariusvniekerk (approved):

wesm:
Merging this, it's getting more and more difficult to review. I have some more tire-kicking to do before I'm comfortable releasing, so I'll address fixes in follow up PRs

PR #436, opened 2026-07-05
Fix IMAP UID enumeration and raw fetch robustness

io41:
## Observed Failures (sanitized)
Before this patch, IMAP sync produced large numbers of fetch failures on iCloud IMAP.

Error examples:
- `missing raw message in batch result`
- `IMAP fetch skipped after earlier chunk failure`
- `in response-data: in envelope: in env-cc: in address: imapwire: expected SP, got "B"`
- `in response-data: in envelope: in env-cc: in address: imapwire: expected SP, got "S"`

## PR Summary
- Constrain IMAP UID enumeration to `UID 1:*` so servers do not return sequence-number-like values from broad UID searches.
- Preserve per-message IMAP batch-fetch diagnostics instead of collapsing missing/failed fetches into generic nil results.
- Avoid IMAP `ENVELOPE` during raw fetch and label-map Message-ID lookup; parse Message-ID from fetched MIME headers instead.
- Refresh existing label `label_type` values so IMAP system labels such as Sent/Drafts/Trash can recover from stale `user` metadata.

## Root Cause
Some IMAP servers (e.g. iCloud) can return non-fetchable values for unconstrained `UID SEARCH`, and malformed `ENVELOPE` address data can make `UID FETCH` fail even when raw MIME is available. The old sync path collapsed these into generic missing-raw errors, which hid the actual failure modes.

Diagnostic run:
- Same isolated 30k-message window changed from `2045` fetch errors to `0` errors after this patch.

## Repro Context (sanitized)
- Provider: iCloud IMAP
- Host/port: `imap.mail.me.com:993`
- Transport: TLS enabled, STARTTLS disabled
- Source URL shape: `imaps://<redacted-user>@imap.mail.me.com:993`
- Auth: IMAP password/app-password auth
- Test commands used:
  - `MSGVAULT_HOME=<isolated-copy> ./msgvault sync-full <redacted-imap-source> --limit 30000 --noresume`
  - `MSGVAULT_HOME=<isolated-copy> ./msgvault sync-full <redacted-imap-source> --noresume`

## Validation
- `make test`
- `go vet -tags "fts5 sqlite_vec" ./...`
- Isolated sync validation against a 30k email window, changed from 2045 fetch errors before this patch, to 2044 added / 0 errors.


wesm:
Thanks for the PR! I ran a detailed review (with some automated verification against the go-imap v2 sources). Overall the diagnostics refactor and the label-type fix look sound — a few things checked out explicitly before the findings below:

- The ENVELOPE → `BODY.PEEK[HEADER.FIELDS (Message-ID)]` switch is behavior-identical for Message-ID: go-imap v2 already ran the server's ENVELOPE value through the same strict `mail.Header.MessageID()` parser, so dedup/label-merge keys are unchanged, and producer/consumer share the same helper.
- Duplicate FETCH responses can't clobber results — go-imap delivers at most one buffer per UID per FETCH command (`recvUIDSet` dedup in `imapclient/fetch.go`).
- The `label_type` refresh is complete: the rename and upsert branches already set it, and this fills the last gap.

### Findings

1. **`internal/imap/client.go:281` — unconditional `UID 1:*` in every SEARCH.** `enumerateMailboxSearchCriteria` now sends `UID SEARCH UID 1:*` even when there's no watermark (`minUID == 0`), where the old code sent a plain SEARCH. Per RFC 3501, `*` is "the largest UID in use," which is undefined in an empty mailbox; strict servers can answer BAD, and both callers then silently skip the mailbox (`skipping mailbox` warn + continue), omitting its messages from the sync. Suggest only adding the UID key when `minUID > 0` (matching old behavior), or a comment naming the server that needs `1:*` if there is one — this is the only place the PR changes what goes over the wire to every server, and its failure mode is silent.

2. **`internal/imap/batch_fetch.go:162` — `GetMessagesRawBatchWithErrors` is 129 lines with a triple-duplicated failure block.** The chunk-failure sequence (`markRawBatchError(results, chunk, err)` + mark-remaining-as-`errIMAPSkippedAfterChunkFailed` + `break chunkLoop`) is copy-pasted three times; a future fourth failure path will likely diverge and leave entries with neither `Message` nor `Err`. The `if chunkStart+len(chunk) < len(items)` guard is also unnecessary — `markRawBatchError` over an empty slice is a no-op. One small helper fixes all of this. Related: the chunked fetch-with-reconnect loop is now duplicated between this function and `fetchMailboxMessageIDs`.

3. **`internal/imap/client.go:780` — `rawBatchMessagesWithError` is a no-op wrapper.** Both branches return `(messages, err)`; the function reduces to `return rawBatchMessages(results), err`. Suggest inlining it into `GetMessagesRawBatch` and dropping `TestRawBatchMessagesWithErrorPreservesPartialResults`, which locks in the dead branch.

4. **`internal/imap/batch_fetch.go:106` — wasted header parse when there's no `\All` folder.** `rawMIMEMessageID` parses every fetched message's MIME header even when both `c.seenRFC822IDs` and `c.msgIDToLabels` are nil (they're only initialized when an `\All` folder is detected — i.e., most non-Gmail servers never set them). Hoisting the nil-check above the parse avoids a per-message allocation + parse for a value that's never read.

5. **Tautological option-constructor tests.** `TestRawBatchFetchOptionsDoNotRequestEnvelope` (`batch_fetch_test.go`) and `TestMessageIDHeaderFetchOptionsDoNotRequestEnvelope` (`client_test.go`) assert each field of a struct literal returned by a constructor — they restate the implementation and break on any legitimate options change without catching real bugs. The actual contract is already covered by `TestApplyFetchResultsDedupsUsingRawMessageIDWithoutEnvelope` and `TestMessageIDsFromHeaderFetchResultsParsesMessageIDHeaders`; these two can go.

6. **`internal/imap/client.go:438` — per-chunk map alloc + copy.** `messageIDsFromHeaderFetchResults` builds a map per fetch chunk that `fetchMailboxMessageIDs` immediately copies key-by-key into its own map, so every Message-ID is hashed and inserted twice. Having the helper write into the destination map (or return a slice) avoids the churn on large mailboxes.

Minor, not blocking: if `messageIDs` ever contained a duplicate composite ID, one entry would come back `Message == nil, Err == nil` (the `seenReturnedUIDs` guard skips marking it), yielding a misleading generic error record. Verified unreachable with the current unique-by-construction ID list — at most worth a defensive dedup.

Item 1 is the one I'd like resolved (or justified) before merging; 2–6 would make a nice single cleanup commit.

wesm:
Taking a crack at fixing these issues

wesm:
Maintainer follow-up pushed as bb31a4f1, addressing the review findings above:

- **Finding 1 (`UID 1:*`)**: kept the constraint — the PR description's iCloud rationale (sequence-number-like values from unconstrained `UID SEARCH`) justifies it, and it's now documented on `enumerateMailboxSearchCriteria`. The empty-mailbox risk is eliminated by skipping the search entirely when SELECT reports zero messages (`selectedNumMessages`), so `*` is never sent without a referent.
- **Finding 2**: `GetMessagesRawBatchWithErrors` decomposed into `batchMailboxOrder`, `selectBatchMailbox`, `fetchChunk`, and `fetchMailboxBatch`; the triplicated chunk-failure block is now one path, and `fetchMailboxMessageIDs` reuses `fetchChunk` instead of hand-rolling the same reconnect loop.
- **Finding 3**: `rawBatchMessagesWithError` inlined; its test removed.
- **Finding 4**: raw-MIME Message-ID is only parsed when a dedup or label map exists.
- **Finding 5**: the two tautological fetch-options tests removed (behavioral coverage retained).
- **Finding 6**: Message-IDs now collect directly into the destination map (`addMessageIDsFromHeaderFetchResults`).

`make test`, `make lint`, and `go vet` are clean. Behavior is otherwise unchanged from your patch — per-item diagnostics, retry semantics, and the label-type fix are as you wrote them.

wesm:
False positive

io41:
Thank you!

PR #446, opened 2026-07-06
Add agent skills pack with msgvault skills install/uninstall

wesm:
Adds an installable pack of agent skills (SKILL.md files per the open agent-skills standard used by Claude Code and Codex) that teach coding agents the msgvault read-only CLI workflows, so agents don't relearn the CLI each session.

## What's new

- `msgvault skills install` — renders three skills from templates embedded in the binary and installs them into detected agent skill directories (`~/.claude/skills`, `~/.codex/skills`). Flags: `--agent claude|codex` to restrict, `--dir <path>` for an explicit skills root (mutually exclusive with `--agent`), `--force` to overwrite hand-edited copies.
- `msgvault skills uninstall` — removes previously installed `msgvault-*` skills; never touches files lacking the msgvault generation marker.
- Three skills:
  - `msgvault-search` — query operators, fts/vector/hybrid modes, reading messages, exporting raw email
  - `msgvault-attachments` — search → `show-message --json` → export by content hash, jq filtering by type/size
  - `msgvault-analytics` — built-in aggregates plus raw SQL over the Parquet analytics views
- Installed files carry a `generated by msgvault` marker; re-running install refreshes them, hand-edited files are skipped unless `--force`.
- `skills` commands never load archive config or create the msgvault home.
- A drift-guard test resolves every `msgvault …` example in the rendered skills against the real cobra command tree (commands, flags, shorthands), so skill content fails CI if the CLI changes under it.
- New docs guide: `docs/guides/agent-skills.md`.

## Usage

```bash
msgvault skills install            # detect agents, install for all
msgvault skills install --agent claude
msgvault skills install --dir .agents/skills
msgvault skills uninstall
```


mariusvniekerk:
q: should these just be a claude-code / codex skills marketplace ?

wesm:
yes probably

wesm:
**Q: should these just be a Claude Code / Codex skills marketplace?**

Considered, and no — the binary-embedded installer is deliberate. Reasons specific to msgvault:

1. **Version coupling.** The skills document the CLI surface of a specific binary — flags, JSON fields, analytics view schemas. A marketplace releases on its own cadence, so a user on msgvault 0.17 could end up with skills describing 0.19's CLI. `msgvault skills install` can't drift: the binary installs docs for exactly itself, and the drift-guard test (`skills_drift_test.go`) enforces skill/CLI agreement at CI time. A marketplace reintroduces the drift problem this design engineered away.

2. **The dependency points the wrong way.** Marketplace discovery suits self-contained skill packs (e.g. superpowers, which is pure methodology). These skills are useless without the msgvault binary and a synced archive — anyone who can use them already has the binary, so distribution-via-binary reaches the whole audience with one command.

3. **Asymmetric maintenance cost.** Claude marketplace support is one JSON manifest, but Codex parity means portal packaging (rootless archive + OpenAI-owned `openai.yaml` metadata — see how superpowers does it) — a second artifact pipeline to keep in sync forever. The installer is one code path for both agents, plus `--dir` for any other SKILL.md-reading tool.

What a marketplace would buy — browse-time discoverability, marketplace-managed updates, true `msgvault:search` namespacing in Claude — is real but thin, and it layers on cleanly later without changing the architecture: a `.claude-plugin/marketplace.json` (or a tiny CI-generated skills repo) as an optional secondary channel, with `skills install` staying canonical. Deferred until someone actually wants it.

PR #448, opened 2026-07-06
Add deletion staging endpoints to the web API

wesm:
Adds first-class deletion staging endpoints to the daemon HTTP API, so web/HTTP consumers can stage messages for deletion without pre-resolving Gmail IDs client-side (previously only possible via the CLI-proxy route with a fully constructed manifest).

## New endpoints (API schema 1.2.0)

- `POST /api/v1/deletions` — stage messages by filter (`sender`, `sender_name`, `recipient`, `recipient_name`, `domain`, `label`, `source_id`, `after`, `before`) and/or internal `message_ids`; the server resolves Gmail IDs, unions and dedupes them, and writes a pending manifest (`created_by: "api"`). `"dry_run": true` returns the match count and a 10-ID sample without staging. Requests with no criteria are rejected (`400 empty_filter`) — staging the entire archive is not possible. Unknown JSON fields are rejected so a typo'd filter key can't silently widen the selection.
- `GET /api/v1/deletions?status=` — list staged manifests (all statuses by default), newest first.
- `DELETE /api/v1/deletions/{id}` — cancel a pending/in-progress manifest (404 unknown, 409 not cancellable).

Nothing here touches Gmail: execution remains exclusively `delete-staged`.

## Supporting changes

- `GetGmailIDsByFilter` now honors `After`/`Before` on both the SQLite and DuckDB Parquet engines.
- New engine method `GetGmailIDsByMessageIDs` resolves internal IDs with the same live-message + Gmail-source scoping as the filter path, so deleted/non-Gmail/dedup-hidden messages can never be staged.
- Manifests gain a `raw_filter` provenance field recording the verbatim staging request; generated manifest IDs now carry a random suffix so same-second requests with the same description cannot overwrite each other.
- `pkg/client` gains a `StageDeletion` wrapper accepting both success statuses (200 dry-run, 201 create).

The new endpoints are documented in `docs/api-server.md`, with a changelog entry under Unreleased.


wesm:
Addressed the High finding in a8c9168d: staging now resolves the account(s) owning the selected Gmail IDs via a new `GetAccountsByGmailIDs` engine method (live Gmail-source scoped, on both SQLite and the Parquet fallback). Selections spanning multiple Gmail accounts are rejected with `400 multi_account_selection`; single-account selections stamp `Filters.Account` on the manifest so `delete-staged` targets the right mailbox, and the account is reported in both dry-run and create responses. Covered by engine tests (multi-account, non-Gmail, deleted exclusions), handler tests, and the e2e test now asserts the persisted manifest carries the account.

PR #455, opened 2026-07-07
Add Beeper Desktop archive core: internal/beeper importer package

sweenzor:
Part 1 of 3 of a Beeper Desktop integration: archives every chat network bridged through [Beeper Desktop](https://developers.beeper.com/desktop-api/) — WhatsApp, Signal, Telegram, Instagram, LinkedIn, X, Facebook, Beeper/Matrix — via its documented local REST API on `localhost:23373`. This PR is the standalone `internal/beeper` package with tests; part 2 wires CLI/config/scheduler/docs, part 3 adds attachment download.

## What changed

- **Read-only API client**: exposes only GET endpoints by construction, so the archiver can never mutate Beeper state. Rate-limited (default 20 QPS), retries 429/5xx honoring `Retry-After`, actionable errors for 401 (mint a token) and connection-refused (Beeper not running).
- **Importer**: one msgvault source per Beeper network account (`source_type="beeper"`, identifier = Beeper accountID) so each network gets its own account filter and stats. Full-history backfill walks newest→oldest with resumable checkpoints every 25 pages (a single chat can exceed a million messages); later runs are cursor-driven incremental, discovering active chats via `lastActivityAfter`. A bounded 24-hour head re-walk catches in-place edits, deletions, and reaction changes that a forward-only cursor cannot observe.
- **Data mapping**: messages persist as `message_type="beeper"` with FTS text (including voice-note transcriptions), mention recipients, reactions, reply links, and soft-deletion tombstones — content archived before a network deletion stays archived. The verbatim API JSON is stored as `beeper_json`.
- **Identity**: participants resolve through an exclusive ladder — E.164 phone → email → Beeper user ID — deduping people against existing SMS/WhatsApp/mail archives.
- **Safety**: Beeper message IDs are only unique per installation; the sync state stores an anchor probe and a reinstall/re-index fails fast with instructions instead of silently duplicating the archive.
- **Store**: new `SetMessageEdited` helper (no writer for `is_edited` existed); `GetLastSuccessfulSync` now tie-breaks same-second completions by `id` so back-to-back syncs load the correct cursor.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #458, opened 2026-07-08
Add Beeper Desktop chat archiving

sweenzor:
Archives every chat network bridged through [Beeper Desktop](https://developers.beeper.com/desktop-api) — WhatsApp, Signal, Telegram, Instagram, LinkedIn, X, Facebook, Matrix — via its local REST API. Supersedes #455 as a single consolidated PR.

## Usage

```bash
msgvault add-beeper       # access token from Beeper Desktop → Settings → Developer
msgvault sync-beeper      # first run backfills full history; later runs are incremental
```

Each connected network account becomes its own source (`signal`, `whatsapp`, …), so networks stay separately filterable in the TUI and search; all rows share `message_type = "beeper"`.

## What's stored

Message text with FTS (voice-note transcriptions included), reactions, replies, mentions, group membership, deletion tombstones (archived content survives network deletions), media in the content-addressed attachment store, and the verbatim API JSON per message (`raw_format = beeper_json`).

## Behavior notes

- The client is GET-only by construction — the archiver can never write to Beeper.
- Backfills checkpoint continuously and resume after interruption; `--full` re-fetches and upserts in place.
- The live API's pagination quirks (`hasMore` never goes false; re-served pages at history edges) are handled and regression-tested against recorded behavior.
- Beeper message IDs are only unique per installation: an anchor probe fails fast after a reinstall/re-index instead of silently duplicating the archive.
- `[beeper]` config covers daemon scheduling, per-network include/exclude (e.g. exclude `whatsapp` when using the native importer), and a media size cap; `backfill-beeper-media` retries failed downloads.

## Store/query changes

- `SetMessageEdited` helper (nothing wrote `is_edited` before) and a same-second tiebreak in `GetLastSuccessfulSync`.
- The three attachment-replace methods now share one helper; media metadata lands in the insert.
- Texts-mode SQL type filters derive from `TextMessageTypes` instead of five hand-maintained literals, and the TUI snippet fallback applies to all text types (was WhatsApp-only).

Verified against a live Beeper Desktop with 8 accounts: full backfill, incremental re-runs, media download, interrupt/resume.

### DIFF
--- a/cmd/msgvault/cmd/add_beeper.go
+++ b/cmd/msgvault/cmd/add_beeper.go
@@ -1,0 +1,195 @@
+package cmd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/mattn/go-isatty"
+	"github.com/spf13/cobra"
+	"go.kenn.io/msgvault/internal/beeper"
+	"go.kenn.io/msgvault/internal/clirun"
+)
+
+var (
+	addBeeperTokenFile         string
+	noDefaultIdentityAddBeeper bool
+)
+
+func newAddBeeperCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-beeper",
+		Short: "Add Beeper Desktop as an archive source",
+		Long: `Add Beeper Desktop as an archive source.
+
+Beeper Desktop bridges many chat networks (WhatsApp, Signal, Telegram,
+Instagram, LinkedIn, X, Facebook, ...) into one local app with a read-only
+API on localhost. Each connected network account becomes its own msgvault
+source, so networks stay separately filterable and searchable.
+
+Requires Beeper Desktop to be running on this machine. Mint an access token
+in Beeper Desktop (Settings > Developer) and provide it via --token-file,
+the MSGVAULT_BEEPER_TOKEN environment variable, or the interactive prompt.
+
+Use [beeper].accounts / exclude_accounts in config.toml to limit which
+networks are archived (e.g. exclude "whatsapp" if you already archive it
+with import-whatsapp).
+
+Examples:
+  msgvault add-beeper
+  msgvault add-beeper --token-file ~/beeper-token.txt
+  MSGVAULT_BEEPER_TOKEN="..." msgvault add-beeper`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !isDaemonCLISubprocess() {
+				token, err := readAddBeeperToken(cmd)
+				if err != nil {
+					return err
+				}
+				// The Beeper Desktop API is loopback-only: validate locally
+				// when the daemon shares this machine so problems fail fast;
+				// with a remote daemon, validation happens daemon-side against
+				// the Beeper Desktop running there.
+				if IsRemoteMode() {
+					_, _ = fmt.Fprintln(cmd.OutOrStdout(), "Remote daemon configured: it must run beside its own Beeper Desktop, which will validate the token.")
+				} else {
+					accounts, err := beeperClient(token).ListAccounts(cmd.Context())
+					if err != nil {
+						return err
+					}
+					_, _ = fmt.Fprintf(cmd.OutOrStdout(), "Connected to Beeper Desktop (%d accounts)\n", len(accounts))
+				}
+				return runDaemonCLICommandHTTPFromCobraWithEnv(cmd, args, map[string]string{
+					clirun.EnvBeeperToken: token,
+				})
+			}
+
+			token := os.Getenv(clirun.EnvBeeperToken)
+			if token == "" {
+				return errors.New("missing Beeper token in daemon subprocess (set MSGVAULT_BEEPER_TOKEN)")
+			}
+			accounts, err := beeperClient(token).ListAccounts(cmd.Context())
+			if err != nil {
+				return err
+			}
+			if err := beeper.SaveToken(cfg.TokensDir(), token); err != nil {
+				return fmt.Errorf("save beeper token: %w", err)
+			}
+
+			s, cleanup, err := openWritableStoreAndInitForIngest()
+			if err != nil {
+				return err
+			}
+			defer cleanup()
+
+			added := 0
+			for _, acct := range accounts {
+				if !cfg.Beeper.AccountIncluded(acct.AccountID) {
+					_, _ = fmt.Fprintf(cmd.OutOrStdout(), "  Skipping %s (filtered by [beeper] config)\n", acct.AccountID)
+					continue
+				}
+				source, err := s.GetOrCreateSource(sourceTypeBeeper, acct.AccountID)
+				if err != nil {
+					return fmt.Errorf("create source for %s: %w", acct.AccountID, err)
+				}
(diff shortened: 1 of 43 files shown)


sweenzor:
Addressed the roborev findings in dcd966a, each with a regression test:

- **High — anchor never re-armed**: runs now select a replacement anchor from recently active chats before completing, so the reinstall guard can't be persisted disabled.
- **Medium — unbounded asset reads**: `max_media_mb` is enforced during the read (Content-Length check + limited reader); declared `fileSize` is treated as untrusted.
- **Medium — transient reaction-target failures**: only 404 is skipped; other errors leave the cursor un-advanced so the event retries next run.
- Also from the first review round: reply pairs are always buffered (newest-first pages can carry a reply before its parent) and unfinished backfills persist unresolved pairs in the sync state across resumes.

sweenzor:
Round-2 findings addressed in the latest commit, both with regression tests: mid-chat checkpoints now persist buffered reply pairs together with the advanced cursor (tested across both interrupt paths), and messages whose media was removed at the source clear their stale attachment rows/pending markers on re-persist.

sweenzor:
Round-3 findings addressed in 77ebf10: BackfillMedia now skips only on 404 (transient fetch failures are recorded as errors and keep their markers — regression-tested), and pending-attachment messages are buffered into a slice before the per-item download/write work instead of holding a read cursor open across it.

wesm:
Thank you for working on this, I am not a Beeper user but this is a good opportunity to set it up!

sweenzor:
> Thank you for working on this, I am not a Beeper user but this is a good opportunity to set it up!

Happy to! I like beeper a lot, good options to run your own bridges, self-host etc. And I generally like breaking open walled-garden chat products.

sweenzor:
Round-4 findings addressed in the latest commit, both regression-tested: a 404'd source message now clears its pending media marker (no more eternal retries), and bare-ID participant cache entries are upgraded when phone/email metadata appears later in the run, preserving cross-archive dedup.

sweenzor:
Round-5 findings addressed in the latest commit, all regression-tested: backfill-beeper-media now runs the same anchor verification as the main sync before trusting stored message IDs; the gone-message cleanup clears only pending markers and preserves downloaded attachment rows; and anchor re-arming falls back to chats known from prior sync state when a run enumerates none.

sweenzor:
Round-6 findings addressed in the latest commit, both regression-tested: backfill-beeper-media re-arms a soft-cleared anchor before persisting its state (it becomes the newest completed baseline), and the anchor chat-probe only reports a reinstall on 404 — transient probe failures now fail the run retryably instead of prompting a remove/re-add.

sweenzor:
Round-7 finding addressed in the latest commit: the sync-beeper CLI loop now collects per-account errors and continues with the remaining networks, rebuilds the cache regardless, and reports failures aggregated — matching the sync/sync-full convention and the scheduler path.

sweenzor:
Round-8 findings addressed in the latest commit: the reinstall guard now keeps up to three anchor probes in distinct chats, so deleting any one anchored chat is ordinary churn while a reinstall (which invalidates all of them) still fails fast — regression-tested for both cases. Transient media-backfill fetch failures now count their kept markers as still-pending and the CLI surfaces the error count.

sweenzor:
Round-9 finding addressed in the latest commit: both sync and media-backfill runs force a final checkpoint with their closing counters before CompleteSync, so quick runs no longer show stale/zero counts in sync diagnostics (asserted in the e2e test).

Separately: the red **test** check is not this PR — govulncheck started flagging two new Go stdlib CVEs (GO-2026-5856, GO-2026-4970) fixed in go1.26.5, with all traces through pre-existing code (imap, api, backup, emlx). Every branch will fail that step until CI's Go is bumped from 1.26.4; happy to send a separate PR for the toolchain bump if useful.

sweenzor:
Round-10 finding addressed in the latest commit: since chat/room IDs survive a reinstall, losing every anchor probe is no longer excused by the chats still existing — recently archived messages now arbitrate. If one resolves at the source with a matching timestamp the losses were churn (e.g. disappearing-message chats); otherwise the run fails as a reinstall before trusting any stored cursor. Covered by the existing churn/reinstall tests plus enriched media fixtures.

sweenzor:
Round-11 finding addressed in the latest commit: phone/email-resolved participants now persist their Beeper user ID as an identifier (re-pointing any existing bare-ID row), so bare sender IDs in later runs resolve to the same person across cache resets — regression-tested in both directions (rich-then-bare and weak-then-upgraded).

sweenzor:
Round-12 finding addressed in the latest commit: verification no longer fast-paths the zero-anchor case — with no probes, recently archived messages are verified instead (no-op for fresh accounts), so the reinstall guard cannot lapse for an account with history even if a prior re-arm failed. Regression test covers both the passing (intact install, re-arms) and blocking (re-assigned IDs) sides.

sweenzor:
Round-13 finding addressed in the latest commit: when a rich (phone/email) resolution discovers the Beeper user ID is owned by a weak bare-ID participant, that participant's messages, reactions, mentions, and membership are merged into the rich row (constraint-safe dedup) and the fork deleted — guarded so a previous owner with real contact metadata is never absorbed. Verified on SQLite and PostgreSQL.

sweenzor:
Rebased onto #460 (Go 1.26.5 bump) and latest main, so CI should now be fully green including the govulncheck gate. The go.mod/flake.nix bump commit will drop out of this diff once #460 merges.

sweenzor:
Rebased onto main now that #460 is merged (the toolchain commit drops out of this diff), and pushed a holistic cleanup pass over the whole PR — a structural review for readability after the round-by-round fixes, plus the round-15 finding:

- **Round 15 (phone-over-email)**: rich resolutions now track their rung, so a phone sighting upgrades a cached email-based resolution and folds the email participant into the phone-deduped one; MergeParticipants preserves contact columns on the survivor. Regression-tested both ways (upgrade merges; a contact-bearing previous owner is never absorbed).
- **Structure**: package doc; the reinstall guard now lives in anchors.go (with its tests in anchors_test.go) so it reviews as one unit; large functions split along their natural seams (ensureConversation, refreshReactionTarget, persistMentions/persistReactions).
- **Simplifications**: the per-anchor GetChat probe is gone (the archived-sample arbiter already decides the no-survivor case with stronger evidence); reply pairs live only in the checkpointed chat state, deleting the mirror machinery; dead exported surface removed.
- **Tests**: new coverage for the discovery fallback when an unfinished chat disappears, incremental stuck-cursor liveness, the watermark-hold-on-fetch-error guard, and the media-type mapping table. Package coverage 80.2% → 82.7%; suite verified on SQLite and PostgreSQL.

sweenzor:
Round-16 finding addressed in the latest commit: the merge guard now blocks only on phone-bearing prior owners, so a cross-run email-to-phone upgrade folds the email participant's history into the phone-deduped one (an email-only owner of the same Beeper user ID is the same person seen with weaker metadata). The email rung also never downgrades a phone-resolved identity. Fresh-resolver cross-run test covers the merge, the preserved email, and the no-downgrade direction.

wesm:
Good stuff. I am going to try to get #464 merged first and then rebase this, bear with me

wesm:
working on this

PR #459, opened 2026-07-08
import-emlx: import *.partial.emlx files instead of skipping them

### LINKED ISSUE #457: import-emlx silently skips *.partial.emlx files (body fully present; only attachments uncached)
### Summary

`import-emlx` silently skips every file named `*.partial.emlx`. In Apple Mail, a `.partial.emlx` file is **not** a truncated/corrupt message: for Gmail/IMAP accounts it is the normal on-disk form of a message whose **body is fully downloaded but whose attachments are not cached locally**. The RFC822 payload (headers + all `text/plain` and `text/html` parts) is complete and readable. By dropping these, `import-emlx` omits large portions of an account with no warning.

### Impact

On this machine (a single Google Workspace + a personal IMAP account), partials vastly outnumber full `.emlx`:

| account store | full `*.emlx` | `*.partial.emlx` |
|---|---|---|
| Google Workspace INBOX | 12,266 | **62,712** |
| personal IMAP (whole store) | 6,098 | **14,627** |

So the archive is silently missing the majority of these mailboxes. The user only noticed because a specific, clearly-received email could not be found via search.

### Reproduction (v0.17.0, macOS arm64)

Take any real `*.partial.emlx`, import it under two names into two throwaway DBs:

```
# A: keep the .partial.emlx name
.../INBOX.mbox/<UUID>/Data/3/1/5/Messages/513139.partial.emlx
msgvault --home /tmp/homeA import-emlx <stageA>/INBOX.mbox --identifier me@example.com
#  -> Mailboxes: 0 discovered, 0 imported ; Messages: 0

# B: same bytes, renamed to .emlx
.../INBOX.mbox/<UUID>/Data/3/1/5/Messages/513139.emlx
msgvault --home /tmp/homeB import-emlx <stageB>/INBOX.mbox --identifier me@example.com
#  -> Mailboxes: 1 discovered, 1 imported ; Messages: 1
#  search confirms full subject/sender/body imported correctly
```

Identical bytes, identical folder structure, identical `--identifier`; the only difference is the filename suffix. The partial-named file is not even *discovered* (the enclosing mailbox reports 0 discovered when it contains only partials).

### Verifying the body is complete

A `.partial.emlx` here contains a well-formed `multipart/alternative` with both `TEXT/PLAIN` and `TEXT/HTML` parts and a properly-closed MIME boundary, followed by the Apple Mail metadata plist. Only the attachment parts are absent (Mail fetches them on demand). So there is real, searchable content being discarded.

### Suggested fix

Include `*.partial.emlx` in the `import-emlx` discovery walk and parse them like normal messages:

- Derive the dedup/message identity from the base name (strip the `.partial` infix) so a later fully-downloaded `.emlx` of the same message dedupes against the partial rather than duplicating it.
- Import the body normally; treat missing attachment parts as best-effort (equivalent to `--no-attachments` for those parts) rather than a reason to skip the whole message.
- Optionally surface a count of partials imported (and attachments deferred) in the import summary.

Happy to help test a patch against these real mailboxes.

wesm:
Fixes #457.

- `import-emlx` now discovers and imports `*.partial.emlx` files. Apple Mail uses this form for IMAP/Gmail messages whose attachments are not cached locally; the RFC822 body is complete, so the message imports normally with only the uncached attachment parts absent. Mailboxes containing only partial files are now discovered instead of reporting 0 mailboxes.
- When both `N.emlx` and `N.partial.emlx` exist in the same directory, the fully-downloaded copy supersedes the partial.
- The import summary reports a `Partial files` count when any were imported.
- A partial re-imported later as a full `.emlx` is stored separately (content-hash identity); `msgvault deduplicate` merges such pairs by RFC822 Message-ID.

### DIFF
--- a/cmd/msgvault/cmd/import_emlx.go
+++ b/cmd/msgvault/cmd/import_emlx.go
@@ -371,6 +371,7 @@ func importAutoAccounts(
 		grandTotal.MessagesAdded += summary.MessagesAdded
 		grandTotal.MessagesUpdated += summary.MessagesUpdated
 		grandTotal.MessagesSkipped += summary.MessagesSkipped
+		grandTotal.PartialFiles += summary.PartialFiles
 		grandTotal.Errors += summary.Errors
 		if summary.HardErrors {
 			grandTotal.HardErrors = true
@@ -445,6 +446,12 @@ func printImportStats(out io.Writer, summary importer.EmlxImportSummary) {
 		"  Skipped (dup):  %d messages\n",
 		summary.MessagesSkipped,
 	)
+	if summary.PartialFiles > 0 {
+		_, _ = fmt.Fprintf(out,
+			"  Partial files:  %d (body imported; attachments not cached by Apple Mail)\n",
+			summary.PartialFiles,
+		)
+	}
 	_, _ = fmt.Fprintf(out,
 		"  Errors:         %d\n",
 		summary.Errors,
--- a/docs/usage/importing.md
+++ b/docs/usage/importing.md
@@ -196,8 +196,8 @@ Both layouts are supported. The importer discovers all `.mbox` and `.imapmbox` d
 
 Apple Mail stores its data at `~/Library/Mail/` on macOS. The auto-discover mode reads `~/Library/Accounts/Accounts4.sqlite` (the macOS accounts database) to map V10 directory GUIDs to email addresses. You can also use a Time Machine backup or a copy of the Mail directory from another machine.
 
-!!! warning
-    The `.partial.emlx` files that Apple Mail creates for incomplete downloads are automatically skipped during import.
+!!! note
+    Apple Mail stores IMAP and Gmail messages whose attachments have not been downloaded as `.partial.emlx` files. The message body in these files is complete, so they are imported normally — only the uncached attachment parts are absent. When both `N.emlx` and `N.partial.emlx` exist for the same message, the fully-downloaded copy is used. The import summary reports how many partial files were imported.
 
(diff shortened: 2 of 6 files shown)


wesm:
Re the Medium finding (`emlx_import.go:482`): the limitation is real and intentional — noted in the PR description — but neither suggested fix is safe as an import-time change:

- **Filename-derived identity** would break the existing cross-mailbox dedup: identical bytes in INBOX and Archive currently collapse to one message with both labels via the content-hash `sourceMsgID`; per-file identity splits them. It's also not stable across Apple Mail mailbox rebuilds (message files get renumbered), and cross-run upgrades would need a file→message provenance table that doesn't exist.
- **Import-time Message-ID dedup** conflicts with the design in `internal/dedup`: Message-ID merging is deliberately confined to the explicit, scoped `deduplicate` command (see the package doc — legitimate distinct copies share Message-IDs, and merging there is reversible with survivor-selection rules). Replacing an existing message's raw during import would also require replace-and-reparse machinery the ingest path doesn't have.

The pair the finding describes *is* grouped by `msgvault deduplicate` today (same source, same Message-ID). The actual gap is survivor selection: `isBetter` in `internal/dedup/dedup.go` tiebreaks on older `ArchivedAt`, so the earlier-imported partial survives and the attachment-complete copy is pruned. The targeted follow-up is to prefer the copy with attachments (larger raw / more attachment rows) ahead of the ArchivedAt tiebreak — a dedup-engine change that needs its own tests and spec update, tracked separately from this PR's fix for the silent omission.

PR #460, opened 2026-07-08
Bump Go to 1.26.5 (go.mod + nix pin)

sweenzor:
go1.26.5 fixes GO-2026-5856 (crypto/tls) and GO-2026-4970 (os), which govulncheck now flags on every branch — CI's test job fails at the vulnerability gate repo-wide. The nix flake pins the Go source tarball by hash, so it moves together with the go directive.


wesm:
looking at the failure

PR #462, opened 2026-07-09
Enforce metadata/body search scopes with native contexts

endolith:
Split MCP keyword search into explicit metadata and body paths. `search_messages` now searches subjects, snippets, senders, and recipients when mode is omitted while retaining vector and hybrid modes; `search_message_bodies` requires free text and returns up to five bounded, backend-native FTS excerpts per hit, with truncation flagged when request-wide work limits omit context.

Enforce the same scopes across SQLite, PostgreSQL, DuckDB, and daemon-backed clients. PostgreSQL uses a versioned weighted FTS layout and native context selection, and body changes invalidate stale index documents. `/api/v1/search/deep?scope=body` (schema 1.3.0) returns exact body hits plus ID-keyed `body_contexts` without changing the generated `MessageSummary` element type.

Keep metadata results, counts, and statistics aligned for Unicode matching and typed filters; reject unsupported or malformed operators before dispatch; add `conversation_id` to MCP `list_messages`; and update the MCP guides and changelog.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -209,6 +209,23 @@ components:
     BackupFreezeEndResponse:
       additionalProperties: true
       type: object
+    BodySearchContext:
+      additionalProperties: true
+      properties:
+        context_snippets:
+          items:
+            type: string
+          type:
+            - array
+            - "null"
+        context_snippets_truncated:
+          type: boolean
+        message_id:
+          format: int64
+          type: integer
+      required:
+        - message_id
+      type: object
     BuildingSummary:
       additionalProperties: true
       properties:
@@ -1099,6 +1116,12 @@ components:
     DeepSearchResponse:
       additionalProperties: true
       properties:
+        body_contexts:
+          items:
+            $ref: "#/components/schemas/BodySearchContext"
+          type:
+            - array
+            - "null"
         count:
           format: int64
           type: integer
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1118,6 +1118,8 @@ components:
           type: integer
         query:
           type: string
+        scope:
+          type: string
       required:
         - query
         - messages
@@ -2389,7 +2391,7 @@ components:
       type: apiKey
 info:
   title: msgvault API
-  version: 1.2.0
+  version: 1.3.0
 openapi: 3.1.0
 paths:
   /api/ping:
@@ -4654,6 +4656,11 @@ paths:
           required: true
           schema:
             type: string
+        - description: "Exact search scope: body; omit for composite full-text search"
+          in: query
+          name: scope
+          schema:
+            type: string
         - description: Sender email/address filter
           in: query
           name: sender
@@ -4773,7 +4780,7 @@ paths:
           description: Error
       security:
         - apiKey: []
-      summary: Run deep aggregate search
+      summary: Run full-text message search
       tags:
         - API
   /api/v1/search/domains:
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,14 +17,16 @@ All notable changes to msgvault, grouped by release.
 - MCP `list_messages` accepts a `conversation_id` filter for listing one
   conversation or thread.
 - The daemon HTTP API schema is now 1.3.0 and supports `scope=body` on deep
-  search so daemon-backed clients can require an exact body-only response.
+  search so daemon-backed clients can require an exact body-only response;
+  bounded excerpts are returned in an ID-keyed `body_contexts` companion.
 
 **Improvements**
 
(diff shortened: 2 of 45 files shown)


endolith:
Clarification on why we proposed to do it this way:

----

The split was a design choice to make the two search paths **explicit and predictable** for the LLM, rather than overloading one tool with a `scope`/`mode` option.  From the PR plan in the conversation notes (and the current code), the original idea was to separate "find relevant messages" from "search inside message bodies" because they have different return shapes and expectations.

## Why two functions instead of one with more options

The original `search_messages` already had a `mode` parameter for `fts`, `vector`, and `hybrid`.  The `fts` mode in the old branch was actually doing **two different searches**:

1. Run `SearchFast` (metadata/subject/sender).
2. If that returned nothing and the query had free text, fall back to `Search` (full FTS, including bodies).

That meant the same tool call could return either a cheap metadata result with an exact `total`, or a body-FTS result with `total=-1`, and you could not tell which you were getting.  It also meant the tool sometimes returned body-matching messages and sometimes did not, depending on whether metadata happened to match.

The split into `search_messages` and `search_message_bodies` replaces that hidden fallback with an explicit choice:

| Tool | What it searches | What it returns | Total |
|------|------------------|-----------------|-------|
| `search_messages` (no `mode`) | `SearchFast` — subject, snippet, sender, labels, dates | `MessageSummary` list | Exact (`SearchFastCount`) |
| `search_messages` (`mode=vector/hybrid`) | vector/hybrid ranking | `MessageSummary` list | `-1` (ranking window) |
| `search_message_bodies` | `Search` — full FTS over bodies | `MessageSummary` + `context_snippets` | `-1` (use `has_more`) |

In other words, the `mode` option is already used for **ranking strategy** (keyword vs. semantic vs. combined).  Re-using it for **search scope** (metadata vs. body) would make the tool have two different axes of behavior in one parameter, and the response shape would change depending on which axis you picked.  Separate tools make the contract clearer:

- `search_messages` = "find messages by metadata or semantic relevance."
- `search_message_bodies` = "find the literal words inside bodies and show me the snippets."

The notes also phrase the intended workflow as a **pipeline**:

1. `search_messages` / `search_message_bodies` → find candidates + short excerpts  
2. `search_in_message` → locate exact offsets inside one message  
3. `get_message` with `center_at` → read a larger window around the match  

Each tool has one job, so the LLM knows which one to call when.

---

**Clarifying the other two modes**

`search_messages` has three `mode` values today:

- `mode=vector` / `mode=hybrid` — these go through `searchMessagesHybrid` (line 512), which calls the vector/hybrid engine, gets ranked message IDs, and then bulk-hydrates summaries via `GetMessageSummariesByIDs`.  They do **not** call the SQL `Search`/`SearchFast` methods, so they do not have the metadata-vs-body ambiguity.  They do not fetch bodies, and they return `total=-1`.
- No mode (or the old `mode=fts`) — this is the path that now calls `searchMessages` (line 328), which uses `SearchFast` and `SearchFastCount`.  This is the only mode that the upstream review is flagging.

So the ambiguous "metadata search that can still match body" behavior only applies to the no-mode/default path, which used to be `mode=fts`. The `vector`/`hybrid` modes are not the problem here.


----

(And then there's a future branch [split/mcp4-vector-chunk-matches](https://github.com/endolith/msgvault/tree/split/mcp4-vector-chunk-matches) that would add vector chunk matches as a unified matches[] array to `search_messages` (vector/hybrid), `search_in_message`, and `search_message_bodies`.)

endolith:
I'm not sure about the split being enforced by the query layer:

> `query.Engine` is not MCP-only. It is used by the HTTP API, CLI, TUI, daemon, and tests. Specifically:
> 
> - `Search` is used by:
>   - `internal/mcp/handlers.go` — `search_message_bodies`
>   - `internal/api/handlers.go` — HTTP deep-search endpoint
>   - `internal/api/cli_handlers.go` — CLI search
>   - `internal/tui/model.go` — TUI deep search
>   - plus many tests
> 
> - `SearchFast` is used by:
>   - `internal/mcp/handlers.go` — `search_messages` (metadata default)
>   - `internal/mcp/handlers.go` — staged-deletion search
>   - `internal/daemonclient/engine_adapter.go` — daemon search
>   - `internal/query/sqlite.go` — `SearchFastWithStats`
>   - plus many tests
> 
> So changing `Search` to be body-only or `SearchFast` to be metadata-only would break the HTTP API, CLI, and TUI. The safest fix is to **leave `Search` and `SearchFast` as the general-purpose engines** and add new MCP-specific query paths, or to add a `scope` parameter internally that the MCP handlers set without altering the public `Engine` interface. If we really want to split the engine semantics, we would need to update all those callers and likely rename/replace the methods.
>

wesm:
Let me take a look

PR #463, opened 2026-07-09
feat: serve attachment content via GET /api/v1/attachments/{hash}

robelkin:
## What

Adds `GET /api/v1/attachments/{hash}` — an authenticated endpoint that streams a stored attachment's raw bytes, addressed by its SHA-256 content hash, with the attachment's own `Content-Type` and a download `Content-Disposition` filename.

## Why

The HTTP API had no way to retrieve file-attachment content. `/messages/{id}/inline` serves only inline images (it requires a `cid` and an `image/*` content type), and attachment bytes were otherwise reachable only from the co-located CLI/MCP paths that read the on-disk store directly. A networked client — a remote TUI, or any integration without filesystem access to the attachments directory — had no route to fetch them.

## Usage

```
GET /api/v1/attachments/{sha256-hex}
Authorization: Bearer <api-key>
```

- `200` with the file bytes and `Content-Type` / `Content-Disposition` / `Content-Length` set.
- `400` for a malformed hash (not 64 hex chars; also guards path traversal).
- `404` when the attachment metadata row or its on-disk file is absent (e.g. an account removed with its attachments, or an archive built with `--no-attachments`).

Auth is inherited from the existing `/api/v1` API-key middleware. `content_hash` is not unique (content-addressed dedup), so any matching row supplies the header metadata.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -2705,6 +2705,66 @@ paths:
       summary: Get nested aggregate rows
       tags:
         - API
+  /api/v1/attachments/{hash}/content:
+    get:
+      operationId: getAttachmentContent
+      parameters:
+        - description: Attachment SHA-256 content hash
+          in: path
+          name: hash
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          content:
+            application/octet-stream:
+              schema:
+                contentMediaType: application/octet-stream
+                format: binary
+                type: string
+          description: OK
+        "400":
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ErrorResponse"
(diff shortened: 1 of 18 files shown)


robelkin:
Addressed the Medium finding in 5bdb2fc: `content_hash` is now included in the message-detail attachment payload (`/api/v1/messages/{id}`), populated from both the query-engine and store-backed paths, so clients can discover the hash to pass to `GET /api/v1/attachments/{hash}`. Added a test asserting the field is present.

wesm:
looking

wesm:
I will rebase this once #464 is merged and then I can get this ready to merge

PR #465, opened 2026-07-09
Add Granola and Circleback meeting sync and TUI browsing

### LINKED ISSUE #456: Support for meeting transcripts and recordings
Storage in msgvault for services like Circleback and Granola

sweenzor:
Closes #456.

This PR archives Granola and Circleback meeting notes and transcripts in msgvault and adds a read-only Meetings mode to the TUI. Meetings use the existing message, search, daemon, and cache infrastructure, so the same archive works locally and through a remote daemon.

## What users get

- Configure one or more Granola or Circleback sources, validate access, and run full or incremental syncs.
- Press `m` in the TUI to switch to Meetings without losing the current Email or Texts position.
- Browse meetings by date, title, organizer, and source; filter to one provider with `A`.
- Search titles, people, notes, action items, insights, tags, and transcripts, then use `/`, `n`, and `N` to find text inside an open meeting.
- Read headings, lists, emphasis, code, and transcript line breaks as formatted Markdown; source content is restricted to safe terminal styling.

## Reliable syncing

- Granola syncs through its Business API, stores each meeting atomically, rejects inconsistent note identities, keeps cursors stable across partial or date-bounded runs, and makes progress across limited syncs when notes share an update timestamp.
- Circleback syncs through its MCP service with browser OAuth. Incremental discovery includes historical backfills, retries transcripts that are not ready yet, follows postponed meeting times, and never replaces an archived transcript with a notes-only response.
- Sync failures remain visible instead of being recorded as clean runs, and removed sources are not silently recreated by scheduled jobs.
- Every meeting source requires an explicit `account_email`; organizer identity never falls back to a source label.
- Expiring Circleback recording URLs remain provider metadata rather than durable attachments.

## Search and cache behavior

- Meeting transcripts and provider metadata are searchable through SQLite, PostgreSQL, DuckDB, the HTTP API, and daemon-backed clients.
- The Parquet cache retains meetings and chat messages for search while ordinary email analytics remain email-only by default.
- Email-mode lists, drill-downs, searches, threads, and deletion filters explicitly exclude meetings and text messages while retaining legacy email records.
- Cache updates distinguish append-only imports from changed messages, detect syncs racing an export, and rebuild after every failed-sync transition that may have persisted content.
- SQLite date filters compare timestamps as instants across time-zone offsets and use a matching expression index for bounded queries.

## Documentation

The Zensical docs cover provider setup, limited first syncs, incremental operation, search, the Meetings TUI, identity configuration, retry behavior, and source removal. See `docs/usage/meetings.md` for the end-to-end workflow.


wesm:
Let me know if you need help here. I use Granola so I would actually use this.

sweenzor:
> Let me know if you need help here. I use Granola so I would actually use this.

what do you think, break it into 2 PRs?

wesm:
Let me rebase this and take a few turns

sweenzor:
sgtm! it's really a very different kind of development work isn't it.

wesm:
I think this is good, I just need to test it out tomorrow and I can get it merged

wesm:
I added TUI for this and markdown rendering in the TUI, will merge soon

PR #470, opened 2026-07-11
Split MCP search tools and add daemon-backed semantic matches

endolith:
Splits MCP message search into dedicated metadata, keyword-body, and semantic tools, while retaining `search_messages` as a deprecated compatibility wrapper. It also makes scored semantic chunk excerpts available through the daemon-backed path used by `msgvault mcp`, not only the in-process server.

- `search_metadata` returns exact-counted metadata matches; `search_message_bodies` returns newest-first keyword hits with bounded excerpts; `semantic_search_messages` returns relevance-ranked vector or hybrid results with scored excerpts. The compatibility wrapper dispatches to the appropriate new tool contract.
- Adds opt-in, post-pagination match enrichment to the daemon search API and carries pagination, `min_score`, scores, and excerpts through the generated client without adding body work to summary-only CLI or TUI searches.
- Reconstructs excerpts from the same plaintext-first, MIME-aware HTML fallback and preprocessing corpus used by indexing. Raw-body locations are returned only for exact unique mappings, and within-message scoring uses the same validated vector generation throughout the request.
- Updates the OpenAPI schema and the README, changelog, search guide, vector-search guide, and Zensical MCP documentation.


endolith:
Response from Kilo bot:

> ## The two Mediums — my recommendation
> 
> Both are **pre-existing in the vector-chunk implementation**, not introduced by this PR's split, so per your bite-size-PR preference I'd handle them as separate follow-ups rather than expanding this one:
> 
> 1. **Daemon path lacks `matches`/`min_score`** (`searchMessageBodiesHybridViaSearcher`): the daemon `HybridSearcher` contract returns ranked message hits but no per-chunk scores, so it structurally *can't* populate `matches[]` or apply `min_score` without extending the cross-process contract (add scored chunks to `HybridSearchResult`, or expose remote chunk scoring). That's a genuine feature, own PR.
> 
> 2. **Chunk offset mapping** (`matches.go:97`): `char_offset`/`line` are computed against the raw body by only subtracting the subject prefix, while chunk bounds are in preprocessed-text space. If preprocessing does more than prepend the subject (CRLF normalize, trim, collapse whitespace), `center_at` targets drift. Fixing needs a preprocessed→raw offset map. Correctness bug, own PR — and it only affects the local hybrid path that *does* emit chunk matches.
>

endolith:
## roborev review (`8e633c4`) follow-up

I addressed the one new finding that was both valid and a quick, in-scope fix, and left the two pre-existing ones as deferred follow-ups consistent with the earlier maintainer decision.

**Medium — `search_in_message` advertises `mode=vector` but it is unavailable in production (`server.go`/`handlers.go`):** Valid. The production `msgvault mcp` path (`cmd/msgvault/cmd/mcp.go` `daemonMCPServeOptions`) wires only `opts.HybridSearcher` (and `SimilarSearcher`), never `opts.HybridEngine` or `opts.Backend`. `vectorMatchesInMessage` requires both `HybridEngine` and a `ChunkScoringBackend`, so in the daemon/CLI deployment `mode=vector` always returns `vector_not_enabled` while the tool still advertised it. Fixed in `c8bf9ee`: `search_in_message` now gates its `mode`/`min_score` advertisement on a new `vectorInMessageAvailable` capability (`opts.HybridEngine != nil && opts.Backend != nil`), mirroring how `semantic_search_messages` already gates its vector params on `vectorAvailable`. The handler's `vector_not_enabled` path stays as a safety net. Added `TestSearchInMessageTool_AdvertisesVectorModeWhenConfigured`.

**Medium — daemon `semantic_search_messages` ignores `min_score` / omits chunk `matches` (`handlers.go:763`):** Valid, but pre-existing in `searchMessageBodiesHybridViaSearcher` and not introduced by this tool-split PR. Resolving it needs the cross-process `HybridSearcher` contract to return scored chunk matches (or a remote chunk-scoring op). Continuing to defer as its own feature PR, per the earlier bite-size-PR decision.

**Medium — preprocessed chunk offsets applied to raw bodies (`matches.go:97`):** Valid, also pre-existing in `chunkHitsToMatches`; only affects the local hybrid path that emits chunk matches. Fixing it properly needs a preprocessed→raw offset map. Deferred to its own follow-up for the same reason.

Net after this review: nil-panic (prior commit `8e633c4`) accepted; new production-advertisement gap fixed in `c8bf9ee`; two pre-existing correctness gaps left as separate follow-ups.

CC @endolith

endolith:
## roborev review follow-up (High + Medium)

Thanks for the re-review. Here's where each item stands:

**High — `semantic_search_messages` omits `matches` / ignores `min_score` (daemon path):**
Valid, but it's a pre-existing cross-process feature gap, not a regression from this
tool-split PR. I traced both routes: `searchMessageBodiesHybrid` (in-process,
handlers.go:626) and `searchMessagesHybridViaSearcher` (daemon, handlers.go:735) both
build items with `MessageSummary` only — **neither populates `Matches`** — and
`HybridSearchRequest` (handlers.go:139) carries no `MinScore`. So `semantic_search_messages`
advertises body chunks in *both* paths today but returns none.

Fixing it properly is a feature, not a one-line fix: extend the daemon HTTP contract
(openapi.yaml + regenerated client) to accept `min_score` and return scored chunk
matches, update the `daemonclient` adapter and the MCP `HybridSearch*` types, and add
per-hit chunk scoring in both routes via `ScoreMessageChunks` / `chunkHitsToMatches`.
It's also entangled with the offset-mapping Medium (`matches.go:97`) — the returned
`char_offset` / `line` are only correct once that's fixed, so they should land together.

This is the same work already scoped as a dedicated follow-up (the earlier "two Mediums"
decision). Expanding this bite-size tool-split PR with a codegen-dependent, multi-package
change would conflict with that policy, so I'm leaving it for a separate PR that combines
daemon chunk matches + `min_score` + offset mapping. Happy to open that.

**Medium — preprocessed chunk offsets applied to raw bodies (`matches.go:97`):**
Valid, pre-existing in `chunkHitsToMatches`; only affects the local hybrid path that emits
chunk matches. Needs a preprocessed→raw offset map. Deferred to the same dedicated follow-up
as above (it must ship with the chunk-match feature so the offsets are correct).

**Already resolved in this PR:**
- Nil-panic in `vectorMatchesInMessage` (prior `(nil, nil)` not-found) → fixed in `8e633c4`.
- `search_in_message` advertised `mode=vector` while production daemon wiring only provides
  `HybridSearcher` (never `HybridEngine`/`Backend`) → fixed in `c8bf9ee` by gating the
  `mode`/`min_score` advertisement on actual in-process capability, mirroring
  `semantic_search_messages`'s existing `vectorAvailable` gating.

Net: the two genuinely quick, in-scope correctness bugs are fixed; the remaining High/Medium
are a pre-existing feature (daemon chunk matches + offset mapping) deferred to its own PR.

wesm:
I can take the wheel from here

PR #482, opened 2026-07-15
Improve daemon startup UX: quiet log echoes and self-repairing analytics cache

wesm:
Fixes two daemon-startup UX problems observed when a CLI command auto-starts the daemon.

## Startup progress dumped raw SQL

The startup progress reporter echoed raw `serve.log` lines whenever a record carried unrecognized logfmt keys. A `sql slow` record embeds the full statement, so a schema migration flooded the terminal with kilobytes of DDL. `sql slow` records now render as `running a slow SQL statement (424ms)`, and every echoed line is capped at 240 characters — the full line remains in `serve.log`, whose path the progress output already prints.

## "Run build-cache, then restart the daemon"

With the default `[analytics] engine = "auto"`, a daemon that started without a usable Parquet cache stayed on live-SQL fallback for its entire lifetime, and the TUI told the user to run `msgvault build-cache` and restart the daemon. Now:

- With `auto_build_cache = true` (the default), the daemon builds the cache in the background after startup — startup is never blocked — and switches its analytics engine onto DuckDB as soon as the build lands.
- Any daemon-run cache build (`msgvault build-cache`, post-sync refreshes, SQL-query rebuilds) is adopted the same way, so a restart is never needed to leave the fallback.
- `/health` gains `analytics_cache_building`, and the TUI launch notice reports the in-progress build instead of prescribing manual steps.
- `engine = "sql"` (deliberate live SQL) and `engine = "duckdb"` (blocking startup build) behave as before.

The engine swap is guarded by a new RWMutex in the API server; handlers read the engine through a getter, mirroring the existing background-installed vector subsystem.


wesm:
Closing in favor of an alternate approach

PR #489, opened 2026-07-19
feat: Slack workspace archiving (user-token Web API sync)

m-j-r:
Archive your own view of a Slack workspace — public/private channels you're a member of, group DMs, and 1:1 DMs — via the Web API, searchable alongside mail and other chat sources through the existing TUI / FTS / Parquet analytics.

## Usage

```bash
msgvault add-slack     # user token (xoxp-…) from an internal app you create — 2-minute setup, scope list in docs/usage/slack.md
msgvault sync-slack    # first run backfills full history; later runs are incremental
```

Each workspace is its own source (`slack`, identified by `<team-id>:<user-id>`); multiple workspaces coexist and stay separately filterable in the TUI. A `[slack]` config block covers daemon scheduling, channel include/exclude filters, a media size cap, and the thread-tracking window.

## What's stored

Message text with FTS (mrkdwn rendered, mentions resolved to names), threads (root+reply via `reply_to_message_id`), reactions, @mention recipient rows, edits, conversation membership, files in the content-addressed attachment store, and the verbatim API JSON per message (`raw_format = slack_json`). Bot/integration messages that carry their content in legacy `attachments` or Block Kit `blocks` get that text extracted into the searchable body. Members' profile emails unify their Slack identity with their archived mail (Teams AAD-resolution precedent).

## Behavior notes

- Reuses the existing chat schema — no new core tables.
- The client is read-only by construction (a strict method allowlist rejects anything outside the read set) and internal-app rate limits apply (999-message pages; the 2025 non-Marketplace clampdown targets distributed apps only).
- Slack never surfaces thread replies in oldest-filtered channel history, so incremental sync tracks thread roots and polls `conversations.replies` per root for a configurable lookback (default 30 days); older late replies and edits beyond the 7-day rescan window are caught by `--full`.
- Backfills checkpoint per conversation and mid-conversation; cursors only advance after a window fully persists, so interrupted or partially-failed runs resume without loss or duplicates.
- File bytes are fetched only from `files.slack.com` (no redirects followed); external/off-host files are recorded as metadata + permalink. Failed downloads retry via `backfill-slack-media`.
- Conversations that enumerate but 404 on read are skipped, not fatal (observed live: a sandbox Slackbot DM).
- Link unfurls on user messages are not indexed (page previews are content the person never wrote).
- Archived messages survive source deletion (archive semantics).

## Store/query changes

- Slack attachment-row methods share prefix-parameterized helpers with the Beeper ones (`listPendingAttachmentMessages`, `messageAttachmentsByPrefix`); `BeeperPendingAttachmentMessage` is now an alias of the shared `PendingAttachmentMessage`.
- `"slack"` registered in `KnownMessageTypes` / `TextMessageTypes`.

Validated against live workspaces: full backfill (1,100+ message channel), interrupt/resume, incremental re-runs, threads incl. cross-run late replies, reactions (incl. custom emoji), media download into CAS packs, bot payloads, a real 429 absorbed via Retry-After (env-gated live test included), and history depth past 8 years on a paid plan.

Planned follow-up (separate PR): opt-in archiving of unjoined public channels via an explicit config allowlist — reads verified live; kept out of this PR so the consent/config design can be discussed on its own.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,6 +77,11 @@ make lint                     # Run linter
 ./msgvault sync-teams you@tenant.com         # Sync Teams chats + channels
 ./msgvault sync-teams you@tenant.com --no-channels --limit 50
 
+# Slack (user token from your own internal app)
+./msgvault add-slack                         # Register a workspace (prompts for xoxp token)
+./msgvault sync-slack                        # Sync channels, group DMs, DMs
+./msgvault sync-slack --full                 # Repair: re-fetch + upsert in place
+
 # Daemon mode (NAS/server deployment)
 ./msgvault serve                                      # Start HTTP API + scheduled syncs
 
--- a/cmd/msgvault/cmd/add_slack.go
+++ b/cmd/msgvault/cmd/add_slack.go
@@ -1,0 +1,150 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/mattn/go-isatty"
+	"github.com/spf13/cobra"
+	"go.kenn.io/msgvault/internal/clirun"
+	"go.kenn.io/msgvault/internal/slack"
+)
+
+var (
+	addSlackTokenFile         string
+	noDefaultIdentityAddSlack bool
+)
+
+func newAddSlackCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-slack",
+		Short: "Add a Slack workspace as an archive source",
+		Long: `Add a Slack workspace as an archive source.
+
+Archives your own view of the workspace — public/private channels you are a
+member of, group DMs, and 1:1 DMs — via the Slack Web API.
+
+Requires a user token (xoxp-...) from an internal Slack app you create:
+
+  1. https://api.slack.com/apps > Create New App > From scratch, in your
+     workspace.
+  2. OAuth & Permissions > User Token Scopes, add:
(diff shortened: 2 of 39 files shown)


m-j-r:
All three confirmed and fixed in [2d129c5](https://github.com/kenn-io/msgvault/pull/489/commits/2d129c56fa90a12491b5468f8fbe798fa404fbe6):
- roots are now pruned only after being successfully polled (and never under --no-threads)
- the edit rescan uses an inclusive upper bound so the cursor message itself is re-checked
- history/replies page sizes are capped to the remaining --limit budget.

Each has a regression test verified to fail against the previous behavior.

wesm:
awesome! I was hoping someone would contribute this. I will look in more detail soon

m-j-r:
Just an update - I'm reviewing the design, based on the consistent scanning window issues, will update soon.

wesm:
no problem

m-j-r:
Closing in favor of a successor PR with a revised reply-capture architecture. Since this opened, live probing established that Slack's search API reliably returns thread replies by creation time (threads:replies modifier, day-granular bounds, stable ascending pagination) — which enables replacing the thread-lookback polling reviewed here with a search-based sweep: simpler state (one UTC watermark replaces per-root cursor maps and pruning), and it removes the 30-day late-reply blind spot this design documented as a limitation. Everything else (client, mapping, media, identity, CLI, tests) carries forward, including the fixes from all three review rounds here. Successor PR follows shortly with the full probe ledger in its design doc.

PR #493, opened 2026-07-20
feat(o365): allow users to configure redirect_uri alongside client_id

mattf:
The OAuth redirect URI is tied to the client ID at the provider — the callback URI registered in the app must match what msgvault sends. This adds a redirect_uri config field so users can specify their own and use it with any client ID, without relying on a hardcoded redirect port/path.

Also supports localhost URIs on privileged ports (e.g. Thunderbird's https://localhost/ on port 443). https redirects use a self-signed certificate so the OAuth flow works end-to-end for local testing.


mattf:
@wesm thank you for getting to those issues for me

PR #494, opened 2026-07-20
feat: Slack workspace archiving (user-token sync + search-based reply sweep)

m-j-r:
Archive your own view of a Slack workspace — public/private channels you're a member of, group DMs, and 1:1 DMs — via the Web API, searchable alongside mail and other chat sources through the existing TUI / FTS / Parquet analytics.

Supersedes #489. Since that PR opened, live probing established that Slack's search API reliably returns thread replies by creation time (`threads:replies`, day-granular bounds, stable ascending pagination) — enabling a simpler and more correct reply-capture architecture than the lookback-window thread polling reviewed there. Design docs: `docs/internal/slack-ingestion-design.md` (base) and `docs/internal/slack-reply-sweep-design.md` (reply sweep, with the full probe ledger).

## Usage

```bash
msgvault add-slack     # user token (xoxp-…) from an internal app you create — 2-minute setup, scope list in docs/usage/slack.md
msgvault sync-slack    # first run backfills full history; later runs are incremental
```

Each workspace membership is its own source (`slack`, `<team-id>:<user-id>`); multiple workspaces coexist and stay separately filterable in the TUI. A `[slack]` config block covers daemon scheduling, channel include/exclude filters, and a media size cap.

## How replies are captured

- **Backfill** walks each conversation's history and fetches every thread's replies inline, before the containing page's cursor advances — interrupt/resume-safe by construction.
- **Incremental** fetches new top-level messages by cursor, then runs a **reply sweep**: a `search.messages threads:replies` query over the days since a per-source UTC watermark discovers replies by *creation time* — a reply to a thread of any age is found (no tracking window, no blind spot). Hits are pointers only; archiving always goes through canonical `conversations.replies` fetches into the standard upsert path.
- Sweep mechanics follow live-probed behaviors documented in the design doc: date modifiers evaluate in the user's *current profile timezone* (offset read fresh each run; watermark stored as a UTC instant so tz changes and DST are gap-free); every query string carries an inert nonce (search results are cached by query string); the pager stops on a mismatched echoed page number (past-the-cap pages are silently clamped to page 1) and logs a tripwire if a single day exceeds the reachable ceiling.
- Edits and reaction changes after capture are **ignored by default** (archive semantics); `sync-slack --maintenance` repairs the recent window, `--full` repairs everything. A backfill run under `--no-threads` leaves a recorded debt that the next threaded run pays automatically.

## What's stored

Message text with FTS (mrkdwn rendered, mentions resolved to names, bot `attachments`/`blocks` payloads extracted), threads (root+reply via `reply_to_message_id`), reactions, @mention recipient rows, conversation membership, files in the content-addressed attachment store, and the verbatim API JSON per message (`raw_format = slack_json`). Members' profile emails unify their Slack identity with their archived mail. Reuses the existing chat schema — no new core tables.

## Safety and correctness properties

- The client is read-only by construction (strict method allowlist) and internal-app rate limits apply (999-message pages; the 2025 non-Marketplace clampdown targets distributed apps only — probed live).
- File bytes are fetched only from `files.slack.com` (no redirects); external/off-host files are metadata + permalink; failed/deferred downloads (incl. `--no-media`) stay discoverable by `backfill-slack-media`.
- Every cursor — backfill page chain, incremental window, sweep watermark — advances only behind persisted work; checkpoints merge atomically per cursor unit; all writes are upserts, so at-least-once fetching yields exactly-once storage.
- Conversations that enumerate but 404 on read are skipped, not fatal (observed live).

## Store/query changes

Slack attachment-row methods share prefix-parameterized helpers with the Beeper ones; `"slack"` registered in `KnownMessageTypes` / `TextMessageTypes`.

Validated against live workspaces: full backfill (1,100+ message channel), interrupt/resume, threads incl. cross-run late replies, reactions (incl. custom emoji), media into CAS packs, a real 429 absorbed via Retry-After, history depth past 8 years, and the sweep's search behaviors (replies-only modifier, unjoined-channel leakage → filtered, tz re-filing, query caching, page clamping — each with an env-gated live pin-test or fake-server regression test).

Planned follow-ups (separate PRs): opt-in archiving of unjoined public channels via an explicit config allowlist; concurrent fetch/persist pipeline.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -77,6 +77,11 @@ make lint                     # Run linter
 ./msgvault sync-teams you@tenant.com         # Sync Teams chats + channels
 ./msgvault sync-teams you@tenant.com --no-channels --limit 50
 
+# Slack (user token from your own internal app)
+./msgvault add-slack                         # Register a workspace (prompts for xoxp token)
+./msgvault sync-slack                        # Sync channels, group DMs, DMs
+./msgvault sync-slack --full                 # Repair: re-fetch + upsert in place
+
 # Daemon mode (NAS/server deployment)
 ./msgvault serve                                      # Start HTTP API + scheduled syncs
 
--- a/cmd/msgvault/cmd/add_slack.go
+++ b/cmd/msgvault/cmd/add_slack.go
@@ -1,0 +1,157 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/mattn/go-isatty"
+	"github.com/spf13/cobra"
+	"go.kenn.io/msgvault/internal/clirun"
+	"go.kenn.io/msgvault/internal/slack"
+)
+
+var (
+	addSlackTokenFile         string
+	noDefaultIdentityAddSlack bool
+)
+
+func newAddSlackCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add-slack",
+		Short: "Add a Slack workspace as an archive source",
+		Long: `Add a Slack workspace as an archive source.
+
+Archives your own view of the workspace — public/private channels you are a
+member of, group DMs, and 1:1 DMs — via the Slack Web API.
+
+Requires a user token (xoxp-...) from an internal Slack app you create:
+
+  1. https://api.slack.com/apps > Create New App > From scratch, in your
+     workspace.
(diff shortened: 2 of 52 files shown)


wesm:
looking

wesm:
This appears to be good now, running CI checks and will merge

wesm:
thank you! Will release soon. Let me know if you encounter any problems syncing from main

PR #502, opened 2026-07-24
feat(imap): add folder listing and filtering for imap sync

### LINKED ISSUE #196: IMAP: add --folders flag for selective folder sync
## Context

The IMAP sync added in #193 always syncs all selectable mailboxes. For large accounts with many folders, users may want to sync only specific folders.

## Proposal

Add a `--folders` flag to `sync-full` (comma-separated list of IMAP folder names). Only the specified folders would be enumerated and fetched. Ignored for Gmail sources.

```bash
msgvault sync-full imaps://user@host:993 --folders "INBOX,Sent"
```

## References

- Yahoo Mail PRD: #93
- IMAP support: #193

mattf:
## What changed

- Adds `msgvault list-folders [account]` to show selectable IMAP folders and approximate message counts. Without an account name, it lists folders for every configured IMAP account.
- Adds repeatable `--folders` and `--skip-folders` options to both `sync-full` and `sync`. Each flag takes one complete folder name, so names containing commas work correctly.
- Supports the same folder filters when the CLI talks to a local daemon or a configured remote server.
- Keeps filtered scans narrow, including on servers with Gmail-style All Mail, Trash, and Junk folders.
- Preserves stable message identities during filtered or incomplete rescans. These scans add newly observed folder labels without removing labels learned earlier. An unfiltered scan replaces stale labels only when every mailbox was collected successfully.
- Refreshes labels for messages that are already archived by fetching only lightweight Message-ID headers. Raw message bodies and attachments are downloaded only for new messages.
- Handles metadata failures per message, so a UID that disappears after listing does not prevent other messages in the batch from updating.
- Adds a Zensical guide, setup examples, and complete CLI reference entries for folder discovery, matching rules, and safe filtered syncs.

## Why

A first sync can take a long time on a large IMAP account. Folder filtering lets someone start with a useful subset, confirm that msgvault fits their needs, and expand the archive later. It also lets ongoing syncs ignore folders that are not useful to the local archive.

## How to use it

First, find the folder names reported by the server:

```bash
msgvault list-folders you@example.com
```

Sync only selected folders by repeating `--folders`:

```bash
msgvault sync-full you@example.com \
  --folders INBOX \
  --folders Archive
```

Or scan everything except selected folders by repeating `--skip-folders`:

```bash
msgvault sync you@example.com \
  --skip-folders Trash \
  --skip-folders Spam
```

Folder names use exact, case-insensitive matching. When include and exclude flags are combined, msgvault applies the include list first and then removes excluded folders. The flags affect IMAP accounts only and apply to the current command. With no folder flags, msgvault scans every selectable folder.

Filtering changes which remote folders are scanned; it does not delete messages from the provider or remove messages already stored in the local archive.

Closes #196.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -6351,6 +6351,20 @@ paths:
           name: email
           schema:
             type: string
+        - description: IMAP folder names to include (repeatable)
+          in: query
+          name: folder
+          schema:
+            items:
+              type: string
+            type: array
+        - description: IMAP folder names to exclude (repeatable)
+          in: query
+          name: skip-folder
+          schema:
+            items:
+              type: string
+            type: array
       responses:
         "200":
           content:
@@ -6404,6 +6418,20 @@ paths:
           name: noresume
           schema:
             type: boolean
+        - description: IMAP folder names to include (repeatable)
+          in: query
+          name: folder
+          schema:
+            items:
+              type: string
+            type: array
+        - description: IMAP folder names to exclude (repeatable)
+          in: query
+          name: skip-folder
+          schema:
+            items:
+              type: string
+            type: array
       responses:
         "200":
           content:
--- a/cmd/msgvault/cmd/imap_folder_state_sync_test.go
+++ b/cmd/msgvault/cmd/imap_folder_state_sync_test.go
@@ -77,6 +77,26 @@ func TestSaveIMAPFolderStates_ErrorsBlockPersistence(t *testing.T) {
 	assert.Empty(t, loaded, "a run with fetch errors must not advance folder high water marks")
 }
 
+func TestSaveIMAPFolderStates_LeavesDeferredFoldersUnpersisted(t *testing.T) {
+	require := require.New(t)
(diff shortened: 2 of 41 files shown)


mattf:
please excuse the force-pushing. i messed up the history and decided to just squash everything.

i'm not using gmail (or postgres, oops), so how would you like me to resolve the \All feedback?

mariusvniekerk:
@mattf force pushing is very normal in this agent era. rebases are now basically free. huzzah!

wesm:
taking a look

mattf:
i went with the repeated arg approach after trying an escaping approach w/ `\,`. i think the escaping was going to be too complicated for users as they'd contend with both shell & msgvault escaping.

granted, a `,` in a folder name is likely a rare event.

wesm:
I can take over from here

mattf:
> I can take over from here

:+1:

wesm:
Will merge pending CI. If I broke something in all my grinding please open another PR!

mattf:
@wesm that was a pretty epic back and forth

PR #504, opened 2026-07-25
Repair missing and implausible message dates

### LINKED ISSUE #501: Handle implausible MIME dates and repair existing message timestamps
## What I was doing

I imported an older mail archive and found messages appearing decades away from where they belonged in the timeline. The raw `Date:` headers were syntactically valid, so msgvault accepted them as `sent_at`, but the timestamps were clearly implausible, including Unix-epoch defaults and dates far in the future.

In these cases, the raw MIME often contains a plausible timestamp in the `Received:` chain:

```text
Date: Thu, 1 Jan 1970 00:00:00 +0000
Received: from sender.example.net by mx.example.net;
        Tue, 2 Jan 2007 15:04:05 +0000
```

## What happened

msgvault distinguishes parseable dates from unparseable dates, but it does not distinguish plausible dates from impossible ones. Once a parseable value reaches `messages.sent_at`, it takes precedence over `received_at` and `internal_date` in timeline ordering.

That leaves two gaps:

1. New imports can store an implausible `Date:` value even when the MIME trace or source metadata contains a better timestamp.
2. Existing archives have no supported way to find and repair rows that already contain these values.

The existing MIME timezone fixes solve parsing inconsistencies. They do not address parseable-but-implausible timestamps.

## Why it matters

A bad `sent_at` value affects more than display. It changes message ordering, date-range searches, statistics, and year-based analytics cache partitions. Reimporting the source archive is also not a sufficient repair path because the same resolver would accept the same header again.

## Proposed behavior

Use one date resolver across every message ingestion path:

1. Accept the parsed `Date:` header when it falls within a defined plausibility window.
2. Otherwise, use the oldest plausible timestamp from the `Received:` chain. Since trace fields are prepended, this is normally the last plausible `Received:` timestamp in source order.
3. Otherwise, use the source's internal timestamp when available and plausible.
4. Otherwise, leave the canonical send time unset rather than storing a known-bad value.

A prototype used a window of `1990-01-01` through `now + 30 days`. The exact bounds are worth discussing, but they should be defined once and shared by ingestion and repair.

Existing archives also need a maintenance command such as:

```bash
msgvault repair-dates          # report proposed changes
msgvault repair-dates --apply  # write them
```

The command should:

- default to a read-only scan;
- inspect only rows with missing or implausible `sent_at` values;
- resolve replacements from the stored raw MIME and source metadata;
- leave already-plausible rows untouched;
- record before-and-after values in an audit ledger;
- rebuild the analytics cache after applying changes;
- never modify original source files or remote servers.

## Acceptance criteria

- All email ingestion paths use the same plausibility and fallback rules.
- A plausible `Date:` header remains authoritative.
- An implausible `Date:` header falls back to the oldest plausible `Received:` timestamp.
- Source internal time is used only when the header and trace chain cannot provide a plausible value.
- The repair command is read-only unless explicitly applied.
- Applying repairs is idempotent.
- Original source data is never modified.

I have a prototype implementation and tests for this behavior that I can clean up and submit after agreeing on the contract.

jesserobbins:
## Motivation

I imported an older mail archive and found messages appearing decades away
from where they belonged because syntactically valid but implausible `Date:`
headers were treated as authoritative. Those timestamps affect timeline
ordering, date-range searches, statistics, and analytics partitions, and
reimporting would accept the same bad values again.

## Summary

- add one plausibility resolver for email ingestion, bounded from 1990 through
  30 days in the future, using a valid `Date:` header, then the oldest
  plausible `Received:` timestamp, then plausible source metadata
- add `repair-dates`, with a read-only default and explicit `--apply` mode
- apply repairs atomically with changed-row protection and a privacy-safe JSON
  audit ledger
- rebuild the Parquet analytics cache for SQLite archives while correctly
  skipping that SQLite-only cache on PostgreSQL
- route the command through the daemon and document recovery for partial
  failures
- warn when ingestion rejects a nonzero implausible header so genuinely old
  archives do not change silently
- fix a `mime.Parse` bug where a mangled Date header (mbox continuation
  artifact, or US month-day ordering) silently dropped the sent date instead
  of resolving it, recovering 62 of the messages this PR's own audit found
  unresolvable
- report *why* each remaining `repair-dates` candidate is unresolved
  (date header present but unusable, Received headers present but unusable,
  or no date signal at all) instead of a bare count

## Note

This branch currently includes the `mime.Parse` fix (commit `6bf31dc9`) as
its own commit, duplicated from #506, which splits that fix out for
isolated review since it's a general ingestion bug, not specific to
`repair-dates`. Kept here too so this branch builds and tests standalone;
will rebase to drop the duplicate once #506 merges.

## Test plan

- [x] `go test -tags "fts5 sqlite_vec" ./cmd/... ./internal/... ./pkg/... ./tools/...`
- [x] `go vet ./...`
- [x] `make lint-ci`
- [x] verify dry-run, apply, idempotency, audit-ledger failure states, daemon
  admission, SQLite cache rebuilding, and PostgreSQL cache selection with
  synthetic fixtures

Closes #501

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,6 +82,8 @@ make lint                     # Run linter
 
 # Maintenance
 ./msgvault repair-encoding                            # Fix UTF-8 encoding issues
+./msgvault repair-dates                              # Preview missing/implausible date repairs
+./msgvault repair-dates --apply                      # Apply repairs (and rebuild SQLite analytics cache)
 
 # Agent skills
 ./msgvault skills install                             # Install agent skills (Claude Code, Codex)
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ available with `msgvault tui`.
 | `update` | Update msgvault to the latest version |
 | `setup` | Interactive first-run configuration wizard |
 | `repair-encoding` | Fix UTF-8 encoding issues |
+| `repair-dates` | Report or repair missing and implausible email sent dates |
 | `list-senders` / `list-domains` / `list-labels` | Explore metadata |
 
 See the [CLI Reference](https://msgvault.io/cli-reference/) for full details.
@@ -149,7 +150,7 @@ msgvault can search your archive semantically using vector embeddings in additio
 
 A separate MCP tool, `find_similar_messages`, returns nearest neighbors for a seed message. See the [Vector Search guide](https://msgvault.io/usage/vector-search/) for setup, backfill, and troubleshooting.
 
-> **Archive writes are daemon-owned.** CLI writer commands such as `msgvault sync-full`, `msgvault embeddings build`, `msgvault repair-encoding`, and `msgvault rebuild-fts` send their work to the configured remote server or local background daemon. The daemon serializes archive mutations and streams progress back to your terminal, so normal CLI ergonomics stay the same without opening a second SQLite writer process.
+> **Archive writes are daemon-owned.** CLI writer commands such as `msgvault sync-full`, `msgvault embeddings build`, `msgvault repair-dates --apply`, and `msgvault rebuild-fts` send their work to the configured remote server or local background daemon. The daemon serializes archive mutations and streams progress back to your terminal, so normal CLI ergonomics stay the same without opening a second SQLite writer process.
 
 Large archives can scope an embedding generation with `[vector.embed.scope] message_types = ["sms", "mms"]`. Scoped vector and hybrid searches must include a matching `message_type` filter so a partial index is never used as if it covered the whole archive.
 
--- a/cmd/msgvault/cmd/repair_dates.go
+++ b/cmd/msgvault/cmd/repair_dates.go
@@ -1,0 +1,518 @@
+package cmd
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"errors"
+	"fmt"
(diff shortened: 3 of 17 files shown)


mariusvniekerk (approved):

jesserobbins:
@mariusvniekerk I think at some point we may want to migrate these various mime/date/utf8 fixes into their own subcommand. Tested and worked against my message archive.

mariusvniekerk:
makes sense, those are all email-repair related things makes sense to keen them together and not pollute the top level subcommand space that much

jesserobbins:
@mariusvniekerk two commits landed after your approval, flagging in case you want another look:

- \`6bf31dc9\` — fixes a \`mime.Parse\` bug (mbox continuation artifact + US month-day date ordering both silently dropped the sent date). Split out as its own PR at #506 for isolated review; kept here too so this branch builds standalone. Will rebase to drop the duplicate once #506 merges.
- \`9018eabc\` — \`repair-dates\` now reports *why* each remaining unresolved candidate is stuck (date header present but unusable / Received headers present but unusable / no date signal) instead of a bare count.

CI has already rerun clean against the new head.

wesm:
rebasing

wesm:
LGTM, thank you!

PR #506, opened 2026-07-25
fix: recover Date headers mangled by mbox continuation artifacts

### LINKED ISSUE #505: mime.Parse silently drops Date header on mbox continuation artifact and US month-day format
## What I was doing

Auditing why archived messages were missing a sent date, tracing a batch of them back to `internal/mime.Parse`.

## What happened

Two Date header patterns that occur in real mailboxes are silently dropped by `parseDate` in `internal/mime/parse.go`, so `Message.Date` ends up as the zero value even though the header is fine:

1. **Stray mbox continuation line.** Some mbox/Thunderbird-derived sources have a lone, unindented `.` line directly after the `Date:` header (a legacy artifact of that format). enmime treats it as a continuation of the previous header rather than erroring, so the header value becomes e.g. `"Thu, 27 May 2004 17:06:42 -0500 ."` — note the trailing ` .`. None of the layouts in `dateFormats` tolerate that suffix, so `time.ParseInLocation` fails for every candidate format and `parseDate` returns the zero time.

2. **Weekday + US month-day ordering.** `"Tue, Oct 17 2000 02:15:24 -0700"` (weekday, comma, month name, day, year — no comma after the day) isn't covered by any layout in `dateFormats`.

Minimal repro for (1), reflecting a real message:

```
X-Mozilla-Status: 7001
X-Mozilla-Status2: 00000000
Date: Thu, 27 May 2004 17:06:42 -0500
.
Content-Type: text/plain; charset=ISO-8859-1

Body text.
```

`mime.Parse` on this returns a `Message` with a zero `Date`, even though the header is perfectly readable.

## Why it matters

This isn't just a historical-data problem — it happens at ingestion time for any message with one of these patterns, so the sent date is lost the moment the message is archived, not just when re-scanning old data. I found it while working on a date-repair pass and confirmed against a real archive that both patterns account for a meaningful share (~43%) of otherwise-unresolvable messages there.

## Fix

Isolated fix and regression tests in #506. The same commit is also currently included directly in #504 (the broader repair-dates PR) so that branch builds standalone; it'll be rebased to drop the duplicate once #506 merges.

jesserobbins:
## Motivation

I ran into this while tracing why some archived messages had no sent date. It's not a historical-data problem — it happens at ingestion time, so the sent date is lost the moment a message with either pattern below is parsed, not just when re-scanning old data. I am submitting this as separate PR from #501 #504 because this is a clear bug with the importer flow. 

Fixes #505

## Summary

- `parseDate` now trims a trailing `" ."` artifact before trying any layout. Some mbox/Thunderbird-derived sources have a lone, unindented `.` line directly after the `Date:` header; enmime folds it onto the header value as a continuation instead of erroring, leaving a trailing `" ."` that no layout tolerated.
- Added a `dateFormats` layout for weekday + US month-day ordering (`"Tue, Oct 17 2000 02:15:24 -0700"`), which wasn't covered by any existing layout.

## Test plan

- [x] `go test ./internal/mime/...`
- [x] New `TestParse_MalformedContinuationLineAfterDate` reproduces the real mbox artifact end-to-end through `Parse`
- [x] New `TestParseDate` cases for both patterns
- [x] `go vet ./...`, `gofmt -l` clean

---
*Authored by [Jesse Robbins](https://jesserobbins.com) (@jesserobbins)*

### DIFF
--- a/internal/mime/parse.go
+++ b/internal/mime/parse.go
@@ -220,6 +220,7 @@ var dateFormats = []string{
 	"2006-01-02T15:04:05-07:00",             // ISO 8601 with offset
 	"2006-01-02 15:04:05 -0700",             // SQL-like format
 	"2006-01-02 15:04:05",                   // SQL-like without TZ
+	"Mon, Jan 2 2006 15:04:05 -0700",        // Weekday, US month-day order, no comma after day
 }
 
 // numericOffsetRe matches numeric timezone offsets like +0000, -0700, +00:00, -07:00.
@@ -256,6 +257,12 @@ func toUTC(t time.Time, numericOffset bool) time.Time {
 func parseDate(s string) time.Time {
 	// Normalize whitespace efficiently: split on whitespace runs and rejoin
 	s = strings.Join(strings.Fields(s), " ")
+
+	// Some mbox-derived sources have a stray, unindented continuation line
+	// (a lone ".") directly after the Date header. enmime folds it onto the
+	// header value instead of treating it as a parse error, leaving a
+	// trailing " ." that no real Date header would ever contain.
+	s = strings.TrimSuffix(s, " .")
 
 	// Strip trailing timezone name in parentheses like "(UTC)" or "(PST)"
 	// but keep the numeric offset for parsing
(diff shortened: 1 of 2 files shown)


jesserobbins:
Also referenced from #504 (the broader repair-dates PR), which currently includes this same commit directly so that branch builds and tests standalone. Once this merges, #504 will be rebased to drop the duplicate.

PR #508, opened 2026-07-25
Route trusted Linux builds to managed runners

wesm:
## Summary

- route canonical same-repository Linux pull requests and main-branch builds to managed Linux runners, with forks and noncanonical copies falling back to GitHub-hosted runners
- invoke reusable workflows from `main` and let checkout use the caller event's immutable SHA without accepting a dispatcher-supplied ref
- lint proposed workflow revisions on GitHub-hosted infrastructure

The managed runner group admits only the listed reusable workflow files from `main`; PR-controlled dispatcher changes therefore cannot target it directly.

Merge #511 first so `main` accepts reusable-workflow calls before this PR replaces the existing pull-request trigger.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -1,0 +1,18 @@
+name: CI
+
+# Fork PRs use GitHub-hosted runners. Same-repository PRs call the
+# main-branch workflow, whose Linux jobs can use the managed public fleet.
+on:
+  pull_request:
+  pull_request_target:
+
+permissions: read-all
+
+jobs:
+  run:
+    if: >-
+      (github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository) ||
+      (github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name == github.repository)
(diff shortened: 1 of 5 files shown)


wesm:
false positive

wesm:
This finding is a false positive under the runner policy currently enforced by GitHub.

The runner group that provides `kenn-linux-x64-public` has selected-repository visibility and `restricted_to_workflows: true`. For this repository, the only admitted workflow identities are:

- `kenn-io/msgvault/.github/workflows/ci.yml@refs/heads/main`
- `kenn-io/msgvault/.github/workflows/docker.yml@refs/heads/main`

A fork can rewrite `ci-pr.yml` or `docker-pr.yml` and request the label, but GitHub will not assign a runner because that PR-controlled caller workflow identity is not in the group allowlist. The admitted main-branch reusable workflows also derive the head and base repository identities themselves before selecting the managed label; the dispatcher does not supply the trust decision.

This would become a real vulnerability if the group stopped being restricted to selected workflows, if a PR-ref dispatcher were added to the allowlist, or if the same label were exposed through another unrestricted runner group. None of those conditions is present.

PR #512, opened 2026-07-26
Fix deduplicate plan timeout: index + batch RFC822 duplicate-group lookup

### LINKED ISSUE #510: deduplicate --content-hash plan times out — N+1 query on missing rfc822_message_id index
## What I was doing
Running `msgvault deduplicate --collection local --content-hash --no-backup -v` on a local archive.

## What happened
After ~30 minutes the CLI errored with `context deadline exceeded` calling `/api/v1/cli/deduplicate/plan`. The server-side scan doesn't stream progress, so the client hits its hardcoded 30-minute timeout (`internal/api/server.go:287`, not configurable via flag or config.toml) with zero output in the meantime.

`serve.log` showed `GetDuplicateGroupMessages` (`internal/store/dedup.go:110`) executing 22,025 near-identical queries at ~190ms each — one per RFC822 duplicate group — because `messages.rfc822_message_id` has no index in either `schema.sql` or `schema_pg.sql`. That N+1 alone can burn the full 30-minute window before the content-hash comparison pass even starts.

Two related things I noticed along the way:
- `/plan` calls `BackfillRFC822IDs` and writes to the DB unconditionally (154k rows in my run) even without `-y`/confirmation.
- The server has no request deadline on this path (`isLongDaemonRequest` in `server.go`), so it keeps scanning after the client gives up — a killed CLI run can leave a scan running in the background.

## Why it matters
Anyone with a non-trivial message count can't run content-hash dedup through the normal `deduplicate` path at all right now — it always times out with no feedback, which makes it look hung rather than just slow. A workaround exists (adding `-y` or `--dry-run` reroutes through `/api/v1/cli/run`, which streams and has no client-side timeout), but the underlying fix is one of:
1. Index `messages.rfc822_message_id`.
2. Batch `GetDuplicateGroupMessages` into a single `IN (...)` query instead of one query per group.
3. Make `/cli/deduplicate/plan` stream progress like the other CLI endpoints do.

jesserobbins:
## Motivation

Running `deduplicate --content-hash` on my own archive, the CLI died after 30 minutes with `context deadline exceeded` and zero output. Traced it to an N+1 query with no supporting index — `Scan` was querying the store once per RFC822 duplicate group, and on a real archive that's tens of thousands of round trips before content-hash comparison even starts.

## Summary

- Add a plain index on `messages.rfc822_message_id` — a partial index (mirroring the existing attachment-hash indexes) turns out to be unusable by SQLite's planner for this query shape, verified via `EXPLAIN QUERY PLAN` before picking the plain form.
- Add `Store.GetDuplicateGroupMessagesBatch`, replacing the per-group query with a handful of chunked `IN (...)` queries.
- Switch `Engine.Scan` to the batched method.

Fixes #510.

## Test plan

- [x] New tests: index-usage assertion, batch/per-group equivalence across multiple chunk rounds, source-scoped filtering.
- [x] `make test` passes (one unrelated pre-existing failure in `internal/config` — a macOS `/var` vs `/private/var` symlink artifact in a file this PR never touches).
- [x] Real archive: `deduplicate --collection local --content-hash --no-backup -v --dry-run` against a personal archive with 161,113 duplicate groups. Before this fix, the equivalent run never completed — it hit the CLI's 30-minute client timeout with zero output, having made partial progress through roughly 22,025 of those groups. After: completes in ~311s.

Out of scope, tracked in #510 but not touched here: `BackfillRFC822IDs` writing to the DB unconditionally during `/plan` before confirmation, and the server not respecting client cancellation once a request times out client-side.

---
*Authored by [Jesse Robbins](https://jesserobbins.com) (@jesserobbins)*

### DIFF
--- a/internal/dedup/dedup.go
+++ b/internal/dedup/dedup.go
@@ -328,8 +328,8 @@ func (e *Engine) Scan(ctx context.Context) (*Report, error) {
 	for i, sg := range storeGroups {
 		rfc822IDs[i] = sg.RFC822MessageID
 	}
-	msgsByGroup, err := e.store.GetDuplicateGroupMessagesBatch(
-		rfc822IDs, e.config.AccountSourceIDs...,
+	msgsByGroup, err := e.store.GetDuplicateGroupMessagesBatchContext(
+		ctx, rfc822IDs, e.config.AccountSourceIDs...,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("get duplicate group messages: %w", err)
--- a/internal/dedup/dedup.go
+++ b/internal/dedup/dedup.go
@@ -324,6 +324,17 @@ func (e *Engine) Scan(ctx context.Context) (*Report, error) {
 		return nil, fmt.Errorf("find duplicates: %w", err)
 	}
 
+	rfc822IDs := make([]string, len(storeGroups))
+	for i, sg := range storeGroups {
+		rfc822IDs[i] = sg.RFC822MessageID
+	}
+	msgsByGroup, err := e.store.GetDuplicateGroupMessagesBatch(
+		rfc822IDs, e.config.AccountSourceIDs...,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("get duplicate group messages: %w", err)
+	}
+
 	report := &Report{
 		TotalMessages:   totalMessages,
 		BySourcePair:    make(map[string]int),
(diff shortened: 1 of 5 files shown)


jesserobbins:
Holding off to do more testing

jesserobbins:
I am running into multiple issues on my super old email archive.

jesserobbins:
Okay, was thrown off by #526.

PR #527, opened 2026-07-28
fix(store): make deleted_from_source_at tombstone writers idempotent

danshapiro:
## Problem

The tombstone writers in `internal/store/messages.go` re-stamp
`deleted_from_source_at` on rows that already carry a tombstone. None of them
guard on the column already being set, so each runs

```sql
UPDATE messages SET deleted_from_source_at = <now>
WHERE source_id = ? AND source_message_id IN (...)
```

over the entire already-deleted population every time.

A provider repair scan is meant to be idempotent — it re-reports the same set of
upstream-deleted IDs on every run. Today that has two consequences:

1. **The original deletion time is lost.** After the first re-scan,
   `deleted_from_source_at` records "the last time a repair scan ran" rather than
   "when the message was deleted upstream."
2. **Every such row's `last_modified` is bumped**, so anything that pages over
   modification time re-delivers the whole deleted population on every repair
   scan.

## Fix

Add `AND deleted_from_source_at IS NULL` to the five writers that set the
column, so a tombstone is written once and never rewritten:

- `MarkMessageDeleted`
- `MarkMessagesDeletedBatch`
- `MarkMessagesDeletedFromReader`
- `MarkMessageDeletedByGmailID` (soft-delete branch only)
- `MarkMessagesDeletedByGmailIDBatch`

`MarkMessageDeletedByGmailID`'s `permanent == true` branch is a hard `DELETE` and
is untouched. `ClearMessageDeletedFromSource` still clears the column, so a
message that comes back and is deleted again gets a correct fresh stamp.

## Tests

`internal/store/tombstone_idempotent_test.go` covers all five writers,
table-driven, on both backends:

- an existing stamp is preserved, not rewritten
- a fresh stamp is still set on a row that has none
- `last_modified` is not bumped for an already-tombstoned row
- a mixed batch tombstones only the rows that need it, leaving existing stamps
  intact
- clear-then-re-mark sets a fresh stamp

Each test was confirmed to fail against the unguarded statements before the fix.

The helper reads the seeded sentinel back from the database and compares against
that rather than against the literal it wrote, because PostgreSQL renders a
`TIMESTAMPTZ` through the session's `TimeZone`; comparing against the literal
only holds under UTC.

## Verification

```
go test ./internal/store/...                                   ok
MSGVAULT_TEST_DB=postgres://…  go test ./internal/store/...     ok
```

### DIFF
--- a/internal/store/messages.go
+++ b/internal/store/messages.go
@@ -1326,7 +1326,7 @@ func (s *Store) MarkMessageDeleted(sourceID int64, sourceMessageID string) error
 	_, err := s.db.Exec(fmt.Sprintf(`
 		UPDATE messages
 		SET deleted_from_source_at = %s
-		WHERE source_id = ? AND source_message_id = ?
+		WHERE source_id = ? AND source_message_id = ? AND deleted_from_source_at IS NULL
 	`, s.dialect.Now()), sourceID, sourceMessageID)
 	return err
 }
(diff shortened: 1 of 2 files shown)


wesm:
thanks!

danshapiro:
Look ma, no roborev issues! (granted that one was a layup)

My Book <http://hotseatbook.com>   |   My Twitter
<http://www.twitter.com/danshapiro>   |   My Turtles
<http://www.robotturtles.com/>   |   My, it'd be nice to work together
<http://glowforge.com/jobs>

PR #492, opened 2026-07-20
feat(web): relationships-centric analytical web UI

wesm:
Rebuilds the daemon-served web UI around relationships: the landing view ranks the people and domains you interact with most, and each opens into a workspace with an activity timeline and a mail-client reading pane.

## Relationships hub (new default workspace)

**Ranked list.** People are scored by reciprocal contact:
- sent mail ×2, meetings together ×3, received mail ×1 (log-compressed so newsletter volume can't drown real correspondents)
- 365-day half-life decay, small bonus for contact across multiple channels
- received credit goes to message authors only — mailing lists and co-recipients don't rank
- noise gate: only people you wrote to or met with; "All senders" lifts it

**Per-person workspace.**
- Timeline of emails and meetings, with chat grouped into per-day bursts; one click opens any item in the reading pane
- "Same person…" links a person's addresses and handles into one identity cluster, unlinkable per chip; links refresh the ranked list immediately, and linked identities merge everywhere (People groupings, drill-downs, file lists)
- Resizable persisted sidebar; optional files pane scoped to the person or domain

## Reading experience

- Single click opens content everywhere; the reading pane shows the message inside its conversation, no metadata detour
- Thread view: collapsed messages expand inline; quote chains fold behind "Show quoted text"
- HTML mail: theme-native rendering for simple replies; designed mail on a centered white canvas with author styles preserved through a strict sanitizer allowlist
- Remote images are blocked by default and shown as placeholder chips; per-message "Load images" fetches them through an SSRF-hardened daemon proxy (DNS answers validated, connections IP-pinned, redirects re-checked per hop), so the browser never contacts a sender's host

## Backend

- New store table `participant_links`: identity clusters as a forest; merges rewrite edges; concurrent mutations serialized
- New endpoints: `POST /api/v1/relationships`, `…/relationships/{id}/timeline`, `…/identity/links` + `/unlinks`, `…/content/remote-image`; conversation windows accept optional UTC time bounds
- API schema 1.18.0 → 1.31.0
- Parquet cache derives `is_from_me` from confirmed account identities and exports owner/cluster datasets; identity link changes take a cheap refresh, account-identity changes a full rebuild

## Performance (measured on a 2.5M-message archive)

| Query | Before | After |
|---|---|---|
| Ranked relationships | 3.4 s | 2–3 ms warm, ~1.6 s cold |
| Everything listing | 4.2 s | 0.3 s |
| Files search | 9 s | 0.18 s |
| Person detail | 0.68 s | ~0.4 s |

[...]

PR #531, opened 2026-07-29
Add durable person profiles over identity clusters

salmonumbrella:
Adds a curated person layer over the identity-cluster substrate from #492.

A cluster's canonical ID is derived from component membership, so it re-points whenever links change elsewhere in the component, and importer-driven merges delete participant rows outright — there is nothing stable to hang curated data on. This introduces that stable identity:

- New `persons` table — stable surrogate ID (`AUTOINCREMENT`/`IDENTITY`, so a deleted person's ID is never recycled), reserved unique vCard UID, editable display-name override, and a `revision` counter driving `If-Match` optimistic concurrency — plus `person_participants` bindings, in both SQLite and PostgreSQL schemas. The design contracts (bindings are the source of truth for membership; UIDs are never reused, the survivor keeps its UID on a future person-merge, deletes retire UIDs with no tombstones) are documented in the schema.
- Linking two clusters that already belong to two different persons returns a conflict instead of silently merging curated data. When exactly one person covers the clusters being joined, linking auto-binds the combined cluster's unbound members to it, so person membership never drifts behind cluster membership; participant merges fill the same way and repoint bindings inside the same transaction. The Beeper importer skips (with a warning) merges that would join two different persons rather than aborting the sync.
- Unlinking never moves or deletes bindings, so a person ID and vCard UID survive arbitrary link/unlink churn.
- HTTP API: `POST /api/v1/persons` promotes a participant's cluster into a person (201 on creation, 200 on idempotent re-promotion); `GET`/list/`PATCH` (display name, with stale-revision conflicts) and `DELETE` (permanent, `If-Match`-guarded) round it out. `GET /api/v1/people/{id}` now carries a `profile` block (person ID, display-name override, revision) when the cluster is promoted, so clients can resolve participant → person without listing persons. OpenAPI document and generated clients updated.
- CLI: `msgvault person promote|get|list|set-display-name|delete`.

Usage: promote any participant of a cluster (`msgvault person promote <participant-id>`), then address that person by its stable ID regardless of later identity linking, unlinking, or importer merges. Profiles are only created by explicit promotion — no automatic promotion of observed participants. A binding conflict is resolved by deleting one of the profiles, linking, and re-promoting.

Web UI exposure and read-side integration are tracked in #532.


salmonumbrella:
Both addressed in 9d910c5: the promote CLI now accepts the 201 response (new expected-status variant on the daemon client, with coverage), and any participant-binding insert/repoint/removal bumps the affected person's revision and updated_at in the same transaction so stale If-Match values are rejected.

salmonumbrella:
All three addressed in 3c1a3f3: display_name is now required and nullable on the PATCH body (clients regenerated, null clears the override); person reads hydrate the profile and its bindings from a single transaction, with create/update committing only after hydration succeeds; and CopySubset now carries persons and bindings for retained participants, with a round-trip test.

wesm:
looking

wesm:
I'm going to do some work on this

wesm:
Pushed fab8e516, which closes the operational gaps found in review so the person layer is safe to build UI on:

- **Delete escape hatch.** `DeletePerson` in the store (revision CAS under the identity mutation lock), `DELETE /api/v1/persons/{id}` with required `If-Match` (204/404/409/428), and `msgvault person delete`. A `person_binding_conflict` is no longer a dead end: delete one profile, link, re-promote.
- **Auto-bind on link and merge.** When exactly one person covers the clusters being joined, `LinkParticipants` binds the combined cluster's unbound members to it and bumps its revision; `MergeParticipants` fills the same way. Person membership can no longer silently drift behind cluster membership, and within a cluster bindings stay all-or-none to at most one person.
- **Participant → person lookup.** `GET /api/v1/people/{id}` now returns a `profile` block (person ID, display-name override, revision) when the cluster is promoted, so the web UI can show curated identity and drive PATCH/DELETE from a detail view without listing all persons.
- **ID durability fix.** The new delete test exposed that SQLite recycles plain `INTEGER PRIMARY KEY` rowids, so a deleted person's ID would have been reassigned to the next person. `persons.id` is now `AUTOINCREMENT` (PostgreSQL's `IDENTITY` already never reuses).
- **Design contracts documented in both schemas.** Bindings are the source of truth for membership; vCard UIDs are random and never reused (a future person-merge keeps the survivor's UID and retires the other); deletes are permanent with no tombstones.
- **Hygiene.** Promote returns 200 on idempotent re-promotion vs 201 on creation (CLI accepts both); `persons` added to the PG exclusive-lock table set (merge bumps person revisions on the import path); 409 message casing unified; If-Match strictness (no `*`, no tag lists) and the intentionally unpaginated persons list documented in the OpenAPI descriptions.

One deliberate semantic to be aware of: auto-bind plus "unlink never unbinds" means a briefly linked participant stays a person member after the link is retracted; the remedy today is delete + re-promote. That trade-off, along with web UI exposure, read-side integration (display-name override in analytics/search, messages-for-person), person-merge, and Beeper skip counts in import stats, is tracked in #532.

Verified: build, vet, golangci-lint clean; full test suite passes with the `fts5 sqlite_vec` tags; OpenAPI YAMLs, Go client, and web `schema.d.ts` regenerated with no drift. The PostgreSQL lane (`make test-pg`) was not run locally — no PG instance here — so it should be covered by CI or a manual run before merge.

wesm:
Both findings confirmed and fixed in 8c8055b5:

- **Client 200 handling** — added a hand-written `CreatePerson` wrapper in `pkg/client/client.go` following the existing `AddAccount`/`StageDeletion` pattern (those exist for exactly this generated-client limitation), returning `JSON201` or `JSON200`. Regression test `TestCreatePersonAcceptsIdempotentOK` covers the 200 path.

- **PG lock inversion** — real deadlock: `MergeParticipants` writes `person_participants`/`persons` before `participants`/`messages`, the opposite of `BeginExclusive`'s `LOCK TABLE` order. Rather than reordering the table list (fragile against future writers), `BeginExclusive` now takes the identity-mutation row lock (the `archive_metadata` identity-revision row that link/unlink/merge/promote/delete all acquire first) before `LOCK TABLE`, serializing the two paths at a single point. Added `TestRemoveSourceSerializedDoesNotDeadlockWithPersonMerge` racing `RemoveSourceSerialized` against `MergeParticipants` over 15 iterations; verified against a live PostgreSQL 17 instance that it fails with `SQLSTATE 40P01` without the row lock and passes with it. Full `make test-pg` suite is green.

wesm:
Confirmed and fixed in 973b040e. `CopySubset` selected participants only by message involvement and never copied `participant_links` (a gap that predates this PR, but the person-binding copy made it concrete: bindings were silently filtered to message-bearing members while the cluster arrived disconnected).

The fix pulls each copied participant's full link component in first — cluster-mates resolved via a recursive CTE over `src.participant_links`, which covers connectivity through intermediate members with no copied messages — then copies every edge inside those components. Destination clusters, `/people/{id}` aggregation scope, and person bindings now match the source.

`TestCopySubset_PreservesIdentityClusters` covers a promoted linked cluster whose second member has no messages in the subset; verified it fails (disconnected cluster, missing binding) without the fix and passes with it.

wesm:
All three confirmed and fixed in c21c2529. The two subset findings are resolved with one consistency policy rather than two patches:

- **Privacy boundary (default)** — the participant set is message-derived again: no participant row, identifier, link edge, or person binding is copied for identities without selected messages. Link edges are kept only when both endpoints are already included (no new PII), and a person is copied only when *every* binding falls inside the subset — a partial binding set under the original revision would misrepresent the curated profile, so partially-covered persons are skipped entirely instead of truncated. `TestCopySubset_DefaultExcludesOffMessageIdentities` pins the boundary.

- **Complete profiles (opt-in)** — `create-subset --include-identity` expands participants through the closure of `participant_links` **and** shared `person_participants` bindings, so included clusters and profiles arrive complete — including the requested regression case of a person left spanning disconnected clusters by an unlink (`TestCopySubset_IncludeIdentitySpansUnlinkedClusters`). The flag help spells out that this exposes identifiers of linked identities with no messages in the subset.

- **Migration lock order** — real inversion: the legacy identity-config migration wrote `account_identities` before bumping the identity revision, opposite to `BeginExclusive`'s row-lock-then-`LOCK TABLE` order. It now takes `lockIdentityMutationTxContext` at the start of its transaction, matching every other revision writer (audited: link/unlink/merge/add/remove-identity all lock first; the phone-unique migration runs inside `InitSchema` before the daemon serves). Added `TestRemoveSourceSerializedDoesNotDeadlockWithLegacyIdentityMigration`; verified against live PostgreSQL 17 that it reproduces `SQLSTATE 40P01` without the fix. The blocked-cancellation test was updated to pin the new up-front lock statement.

Full SQLite and PostgreSQL suites plus lint are green.

wesm:
CI failures on c21c2529 had a single root cause: the four failing jobs run the PR merged with latest main, and main's new cache-staleness tests (#528) link participants on the hand-rolled fixture schemas in `build_cache_test.go` — which predate the `persons`/`person_participants` tables that `LinkParticipants` now maintains, so they failed with `no such table: person_participants`. Fixed in 42edcc76 by merging main into the branch and mirroring the persons tables into both fixture schemas. Full SQLite and PostgreSQL suites plus lint are green on the merged tree.

wesm:
CI is green on 0cf7f015 (all jobs, including the four that failed and test-pgvector).

Two fixes were needed:

1. **Fixture drift from the merge with main** (42edcc76): main's new cache-staleness tests link participants on the hand-rolled fixture schemas in `build_cache_test.go`, which predate the `persons` tables that `LinkParticipants` now maintains. Mirrored the two table definitions into both fixture schemas.

2. **test-pgvector timeout was a real perf regression** (0cf7f015): the deadlock-ordering fix had the legacy identity migration take the identity-mutation row lock unconditionally at transaction start — turning every store open into a write transaction (row-version churn + WAL commit on one `archive_metadata` row, and a serialization point between the daemon and CLI subprocesses sharing a DB). Cheap locally, but it blew the `cmd` package past the 10-minute timeout on CI (main baseline: 72s). The no-op path is read-only again: marker check first without locks; only when the migration will actually write does it take the row lock (still before any `account_identities` write, preserving the ordering contract), then re-checks the marker under the lock. Local PG timing for the package went from 155s back to under the main baseline.

wesm:
Confirmed and fixed in 52bcec1e. `SetParticipantIdentifier` wrote `participant_identifiers` before its conditional owner-evidence identity-revision bump — the same inversion class as the migration. The no-op fast path (constant on importer re-runs) stays read-only; the identity-mutation row lock is taken only on the write path, before the table write, with the no-op re-checked under the lock. New regression test `TestRemoveSourceSerializedDoesNotDeadlockWithSetParticipantIdentifier` reproduces `SQLSTATE 40P01` 8 times across 5 runs without the fix on live PostgreSQL 17, and passes with it.

Also completed the requested audit of every identity-revision writer: link, unlink, merge, add/remove account identity, the legacy identity migration, and person mutations all lock first. One more late acquirer surfaced — the one-shot phone-unique migration's `mergeParticipant` took the lock mid-transaction via `rewriteLinksForMerge`; it now locks at entry. The ordering contract is now documented on `lockIdentityMutationTx` so future revision writers have a single rule to follow: acquire the identity row before the first write to any `exclusiveLockTables` table, keeping no-op fast paths read-only.

Full SQLite and PostgreSQL suites plus lint green; the PG `cmd` package stays at ~90s (no repeat of the timeout regression).

PR #533, opened 2026-07-30
Add lossless registry-backed vCard codec

salmonumbrella:
## What changed

- Vendors the current IANA vCard property and parameter registries with explicit handling declarations and drift/update commands.
- Adds bounded, ordered decode and render support for vCard 2.1, 3.0, and 4.0 while preserving groups, source spelling, parameter order and quoting, raw values, repeated properties, and unknown extensions.
- Adds reusable text, structured/list, RFC 6868, quoted-printable, partial-date, registry lookup, and semantic validation helpers.
- Keeps the existing `Contact` and `ParseFile` API as a compatible projection over the lossless codec.

## Why

Richer person profiles and CardDAV need a standards boundary that can preserve source syntax and data Msgvault does not yet understand. The previous contact projection discarded that information, making future native mapping and round trips lossy.

This is the first focused step in #534. Native profile mapping, resource persistence, and CardDAV serving remain separate follow-up boundaries.

## Usage

Run `make vcard-registry-check` to report upstream registry drift. Run `make vcard-registry-update` to fetch and atomically write a validated snapshot for review.


salmonumbrella:
Fixed in `cb3df06`.

- TEL URI projection now excludes `ext` from E.164 identity keys and resolves local subscribers with numeric `phone-context` separately.
- Legacy text decoding now applies quoted-printable transfer decoding before the declared `CHARSET` conversion; literal and quoted-printable ISO-8859-1 vCard 2.1 values are covered.
- Undeclared invalid UTF-8 remains rejected.

The PR remains one commit.

salmonumbrella:
Fixed in `600646e`.

- Quoted-printable folding now chooses boundaries outside `=XX` escapes, preventing trailing `=` from being consumed as a soft break.
- Charset-declared legacy bytes accepted by the decoder are preserved by the lossless encoder and can round-trip through folding.
- Quote-aware parsing now caps each property at 100 parameters and each parameter at 100 values before allocation amplification; boundary tests cover 100 and 101.

Focused and repository-wide tests, lint, vet, and testify checks pass. The PR remains one commit.

salmonumbrella:
Fixed in `6e218d6`.

- Decoding now streams physical lines through unfolding and card parsing, so card and logical-line limits stop reads before trailing input is consumed.
- Folded continuations use amortized byte-buffer growth, and quoted-printable header state is cached instead of repeatedly rescanning a growing line.
- Added early-stop regressions for card/logical limits and allocation coverage for heavily folded input.

The PR remains one commit.

salmonumbrella:
Fixed in c50f6e9. RFC 6868 parameter escaping now follows the containing card version: vCard 2.1/3.0 retain literal caret sequences, while 4.0 continues to decode and encode RFC 6868. Added 2.1/3.0 decode and constructed-parameter regression coverage.

wesm:
Looking at this today

wesm:
Addressed the follow-up review in `c4f89ce8`:

- Hardened constructed parameter encoding so preserved raw values and explicit bare v2.1 parameters cannot change structure on reparse.
- Made `DecodeOptions` zero and partial values accept v2.1 through explicit `DisallowV21`, and documented the codec invariants.
- Added v4 `VERSION` ordering validation and fixed configured registry source metadata.
- Removed the unused handling notes field and clarified the manual versus CI registry drift boundary.

Registry lookup indexing/caching and strict third-party v2.1 quoted-printable folding remain separate follow-ups.

wesm:
Addressed the combined-review finding in `afbd98da`: bare parameters now use the same quote-aware, comma-separated value parsing as named parameters, and the encoder preserves quoted and multiple bare values without adding `TYPE=`. Round-trip regressions cover quoted values, `HOME,VOICE`, and quoted delimiter values.

wesm:
Addressed the resource-bound finding in `202d595f`: `DecodeOptions` now exposes `MaxPropertiesPerCard` with a 10,000-property default, validates explicit limits, and rejects the next property before appending it. Regression coverage includes the custom and default boundaries, invalid options, and early termination before trailing input is consumed.

Issue #534, opened 2026-07-30
Roadmap: richer people profiles, relationships, and CardDAV


salmonumbrella:
The durable person root introduced in #531 is the base for a broader People and CardDAV upgrade. This issue tracks the direction so each focused pull request can stay reviewable without losing the larger context.

## Outcomes

- Rich, typed person profiles with provenance, history, custom fields, structured names, contact points, addresses, dates, media, and categories.
- Organizations, complete employment history, and typed temporal relationships represented as first-class records rather than copied display fields.
- Explainable identity resolution that preserves source evidence and requires review for ambiguous matches instead of merging people by name or email alone.
- A dated activity layer that derives contact recency and cadence from archived messages, chats, and meetings without request-time full-archive scans.
- Person-scoped attachments: scoped collection policy that keeps the media worth keeping, searchable document and visual attachments, a per-person Media and Files gallery, and the same person-scoped question answerable from the web, MCP, and the CLI.
- A scheduled people sweep that keeps curated attributes current from the archive: deterministic fields for everyone, evidence-backed extracted candidates for tracked people with a per-attribute auto-apply policy, bitemporal supersession instead of overwrites, and a dated "last time we talked" brief.
- Lossless vCard support with stable UID continuity and preservation of unknown properties, and a standards-only CardDAV client that syncs curated people into the user's existing address book server. Msgvault does not serve CardDAV.
- A web Directory workspace and bounded people-network projection built on the same typed APIs.
- Optional contact enrichment inspired by [Mesh](https://me.sh/), with contact details, public profiles, organizations, and employment changes.
- A fresh install that reaches all of the above from a pasted Voyage key and Mistral key, one consent per hosted provider, and recommended defaults everywhere else.

## Delivery sequence

1. Establish the lossless, registry-backed vCard syntax boundary (#533).
2. Add portable field metadata and typed, historized custom attributes (#535).
3. Add structured profile primitives and service-scoped contact observations (#565).
4. Add reviewable identity candidates for importer evidence (#596).
5. Add organizations and temporal employment associations (#593).
6. Add canonical typed person relationships (#592).
7. Make chats and meetings semantically retrievable for relationship queries: conversation-window embedding units and turn-aware transcript chunking (#589).
8. Embed curated person attributes into one document per person so semantic people search ("a finance expert in New York") returns the contact directly (#656).
[...]

PR #535, opened 2026-07-30
Add portable field metadata and typed person attributes

salmonumbrella:
## What changed

- Added a portable field-definition registry with stable universal IDs, immutable slugs, mutable labels, typed storage, cardinality, capabilities, options, vCard mapping, and lifecycle state.
- Added typed, historized person attribute values with value-level provenance, confidence, actor, ordinal, world-time intervals, and transactional supersession.
- Seeded `primary_channel`, `contact_frequency`, `ask_me_about`, and the read-only derived `last_contacted`.
- Added definition and person-attribute API routes, generated clients, and non-interactive CLI workflows with dry-run support.

## Why

A JSON-only profile blob cannot preserve independently addressable values with their own provenance, confidence, actor, ordinal, and active interval. Typed value rows retain that history and keep current-value lookups indexable.

Dynamic per-user DDL would require migration, synchronization, permission, and API machinery for every custom field. Metadata rows plus concrete typed tables keep field creation portable across SQLite and PostgreSQL without issuing runtime DDL.

This is the second implementation slice in #534.

## Usage

- `msgvault attribute-definition list --object-type person`
- `msgvault attribute-definition create --definition @field.json`
- `msgvault person attributes set <person-id> ask_me_about --value sailing`
- `msgvault person attributes list <person-id> --history`

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -195,6 +195,159 @@ components:
         - mime_type
         - size_bytes
       type: object
+    AttributeChoice:
+      additionalProperties: false
+      properties:
+        label:
+          type: string
+        value:
+          type: string
+      required:
+        - value
+        - label
+      type: object
+    AttributeDefinition:
+      additionalProperties: true
+      properties:
+        api_mutable:
+          type: boolean
+        cardinality:
+          type: string
+        created_at:
+          format: date-time
+          type: string
+        derived_source:
+          type: string
+        description:
+          type: string
+        display_order:
+          format: int64
+          type: integer
+        field_type:
+          type: string
+        history_exempt:
+          type: boolean
+        id:
+          format: int64
+          type: integer
+        is_active:
+          type: boolean
+        is_audited:
+          type: boolean
+        is_deletable:
+          type: boolean
+        is_required:
+          type: boolean
+        is_searchable:
+          type: boolean
+        label:
+          type: string
+        object_type:
+          type: string
+        options:
+          $ref: "#/components/schemas/AttributeOptions"
+        ownership:
+          type: string
+        record_target:
+          type: string
+        revision:
+          format: int64
+          type: integer
+        slug:
+          type: string
+        ui_creatable:
+          type: boolean
+        ui_editable:
+          type: boolean
+        universal_id:
+          type: string
+        updated_at:
+          format: date-time
+          type: string
+        value_type:
+          type: string
+        vcard_property:
+          type: string
+      required:
+        - id
+        - universal_id
+        - object_type
+        - slug
+        - label
+        - value_type
+        - field_type
+        - cardinality
+        - display_order
+        - is_required
+        - ownership
+        - ui_creatable
+        - ui_editable
+        - api_mutable
+        - is_searchable
+        - is_audited
+        - is_deletable
+        - history_exempt
+        - is_active
+        - revision
+        - created_at
+        - updated_at
+      type: object
+    AttributeDefinitionsResponse:
+      additionalProperties: true
+      properties:
+        definitions:
+          items:
+            $ref: "#/components/schemas/AttributeDefinition"
+          type:
+            - array
+            - "null"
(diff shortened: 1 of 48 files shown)


salmonumbrella:
Fixed in `69feb99`.

- The production `storeAPIAdapter` now implements both attribute capabilities through explicit pass-through methods, with compile-time assertions and a real daemon-adapter route regression.
- Person deletion now returns a conflict while another profile has any stored person reference; reference creation and deletion share the identity-mutation lock so the check is atomic on SQLite and PostgreSQL.
- Attribute closure now records `active_until` from the requested validity boundary and `superseded_at` from the actual write time, including backdated replacement coverage.

The PR remains one commit.

salmonumbrella:
Fixed in `43f836f`.

- Subset creation now reconciles person attribute definitions by `universal_id`, remaps value/history rows to destination-local definition IDs, omits references to excluded people by default, and follows person references directionally under `--include-identity`.
- Select and multiselect definitions now reject non-canonical JSON/record-reference value types, parse scalar choices by their declared type, store canonical forms, and reject invalid or canonically duplicate choices.
- Added real subset history/reference-policy coverage and scalar choice validation coverage.

The PR remains one commit.

salmonumbrella:
Fixed in amended head 24630e3: shared subsets now exclude person attribute definitions and values by default; --include-attributes explicitly includes current and historical values and emits a sensitive-data warning. Explicit --expected-value-id values below 1 are rejected before any API request, while valid IDs are forwarded. Record-reference widgets now require a matching target, so organization widgets are rejected until organization targets are supported.

salmonumbrella:
Fixed in c8d111e6cdb584a491ed5789886658634862801d. Non-record-reference values now reject either record field, and the request-level regression covers stray record_type and record_id as 400 responses.

salmonumbrella:
Resolved in `a2316eb4e918d10c304d65f02ef42c535f0b365c`. The person-attribute DELETE API now accepts an optional `expected_value_id`, validates and forwards it to the store CAS guard, and returns `409` without clearing when the current value has changed. Generated Go/web clients and `msgvault person attributes clear --expected-value-id` expose the guard; regression coverage verifies a stale clear preserves the newer value.

wesm:
Reviewed this against the intent in #534 (slice 2) and applied fixes on a copy of the branch rebased onto current main.

**Review outcome**
- The slice matches #534's constraints: field creation is metadata-only (no runtime DDL), values are typed and provenance-aware with retained history, SQLite/PostgreSQL behavior is kept equivalent via identical schema contracts plus shared Go-side validation, and fixtures are synthetic. `last_contacted` is seeded with no producer yet, which the CLI discloses.

**Fixes applied**
- Person deletion: only *current* record references block deletion. Previously a superseded reference made the target person permanently undeletable, since there is no value purge path.
- Ordinal allocation scans historical rows too, so an appended multi-value can no longer reuse a superseded slot's ordinal and show up as that slot's successor in history.
- Inactive definitions: stored values stay listed (while any remain) and can still be cleared; only setting new values requires an active definition. The `inactive` and `has_values` 409 messages now explain how to recover.
- Supersede uses the same busy/conflict retry loop as set (SQLite read-then-write transactions can fail snapshot upgrade under contention).
- Reseeding no longer resets a user's `display_order` override on restart.
- Added a partial index covering the delete-time record-reference scan, which was a full table scan of `person_attribute_values`.
- PUT with a non-positive `expected_value_id` returns 400, matching the DELETE path.
- CLI tests reset bound flag values between runs — a leaked `--dry-run` from one test was silently appended to a later test's request, masking a potential regression. Also dropped a tautological subcommand test.

**Left for follow-up**
- If-Match parsing duplication, the unused `GetAttributeDefinitionBySlugContext` interface method, and 400-response enumeration on list routes in the OpenAPI spec — each follows an existing pattern or is guarded elsewhere.

wesm:
continuing on this

PR #536, opened 2026-08-01
feat(mcp): add bearer authentication to Streamable HTTP

salmonumbrella:
Streamable HTTP MCP currently has no built-in authentication, so operators must keep it on loopback or provide their own authenticated network boundary. msgvault already has an application credential for archive access in `[server].api_key`; this reuses that credential for the HTTP MCP transport instead of introducing a second secret.

When `[server].api_key` is configured, every `/mcp` request must send `Authorization: Bearer <api_key>`. Missing or incorrect credentials return `401 Unauthorized`, including session `GET` and `DELETE` requests. A configured key also permits non-loopback binding without `--http-allow-insecure`.

Usage:

```toml
[server]
api_key = "replace-with-a-long-random-key"
```

Configure the MCP client to send the same value as a bearer token on every request. Without an API key, existing behavior is unchanged: loopback works normally, while non-loopback binding still requires the explicit insecure override. The built-in listener remains plain HTTP, so non-loopback connections still need a TLS proxy or encrypted private network.

### DIFF
--- a/cmd/msgvault/cmd/mcp.go
+++ b/cmd/msgvault/cmd/mcp.go
@@ -18,6 +18,7 @@ var mcpForceSQL bool
 var mcpNoSQLiteScanner bool
 var mcpHTTPAddr string
 var mcpHTTPAllowInsecure bool
+var serveMCPHTTPWithOptions = mcpserver.ServeHTTPWithOptions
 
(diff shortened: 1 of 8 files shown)


wesm:
Looks good. The documentation needs to be updated, I'm doing that now

PR #537, opened 2026-08-02
docs: add msgvault favicon branding

salmonumbrella:
## Summary

- add a self-contained 256×256 vector mark based on the existing Kenn wordmark style
- center the new mark, project title, badges, and navigation links in the README

image

## Maintainer note

This is intentionally offered as an optional branding addition. Please feel free to merge it as-is, reuse only the SVG, or skip it if the current presentation is preferred. The SVG is scalable and can also feed the existing docs favicon asset workflow if desired.


wesm:
Seems reasonable, we can always refine later once Kenn has more brand assets

Issue #397, opened 2026-06-19
Planned (deferred) work: survivor metadata-quality tie-break and identity discovery signals


jesserobbins:
## Context

The AIC/dedup [design spec](https://github.com/kenn-io/msgvault/blob/0cfc8b7147bbab8fda2dd560ec15ac6afc266df3/docs/accounts-identities-collections-dedup/spec.md) describes two capabilities that the initial implementation deliberately did not ship. These were **planned from the start** — not drift, not accidental omissions, and not errors in the spec. They're future work that's still worth doing. Filing this so the design-vs-shipped gap is tracked and can be picked up.

The user docs were aligned to the *shipped* behavior ([kenn-io/msgvault-docs#30](https://github.com/kenn-io/msgvault-docs/issues/30), [#31](https://github.com/kenn-io/msgvault-docs/pull/31)) so they don't describe unbuilt features. This issue is the home for the planned remainder.

### 1. Survivor selection — "source metadata quality" tie-break

The spec's [priority list](https://github.com/kenn-io/msgvault/blob/0cfc8b7147bbab8fda2dd560ec15ac6afc266df3/docs/accounts-identities-collections-dedup/spec.md#L306-L320) includes [step 3 — source metadata quality](https://github.com/kenn-io/msgvault/blob/0cfc8b7147bbab8fda2dd560ec15ac6afc266df3/docs/accounts-identities-collections-dedup/spec.md#L312) (provider IDs, threading info, Message-ID presence). The shipped comparator [`Engine.isBetter`](https://github.com/kenn-io/msgvault/blob/0cfc8b7147bbab8fda2dd560ec15ac6afc266df3/internal/dedup/dedup.go#L792-L811) implements the other five steps (source priority → raw MIME → label count → earliest `archived_at` → row ID) but not this one.

**Planned:** add the metadata-quality comparison as a tie-break between raw-MIME presence and label count, per the spec ordering.

### 2. Identity discovery signals

The spec's [Discovery signals](https://github.com/kenn-io/msgvault/blob/0cfc8b7147bbab8fda2dd560ec15ac6afc266df3/docs/accounts-identities-collections-dedup/spec.md#L182-L205) section describes deriving identities from observed message data — `is_from_me` (ingest metadata), `sent-folder`/`sent-label` (sent-mail placement), and `oauth` (provider account metadata). The shipped code confirms identities only at account-setup time, writing `account-identifier`, `phone-e164`, `config_migration`, and `manual`.

[...]

PR #540, opened 2026-08-03
feat: discover source-scoped email identities

### LINKED ISSUE #498: identity commands cannot target accounts sharing an identifier
## What I was doing

I was adding a manual identity to an account whose identifier is shared by sources of different types, such as a live Gmail account and an MBOX import.

I authored the original accounts, identities, and collections model in #278. Identity is intentionally scoped to one account/source, so these sources need separate identity sets.

## What happened

Given two valid sources:

```text
test@example.com (gmail, id=1)
test@example.com (mbox, id=2)
```

Running:

```console
msgvault identity add test@example.com alias@example.com
```

returns:

```text
ambiguous account "test@example.com" matches multiple sources:
[test@example.com (gmail, id=1) test@example.com (mbox, id=2)]
```

The error identifies both source types and IDs, but `identity add` accepts neither as a selector. The same account resolver is used by `identity show`, `identity remove`, and `identity list --account`.

There is no supported CLI or daemon API path to target either source.

Reproduced on `upstream/main` at `74ea3045`.

## Expected behavior

Identity commands should accept a unique source selector when an account identifier or display name is ambiguous.

Source ID is the reliable selector because display names can also collide within the same source type. A `--type` selector, like the one supported by `remove-account`, would handle the common cross-type case but not every valid collision.

The existing account-name syntax should continue to work when it resolves to one source.

## Why it matters

The schema permits the same identifier across source types through its `(source_type, identifier)` uniqueness rule. This is a normal archive shape, especially when the same account exists as both a live source and an imported archive.

Identity is source-scoped by design. The current command surface makes those source-scoped identities unreachable precisely when more than one valid source shares a name.

## Related issues

- #278 introduced the source-scoped accounts, identities, and collections model. This bug is a follow-up to that design.
- #311 and #315 track identity comparison and storage invariants. They do not cover account selection.
- #307 covers a missing auto-default identity after WhatsApp imports.
- #397 tracks deferred automatic identity discovery signals.
- #12 led to `remove-account --type`, which is the existing disambiguation precedent.

## Acceptance criteria

- `identity show`, `add`, `remove`, and scoped `list` can target one source using a unique selector.
- Unambiguous account names continue to work unchanged.
- Ambiguous names without a selector still fail and list the matching source types and IDs.
- The CLI and daemon API support the same selection behavior.
- CLI documentation explains how to use the selector.
- Tests cover identical identifiers across source types and duplicate display names within one source type.

salmonumbrella:
## What changed

- add source-scoped `identity discover` preview/apply and bulk `identity import`, with unambiguous `--source-id` selection across identity commands
- discover strong sender identities from trusted Sent metadata during full and incremental sync while keeping recipient-only evidence review-only
- optionally inventory Fastmail masked and send-as addresses from a source-scoped API token, retaining disabled and deleted addresses for historical mail
- expose matched sender/recipient identities and source-scoped identity filtering through the API and web UI

## Why

A mailbox can receive and send through many masked, delegated, or historical addresses without those addresses being separate ingestion sources. Confirming them under the existing account identity set makes sent-copy classification and dedup planning accurate while preserving the original message headers and mailbox provenance.

## Usage

Preview archived evidence, then apply strong candidates:

```console
msgvault identity discover --source-id 14
msgvault identity discover --source-id 14 --apply
```

For Fastmail inventory, configure a source-specific token and opt in on the command:

```toml
[[fastmail]]
source_id = 14
api_token = "replace-with-a-Fastmail-API-token"
```

```console
msgvault identity discover --source-id 14 --provider
msgvault identity discover --source-id 14 --provider --apply
```

Addresses can also be previewed or confirmed from a provider export with `msgvault identity import --source-id 14 --file aliases.json [--apply]`.

Addresses the identity-discovery portion of #397.

Closes #498.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -81,8 +81,10 @@ components:
           type: string
         signal:
           type: string
+        source_id:
+          format: int64
+          type: integer
       required:
-        - account
         - identifier
         - signal
       type: object
@@ -812,6 +814,55 @@ components:
       required:
         - id
         - status
+      type: object
+    Candidate:
+      additionalProperties: true
+      properties:
+        already_confirmed:
+          type: boolean
+        classification:
+          enum:
+            - confirmed
+            - strong
+            - weak
+          type: string
+        first_seen_at:
+          format: date-time
+          type: string
+        identifier:
+          type: string
+        last_seen_at:
+          format: date-time
+          type: string
+        normalized_identifier:
+          type: string
+        provider_states:
+          items:
+            type: string
+          type: array
+        received_message_count:
+          format: int64
+          type: integer
+        sent_message_count:
+          format: int64
+          type: integer
+        signals:
+          items:
+            type: string
+          type:
+            - array
+            - "null"
+      required:
+        - identifier
+        - normalized_identifier
+        - classification
+        - already_confirmed
+        - signals
+        - provider_states
+        - sent_message_count
+        - received_message_count
+        - first_seen_at
+        - last_seen_at
       type: object
     ChangedMessageJSON:
       additionalProperties: true
@@ -1556,6 +1607,114 @@ components:
         - description
         - message_count
       type: object
+    DiscoverError:
+      additionalProperties: true
+      properties:
+        code:
+          type: string
+        message:
+          type: string
+      required:
+        - code
+        - message
+      type: object
+    DiscoverEvent:
+      additionalProperties: true
+      properties:
+        error:
+          $ref: "#/components/schemas/DiscoverError"
+        progress:
+          $ref: "#/components/schemas/DiscoverProgress"
+        result:
+          $ref: "#/components/schemas/DiscoverResult"
+        type:
+          enum:
+            - progress
+            - result
+            - error
+          type: string
+      required:
+        - type
+      type: object
+    DiscoverProgress:
+      additionalProperties: true
+      properties:
+        candidates:
+          format: int64
+          type: integer
+        done:
+          format: int64
+          type: integer
+        total:
+          format: int64
+          type: integer
+      required:
+        - done
+        - total
+        - candidates
+      type: object
+    DiscoverRequest:
+      additionalProperties: false
+      properties:
+        account:
+          type: string
+        apply:
+          type: boolean
+        confirm:
+          items:
+            type: string
+          type:
+            - array
+            - "null"
+        page_size:
+          format: int64
+          type: integer
+        provider:
+          type: boolean
+        source_id:
+          format: int64
+          type: integer
+      type: object
+    DiscoverResult:
+      additionalProperties: true
+      properties:
+        account:
+          type: string
+        applied:
+          items:
+            $ref: "#/components/schemas/IdentityConfirmationOutcome"
+          type:
+            - array
+            - "null"
+        candidates:
+          items:
+            $ref: "#/components/schemas/Candidate"
+          type:
+            - array
+            - "null"
+        rejected:
+          items:
+            $ref: "#/components/schemas/RejectedCandidate"
+          type:
+            - array
+            - "null"
+        scanned_messages:
+          format: int64
+          type: integer
+        source_id:
+          format: int64
+          type: integer
+        source_type:
+          type: string
+      required:
+        - account
+        - source_id
+        - source_type
+        - scanned_messages
+        - candidates
+        - rejected
+        - applied
+      type: object
     DomainContextSummaryHTTPResponse:
       additionalProperties: true
       properties:
@@ -1666,6 +1825,14 @@ components:
           type: string
         match:
           $ref: "#/components/schemas/MatchSummary"
+        matched_recipient_identities:
+          items:
+            type: string
+          type: array
+        matched_sender_identities:
+          items:
+            type: string
+          type: array
         message_count:
           format: int64
           type: integer
@@ -1710,6 +1877,8 @@ components:
         - conversation_type
         - title
         - preview
+        - matched_sender_identities
+        - matched_recipient_identities
         - message_count
         - has_attachments
         - attachment_count
@@ -1894,6 +2063,7 @@ components:
             - after
             - before
             - deletion
+            - identity
           type: string
         values:
           items:
@@ -2781,6 +2951,24 @@ components:
         - took_ms
         - results
       type: object
+    IdentityConfirmationOutcome:
+      additionalProperties: true
+      properties:
+        added:
+          type: boolean
+        identifier:
+          type: string
+        signals:
+          items:
+            type: string
+          type:
+            - array
+            - "null"
+      required:
+        - identifier
+        - added
+        - signals
+      type: object
     IdentityLinkRequest:
       additionalProperties: false
       properties:
@@ -2846,6 +3034,60 @@ components:
       required:
         - field
         - direction
+      type: object
+    ImportEntry:
+      additionalProperties: false
+      properties:
+        identifier:
+          type: string
+        state:
+          type: string
+      required:
+        - identifier
+      type: object
+    ImportRequest:
+      additionalProperties: false
+      properties:
+        account:
+          type: string
+        apply:
+          type: boolean
+        entries:
+          items:
+            $ref: "#/components/schemas/ImportEntry"
+          type: array
+        signal:
+          type: string
+        source_id:
+          format: int64
+          type: integer
+      required:
+        - entries
+      type: object
+    ImportResult:
+      additionalProperties: true
+      properties:
+        account:
+          type: string
+        applied:
+          items:
+            $ref: "#/components/schemas/IdentityConfirmationOutcome"
+          type: array
+        candidates:
+          items:
+            $ref: "#/components/schemas/Candidate"
+          type: array
+        signal:
+          type: string
+        source_id:
+          format: int64
+          type: integer
+      required:
+        - account
+        - source_id
+        - signal
+        - candidates
+        - applied
       type: object
     ListDeletionsResponse:
       additionalProperties: true
@@ -3662,6 +3904,17 @@ components:
         - rows
         - row_count
       type: object
+    RejectedCandidate:
+      additionalProperties: true
+      properties:
+        identifier:
+          type: string
+        reason:
+          type: string
+      required:
+        - identifier
+        - reason
+      type: object
     RelationshipRow:
       additionalProperties: true
       properties:
@@ -3834,8 +4087,10 @@ components:
           type: string
         identifier:
           type: string
+        source_id:
+          format: int64
+          type: integer
       required:
-        - account
         - identifier
       type: object
     RemoveResult:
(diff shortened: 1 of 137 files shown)


wesm:
There is a lot of strong work here: source scoping, endpoint security, redaction, idempotency, and test coverage are all thoughtfully handled. I do think four issues need to be addressed before merge:

1. Discovery currently creates a circular trust path. It reads the effective `messages.is_from_me`, which includes attribution derived from confirmed identities, and then treats that value as new strong evidence. Confirming one identifier can therefore cause another address on the same participant to be auto-confirmed later. Please have discovery read `source_is_from_me` so only provider-native attribution counts as independent evidence.

2. Sync automatically confirms any single-From address found in trusted Sent placement. One misplaced, migrated, or spoofed message can silently widen the account identity. Although `identity remove` exists, a later sync can add the address again. Please require `identity discover --apply` for first-time confirmation, add a persistent rejection state, or limit sync-time discovery to refreshing already-confirmed identities.

   This matters because confirmed identities affect dedup survivor selection and can flow into the separately opt-in remote deletion plan.

3. An identity-discovery error currently fails the whole sync after the messages have already been saved. For full sync, the run is marked failed and will not resume from its checkpoint, so a late ancillary failure can force the next sync to start over. Please move discovery to a durable post-sync/retry path or otherwise preserve resumability. The bounded retry should also use backoff and retain the underlying error. Logging a warning alone is insufficient if advancing the checkpoint permanently skips discovery for that page.

4. Each insert-bearing 500-identity chunk refreshes attribution across every message in the source while holding the writer transaction. This will be expensive for large archives and masked-address imports. Please update only affected senders or refresh once after the complete operation, and add normalized/expression indexes for attribution and explore-filter resolution.

A few smaller correctness and operability items are worth fixing here or tracking explicitly:

- Preserve a safe Fastmail failure class or HTTP status in logs so expired credentials are diagnosable.
- Do not classify database failures from `FastmailSourceFor` as HTTP 400.
- Avoid a JMAP inventory request on every no-op incremental sync.
- Give identity import a timeout policy consistent with the other identity mutations.
- Make explore-filter matching follow the same phone, identifier-type, and case rules as baked `is_from_me` attribution.
- Keep an active identity filter removable when the catalog request fails, and display restored identities that are no longer in the catalog.
- Remove the redundant SQLite `(source_id, id)` index and unused label-role index, and avoid blocking PostgreSQL writes during the one-time index build.

Once the four main issues are resolved, this looks close to ready.

salmonumbrella:
All items addressed in 9b960481.

**1. Circular trust** — Discovery scans now read `source_is_from_me` only, with a regression test pinning that identity-derived attribution is never read back as evidence. Side effect worth knowing: Gmail/IMAP sync produces no `is_from_me` signal at all now (it was purely identity-derived there); Sent placement remains the strong signal.

**2. Auto-confirm** — Sync-time discovery is refresh-only, enforced at the store layer: its merge write has no insert path and checks row existence inside the writer-locked transaction, so sync can never confirm a first-time identity — even racing `identity remove` (test-pinned). First-time confirmation requires explicit `--apply` / `identity add`. No rejection tombstone needed: message evidence can no longer re-add a removed address. One carve-out: the opt-in `auto_confirm_identities` provider-inventory refresh can still insert (the provider is authoritative for its own aliases); can gate that behind apply too as a follow-up if you prefer.

**3. Resumability** — Per-page discovery gets bounded retry (3x, exponential backoff, error retained via `%w`); on persistent failure the run logs the error, parks a durable per-source backlog marker, and completes/checkpoints normally. The marker drains at the next sync start (including no-op incrementals, so wound-down mailboxes settle their debt) and at end of run. Two notes: acknowledgment no longer waits on discovery (pages ack once the debt is recorded), and a source removed from scheduling holds its marker until a manual `identity discover`.

**4. Refresh cost** — Insert-bearing chunks now refresh only the affected senders' messages (participant-scoped, same matching rules as attribution); unseen aliases skip the refresh. Added `LOWER()` expression indexes on `participants.email_address` and `participant_identifiers.identifier_value` in both schemas.

Smaller items: Fastmail errors (method/host/status; redaction-safe, test-pinned) now reach internal logs at both former drop sites with token-absence assertions restored; `FastmailSourceFor` store failures are 500 while config errors stay 400 (raw error text no longer leaks into the 400 body); no JMAP call on no-op incrementals; `identity import` gets the same 30-min protective ceiling as the other mutations; explore filter resolution mirrors attribution rules exactly (per-type matching, `phone_number` column dropped); the identity filter stays removable on catalog failure and shows unlisted restored identities as explicitly unavailable; redundant SQLite composite and unused label-role index removed (net-zero vs main); the PG one-time index build is now `CREATE INDEX CONCURRENTLY` on an autocommit connection with INVALID-leftover recovery and a connection-scoped timeout reset.

salmonumbrella:
Roborev's envelope-address finding is fixed in c2804bf2 (squashed into the review-response commit): recipient records now preserve the envelope address at ingest, and discovery evidence reads it in preference to the participant's current email, so participant merges can no longer repoint one address's sent evidence onto the merge survivor. Regression tests cover discovery after merging participants with different primary emails, through the real merge and persist paths. One accuracy note: rows ingested before this column existed fall back to the participant's current email until their message is re-ingested — routine re-sync skips already-archived messages, so a follow-up repair command backfilling from stored raw MIME would be the way to converge old archives.

salmonumbrella:
Both round-2 findings addressed in 92a7fa95.

**Identity filtering across aliases** — Fixed as suggested: the analytical predicate carries the confirmed identifier and compares email identities against `message_recipients.email_address` envelope-first, with participant fallback only for legacy NULL rows; badge hydration follows the same rule. Non-email identifiers keep participant matching (they have no envelope snapshot, and it preserves the attribution-parity rules requested in the original review). The Parquet cache now exports the envelope column, gated by a schema-version bump (16→17) that forces a full rebuild so pre-upgrade shards degrade to participant fallback instead of mixing. Regression tests cover post-merge filtering, direction scoping, badges, confirmed-but-unseen identities, and legacy caches without the column.

**Fastmail inventory vs JMAP limits** — The suggested chunked enumeration isn't implementable: the MaskedEmail extension defines only `/get` and `/set` (no `/query` or `/changes`), so IDs can't be enumerated without fetching, and the relevant JMAP error is `requestTooLarge` (RFC 8620 §3.6.2) rather than `tooManyObjects`. What's implemented instead: the client parses the session's advertised `maxObjectsInGet` and classifies a `requestTooLarge` on `MaskedEmail/get`/`Identity/get` as a typed object-limit error, so discovery and auto-refresh fail with a diagnosable message in logs and as a typed error (public payloads stay generic; token/body redaction unchanged). Fastmail's observed session limit is 4096 objects, so only accounts beyond that hit this boundary, and they now get a clear failure instead of an opaque one.

salmonumbrella:
Addressed both findings from the `c410dc4` review in 1cc341f:

- **`internal/store/store.go` ctx** — `InitSchemaContext` now passes its `ctx` into `buildLargeIndexesConcurrently`, so cancellation interrupts the concurrent index build (an aborted build leaves at worst an INVALID leftover that the next start already drops and rebuilds). The deferred `RESET statement_timeout` runs on a bounded 10s context detached from cancellation.

- **Envelope alias snapshots** — `message_recipients` uniqueness now spans the normalized envelope address (`idx_message_recipients_envelope`), replacing the table-level `UNIQUE(message_id, participant_id, recipient_type)` via a run-once migration (PostgreSQL drops the catalog-discovered constraint; SQLite rebuilds the table since a table-level UNIQUE lives in an undroppable autoindex). `buildRecipientSet` and `replaceMessageRecipientsTx` dedupe by participant + normalized envelope address, and `MergeParticipants` only deletes rows whose full unique key collides — distinct alias snapshots are repointed onto the survivor instead of destroyed, so identity discovery keeps observing them. Covered by new SQLite and PostgreSQL migration tests plus merge/discovery tests.

salmonumbrella:
Addressed the `1cc341f` review finding in 1d8797d:

- **`messageIdentityAttributionMatch`** now also matches the message's own `from` envelope snapshot (`message_recipients.email_address`) against `account_identities`, so every attribution refresh — identity add/remove, merges, batch confirmation — repairs messages whose confirmed alias survives only in the envelope after a participant merge.
- **Batch confirmation scoping** (`participantIDsForConfirmationsContext`) additionally resolves sender IDs of the confirming source's messages whose `from` envelope matches a confirmed email address (one source walk per chunk, probing the envelope through the message-id index). Previously a merged-away alias resolved to zero participants and the refresh never ran.
- **Cache export** (`build_cache.go`) gains the same envelope clause in its `is_from_me` derivation (both the partitioned export and the empty-shard schema stub), guarded for snapshots that predate the envelope column.

Regression tests: a non-source-native sent message confirmed after its alias participant was merged, covered via `AddAccountIdentitiesBatchContext`, `AddAccountIdentity`, and the Parquet export on both the sqlite-scanner and CSV snapshot paths. Store suites pass on SQLite and PostgreSQL.

salmonumbrella:
Addressed the `1d8797d` review finding in ee5d71f:

- **`persistMessageWith` now settles attribution after recipients are final.** The message upsert's attribution CTE runs before the transaction writes the `from` envelope snapshot, so an envelope-only identity persisted as `identity_is_from_me=false` and a re-persist could clear previously repaired attribution. A guarded single-row recompute (reusing `messageIdentityAttributionMatch`) now runs in the same transaction after `replaceMessageRecipientsTx` — unconditionally, since even a persist that touches no recipients re-ran the envelope-blind CTE and must be settled against the rows already in the table. The change guard keeps the common agreeing case write-free (no `last_modified` trigger fires).
- Standalone `UpsertMessage` callers were audited: all are non-email importers whose recipient rows carry no envelope addresses, so the upsert-time CTE is unchanged.

Regression tests cover both cases the review asked for — initial persistence of an envelope-only identity and re-persistence of a repaired merged-alias message (same `source_message_id`, sender resolved to the merge survivor). Both fail without the recompute; store suites pass on SQLite and PostgreSQL.

salmonumbrella:
Addressed the `ee5d71f` review finding in 3606dc4:

- **`MatchMessageIdentitiesContext` now applies the same comparison rules as `messageIdentityAttributionMatch`.** The candidate query selects `pi.identifier_type`, and candidates are kept in one set per rule, intersected against a per-source identity index keyed the same way: email-typed identifier rows compare case-insensitively (`LOWER = LOWER`), every other identifier type compares verbatim, and envelope/participant-email candidates keep the existing shape rule (preserving the pinned synthetic-address exactness).
- **`participants.phone_number` is no longer a candidate** — attribution never consults it (`EnsureParticipantByPhone` always backs a phone identity with a `participant_identifiers` row), so badging through it kept matching a phone identity after its identifier mapping was removed.

Parity tests mirror the existing `ResolveAccountIdentityContext` guards for the hydration path: the phone badge vanishes when the identifier row is deleted (the merge edge), and an email-shaped `matrix` identifier badges only its exact casing. Both fail against the old candidate logic; a third test locks in case-insensitive email-typed badging. Store/sync/api suites pass on SQLite and the store suite on PostgreSQL.

salmonumbrella:
Addressed the `3606dc4` review finding in 8794be7:

- **`CopySubset` no longer copies labels positionally.** The schema-derived column-list copy that `copyMessages` already used is generalized into `copyByName`, and the labels copy goes through it — the INSERT names the columns the source and destination schemas have in common, so an upgraded source (where the `ADD COLUMN` migration appended `system_role` after `color`) can no longer swap the two fields and lose the `sent` role.
- **`participants` and `conversations` had the same positional hazard and are routed through the same helper** — their late-added columns (`phone_number`/`canonical_id`, `title`/`conversation_type`) also sit mid-table in a fresh schema. `message_recipients` is unaffected: its late column (`email_address`) is declared last in the fresh schema, and the envelope rebuild migration recreates the table in that order.

The regression test rebuilds all three tables in the upgraded (appended-at-the-end) column order, seeds a label with `system_role = 'sent'` plus a color, copies into a fresh subset, and asserts every value lands in its own column; it fails without the fix with `system_role` and `color` swapped. The store suite passes on SQLite; subset creation is SQLite-only (hardcoded `sqlite3` driver, `ATTACH DATABASE`, `pragma_table_info`), so the PostgreSQL suite is unaffected.

wesm:
rebasing

wesm:
fixing and rebasing

PR #544, opened 2026-08-03
fix(beeper): discover accounts missing from the accounts API

### LINKED ISSUE #543: add-beeper never registers networks Beeper serves natively (iMessage): they are absent from /v1/accounts
## What I was doing

Setting up `add-beeper` on a machine whose Beeper Desktop carries iMessage
alongside the usual bridged networks, then checking which sources it had
registered.

## What happened

`add-beeper` succeeded, registered a source per bridged network, and never
mentioned iMessage. No error, no warning — the network is simply absent from
`list-accounts` afterwards, and `sync-beeper` therefore never archives a single
message from it.

Beeper Desktop does not report every account from its accounts endpoint.
Networks it serves natively rather than through a Matrix bridge are missing
there, while every other endpoint serves them normally. Against Beeper Desktop
4.3.0 with iMessage connected and syncing:

- `GET /v1/accounts` returns the bridged accounts only, with no iMessage entry.
- `GET /v1/accounts/{imessage-account-id}` returns
  `{"message":"Account not found","code":"NOT_FOUND"}`, so this is not a
  list-filter quirk — the account is absent from the whole Accounts surface.
- `GET /v1/chats/search?limit=200&type=any&accountIDs={imessage-account-id}`
  returns a full page of chats with `hasMore: true` and a valid `oldestCursor`.
- `GET /v1/chats/{id}?maxParticipantCount=-1` returns `network: "iMessage"`,
  `type`, `lastActivity`, and a participant list including the `isSelf: true`
  member with its `email` / `phoneNumber`.
- `GET /v1/chats/{id}/messages` returns ordinary message objects: `id`,
  `chatID`, `accountID`, `senderID`, `senderName`, `timestamp`, `sortKey`,
  `text`, `isSender`.
- `GET /v1/messages/search` returns iMessage hits interleaved with hits from
  bridged networks.

Beeper's own local index describes the account with `bridgeProvider:
"platform-sdk"` rather than a Matrix bridge, which is consistent with it being
absent from an endpoint that enumerates bridge accounts.

`add_beeper.go` calls `beeperClient(token).ListAccounts()` and creates one
source per returned account; `internal/beeper/client.go`'s `ListAccounts` is a
thin `getJSON("/v1/accounts", &out)`. That single enumeration is the whole
defect.

## Why it matters

Nothing else has to change for these accounts to work. `internal/beeper`
already pages their chats and messages, resolves participants, downloads media,
and stores cursors and anchors exactly as it does for bridged networks — the
endpoints it calls all serve them. `resolveBeeperSyncAccounts` iterates
registered sources and never re-queries Beeper. The only thing standing between
msgvault and a live iMessage archive is that no source row is ever created.

The practical effect is that iMessage is the one network Beeper carries that
msgvault cannot reach at all. The fallback is `import-imessage`, a one-time
local import against `chat.db` that needs a Mac with Full Disk Access, instead
of the scheduled sync every other network gets.

It is also not iMessage-specific: any account Beeper adds under the same
native-account mechanism will be invisible to `add-beeper` in the same way.

## Fix

Stop treating the accounts endpoint as the only source of truth for discovery.
Union it with the distinct `accountID` values observed in a bounded, unfiltered
`/v1/chats/search` sweep, and synthesise an account for any ID that endpoint did
not report — taking the network name from the chat and the account owner's
identity from the chat's `isSelf` participant. That fixes the general case
without special-casing any network, keeps discovery in `add-beeper` rather than
on the sync path, and degrades to today's behaviour if the sweep fails.

Implemented in the linked PR.

salmonumbrella:
## What changed

- discover Beeper accounts by unioning `GET /v1/accounts` with the distinct account IDs seen in a bounded, unfiltered chat search, synthesising an account for each ID the endpoint did not report
- resolve a synthesised account's network name and own identity from chat data, merging the `isSelf` member's identity fields across a few of its chats
- point `add-beeper` at the new discovery and mark such sources as *found via chats* in its output
- correct help text and docs that described Beeper as bridging every network

## Root cause

Beeper Desktop serves some networks natively rather than through a Matrix bridge and leaves those accounts out of its accounts endpoint — `GET /v1/accounts/{id}` answers 404 for them too — while `/v1/chats/search`, `/v1/chats/{id}`, `/v1/chats/{id}/messages` and `/v1/messages/search` all serve them normally. `add-beeper` enumerated only that endpoint, so no source was ever created for such a network and nothing downstream ran for it. iMessage is the case that matters today. Details and endpoint-by-endpoint evidence in #543.

Nothing keys off a network name or an ID prefix: the defect is "absent from `/v1/accounts`", so any account Beeper adds under the same mechanism is picked up too, and if a later release starts listing these accounts they are found in the endpoint's own result and the sweep skips them.

## Impact

Re-running `msgvault add-beeper` on an existing install registers the missing networks and leaves existing sources untouched; `sync-beeper` then backfills and syncs them incrementally with the same cursors, anchors, resume behaviour and `[beeper]` include/exclude filtering as every other Beeper source, keeping `message_type = beeper` with the network carried by the source. No configuration or schema changes are required.

The sweep is bounded at five pages and runs only where sources are created, never on the sync path or in the token pre-flight. It is additive: a failed chat search still registers whatever the accounts endpoint returned, and both early exits are logged so a stopped sweep is distinguishable from a network that genuinely has no chats. Beeper carries iMessage on macOS only, so the daemon still has to run beside a Mac running Beeper Desktop.

Two things I left alone deliberately: only one identifier is persisted for a discovered account, matching what `add-account` and `add-imap` do — collecting several and writing them all is a worthwhile follow-up but is not specific to this change; and attachment handling is shared code with no network-conditional branches, so media should behave identically here, though I have not verified an iMessage attachment end to end.

Closes #543

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ export formats.
 - **Microsoft Teams sync**: archive delegated Graph chats, channels, replies, and inline media with `message_type = teams`
 - **Discord sync**: archive guild channels, threads, forum posts, and attachments through a read-only bot with `message_type = discord`
 - **Meeting notes**: sync Granola and Circleback notes and transcripts, then browse them in the TUI
-- **Beeper Desktop sync**: archive chats and media from every network bridged through Beeper's local API
+- **Beeper Desktop sync**: archive chats and media from every network connected to Beeper, including iMessage, through its local API
 - **IMAP sync**: archive mail from any standard IMAP server
 - **Incremental backup snapshots**: verifiable `msgvault backup` repositories for the SQLite archive and attachments
 - **MBOX / Apple Mail / PST import**: import email from local export formats
--- a/cmd/msgvault/cmd/add_beeper.go
+++ b/cmd/msgvault/cmd/add_beeper.go
@@ -29,6 +29,10 @@ Instagram, LinkedIn, X, Facebook, ...) into one local app with a read-only
 API on localhost. Each connected network account becomes its own msgvault
 source, so networks stay separately filterable and searchable.
 
+Accounts Beeper serves natively rather than through a bridge (iMessage) are
+missing from its accounts API, so they are found from chat data instead. Re-run
+this command after connecting a network in Beeper to register it.
+
 Requires Beeper Desktop to be running on this machine. Mint an access token
 in Beeper Desktop (Settings > Developer) and provide it via --token-file,
 the MSGVAULT_BEEPER_TOKEN environment variable, or the interactive prompt.
@@ -55,6 +59,9 @@ Examples:
 				if IsRemoteMode() {
 					_, _ = fmt.Fprintln(cmd.OutOrStdout(), "Remote daemon configured: it must run beside its own Beeper Desktop, which will validate the token.")
 				} else {
+					// A token check, not discovery: the daemon subprocess below
+					// does the real enumeration, so this stays the single
+					// cheapest authenticated request.
 					accounts, err := beeperClient(token).ListAccounts(cmd.Context())
 					if err != nil {
 						return err
@@ -70,7 +77,7 @@ Examples:
 			if token == "" {
 				return errors.New("missing Beeper token in daemon subprocess (set MSGVAULT_BEEPER_TOKEN)")
 			}
-			accounts, err := beeperClient(token).ListAccounts(cmd.Context())
+			accounts, err := beeperClient(token).DiscoverAccounts(cmd.Context())
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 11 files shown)


salmonumbrella:
Good catch — fixed in `c314b7a5`.

The sweep already refused to treat cancellation as a tolerable probe failure, but `resolveCandidate` still did: it swallows per-chat fetch errors so an account with unreadable chats is registered anyway, and a cancelled context went down that same path. `add-beeper` would then save the token and create sources whose identity had never actually been looked up.

Both loops now check `ctx.Err()` and return it, so cancellation fails the command instead. `resolveCandidate` returns `(Account, error)` to make that explicit rather than relying on a post-hoc check. The cancellation test is now table-driven over both points — cancelling on `/v1/chats/search` and on the chat-detail probe — and asserts `context.Canceled` reaches the caller either way.

PR #545, opened 2026-08-04
feat(api): add GET /api/v1/messages/changes content-change feed

danshapiro:
## What this adds

- Adds `GET /api/v1/messages/changes` so consumers can refresh changed messages without scanning the full archive.
- Returns current message snapshots ordered by `(content_changed_at, id)`; multiple edits between polls may collapse into one result.
- Uses an opaque, archive-bound cursor that clients store and return unchanged.
- Reports `server_time` and nullable `complete_through` so clients can distinguish “caught up” from a feed held back by an open write.
- Includes deduplication-hidden and source-deleted messages so consumers can update their visibility state.

## Storage and API changes

- Adds, backfills, and indexes `messages.content_changed_at`.
- Adds triggers that move the watermark only when tracked message content or bodies actually change.
- Installs trigger definitions through a versioned migration instead of recreating them at every startup.
- Makes schema initialization and archive identity lookup context-aware so cancellation reaches database waits.
- Rejects corrupt watermarks, treats legacy `NULL` attachment flags as `false`, and exposes `complete_through` as a nullable date-time.
- Bumps the API schema from 1.33.0 to 1.34.0 and regenerates the Go and TypeScript clients.

## Safety and operational notes

- In keyless loopback mode, browser requests to the feed must be same-origin. Cross-origin clients can use an API key.
- Every caller is limited per client IP to 2 feed requests per second with a burst of 4. Clients should honor `429 Retry-After`.
- This is an invalidation feed, not an audit log. It reports that a row changed, not every individual edit.
- Labels, recipients, attachment metadata/content, raw MIME, body deletion, participant edits, conversation metadata, and hard deletions do not move the watermark.
- Existing archives upgrade on first open. The backfill resumes after interruption; the index build can temporarily block writes, especially on PostgreSQL.
- SQLite briefly takes the writer lock to establish `complete_through`, so consumers should poll on an interval instead of in a tight loop.
- A backward database clock or restoring an older snapshot can still make a consumer diverge; solving those cases requires a durable change ledger outside this PR.


wesm:
Looking at this

wesm:
Addressed in 0e7ae8cd. The feed query now selects `COALESCE(has_attachments, FALSE)`, so legacy nullable rows decode as `false` instead of returning HTTP 500. `TestListChangedMessages_NullHasAttachmentsDefaultsFalse` covers the regression through the shared store test harness used by both SQLite and PostgreSQL CI.

wesm:
Addressed in `a5c8d22b`. Keyless loopback change-feed requests now reject mismatched `Origin` values and `Sec-Fetch-Site` values other than `same-origin` or `none` before reaching storage. The endpoint also has a dedicated per-client-IP limit of 2 requests/second with a burst of 4, with no loopback or API-key bypass. Same-origin browser access, API-key cross-origin access, and headerless CLI clients remain supported. Regression coverage exercises each path and confirms rejected requests never reach the lock-taking store method.

wesm:
thanks @danshapiro!

danshapiro:
Had words with the bots about the quality of the last commit. ;)

PR #548, opened 2026-08-05
fix(daemon): prevent deletion of live daemons' runtime records

danshapiro:
## Summary

Fixes a critical bug in daemon lifecycle management where a daemon's runtime record could be permanently deleted by CLI-side housekeeping operations, breaking subsequent `msgvault backup create` runs with a false "a local msgvault write operation is in progress" error until the daemon restarts. This fix uses three layered protections:

1. **Tolerate create-time skew in liveness validation** — Container clock quirks can cause false "dead daemon" positives; the fix implements a three-state comparison with ±2s tolerance so transient time discrepancies don't trigger deletion.

2. **Probe-corroborated identity check before any deletion** — Before pruning a record, verify identity via HTTP probe. If uncertain (probe fails, container unreachable), leave the record alone rather than risking deletion of a live daemon.

3. **Automatic self-healing via periodic republish** — The `serve` command now republishes its runtime record every 30s, so even if a record is wrongfully pruned, it self-heals within one heartbeat interval.

## Commits included

- `bf5b5938` fix(daemon): tolerate create-time skew when validating runtime records
- `05655a8f` fix(daemon): stop deleting live daemons' runtime records on identity doubt
- `b4075805` docs(daemon): document runtimeRecordIdentityMismatched's never-delete contract
- `c580a576` feat(serve): republish the daemon runtime record when it goes missing
- `0cbb8430` docs: describe daemon runtime record self-healing
- `e442f2e5` fix(serve): wrap runtime-store path error to satisfy wrapcheck

Excluded two internal planning docs (docs/plans/) that were intermediate development notes.


wesm:
good catch, looking at this now

PR #552, opened 2026-08-06
Add loose-only attachment storage mode

MSch:
Add `[data].loose_attachments` for installations that need attachments to remain as individual files.

My use case is an attachment directory being backed up by file-based backup software. Packing and repacking creates filesystem churn that causes otherwise unchanged attachment data to be backed up again, increasing storage costs with the cloud backup provider. Since most attachments are already compressed, the resulting space savings do not justify that additional backup activity.

When enabled, msgvault skips automatic pack and repack maintenance, rejects explicit `pack-attachments` and `repack-attachments` commands, and restores backup attachments as loose files.

Existing packs are not converted automatically. Operators can stop the daemon and run `msgvault unpack-attachments` once to migrate existing packed content.

### DIFF
--- a/cmd/msgvault/cmd/attachment_maintenance.go
+++ b/cmd/msgvault/cmd/attachment_maintenance.go
@@ -25,17 +25,19 @@ const (
 // callers already hold the daemon operation gate; these methods must not try
 // to acquire it again.
 type attachmentMaintenance struct {
-	store       *store.Store
-	maintainer  *packstore.Maintainer
-	coordinator *packstore.Coordinator
-	blob        *attachmentstore.Store
-	logger      *slog.Logger
+	store               *store.Store
+	maintainer          *packstore.Maintainer
+	coordinator         *packstore.Coordinator
+	blob                *attachmentstore.Store
+	logger              *slog.Logger
+	packCreationEnabled bool
 }
 
 func newAttachmentMaintenance(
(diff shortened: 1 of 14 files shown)


wesm:
Not a bad idea per se, but have you thought about using `msgvault backup`?

wesm:
rebasing

MSch:
> Not a bad idea per se, but have you thought about using msgvault backup?

If I read the docs correctly, that
1. duplicates the required storage, because I have both the actual msgvault storage (using packfiles or whatever), and the backup repository where these attachments then get copied to a second time.
2. doesn't really give me any advantage over filesystem snapshots. I already run on a CoW filesystem, and I use Kopia for backups, so I can just take a snapshot before, do the backup, delete the snapshot. Right now I don't intend to keep the msgvault daemon around, but if that changes I'd prefer [an approach like Postgres](https://www.postgresql.org/docs/8.1/backup-online.html#BACKUP-BASE-BACKUP) where backup tools can register that they're doing a backup, and so Postgres doesn't truncate its WAL until they're done, over explicit "make a backup in this folder". Pull not push so to say.

I also don't want to have to add regular verification of msgvault's backup code to my backup routine. I already trust&verify Kopia, and having to also make sure msgvault can keep reading its own backup, doesn't silently corrupt it, etc adds complexity.

PR #553, opened 2026-08-07
perf(test): cut PostgreSQL test wall clock by pre-warming schemas and not running the suite twice

danshapiro:
## Summary

- Pre-warms private PostgreSQL test schemas in small synchronous batches, reducing fixture setup time without sharing state or adding background database work.
- Scopes migration and backfill test hooks to each `Store`, preventing concurrent fixtures from triggering hooks installed by another test.
- Detects full-text search support when reopening read-write stores, avoiding an unnecessary fallback to slower search behavior.
- Updates `make test-pg-both` to run the full pgvector suite once, then run the shipped-build variant only for packages whose test binaries differ. A dependency check keeps that package list from drifting.
- Preserves test coverage: no tests are skipped, shortened, shared, or newly parallelized.
- Improves measured PostgreSQL test time: `internal/store` fell from 350s to 256s (27%), and the shipped-build lane fell from 881s to 305s on the development host.

## Configuration

- Set `MSGVAULT_TEST_PG_WARM_POOL_BATCH` to lower the default batch size of 4.
- Set `MSGVAULT_TEST_PG_WARM_POOL=0` to disable schema pre-warming and use the previous fixture behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -526,10 +526,14 @@ jobs:
       # This lane uses a stock postgres:16 image (no pgvector extension), so it
       # must NOT build the pgvector tag — those tests CREATE EXTENSION "vector"
       # and fail here. pgvector is covered by the dedicated test-pgvector job
-      # against a pgvector/pgvector image. (Run go test directly rather than
-      # `make test-pg`, whose PG_TEST_TAGS include pgvector.)
+      # against a pgvector/pgvector image. That is why this runs test-pg-shipped
+      # (BUILD_TAGS) and not test-pg (PG_TEST_TAGS, which include pgvector).
+      #
+      # The target carries -timeout 20m for the same reason `make test` does:
+      # the heavy DuckDB packages can exceed go test's 10m per-package default
+      # on a contended runner, and this lane is the slowest one there is.
       - name: Test (PostgreSQL)
-        run: go test -tags "fts5 sqlite_vec" ./...
+        run: make test-pg-shipped
 
       - name: Concurrency tests (-count=5)
         run: |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,8 +165,20 @@ See `docs/internal/PG_STATUS.md` for the current implementation state and
 follow-up work.
 
 **Test env**: `MSGVAULT_TEST_DB=postgres://...` runs PostgreSQL-backed
-tests (`make test-pg`). pgvector tests require a PostgreSQL instance
-with the `vector` extension and the `pgvector` build tag.
+tests. pgvector tests require a PostgreSQL instance with the `vector`
+extension and the `pgvector` build tag.
+
+There are two PostgreSQL configurations to cover: the pgvector build
+(`make test-pg`) and the shipped build, which has no pgvector tag
+(`make test-pg-shipped`). Run `make test-pg-both` rather than both of those —
+the tag changes the test binary of only nine packages, so the second full
+run would reprove the first. Fixtures build their schemas in concurrent
+batches (`internal/testutil/pg_warm_pool.go`) rather than one at a time: a
+fixture that finds the buffer empty builds several and the next few claim
+theirs for free. All of that happens inside the claiming fixture's own setup —
+the pool runs no background work, so it can never issue a statement while a
+test body is running. Each schema is still private to its test, still built by
+the same `InitSchema()` path, and still dropped on cleanup.
 
 ## Parquet Analytics
 
@@ -284,7 +296,9 @@ automatically:
 ```bash
 make install-hooks             # Install pre-commit hook via prek
 make test                      # Run tests (SQLite default)
-make test-pg                   # Run PostgreSQL-backed tests with MSGVAULT_TEST_DB set
+make test-pg-both              # Both PostgreSQL configurations, needs MSGVAULT_TEST_DB
+make test-pg                   # PostgreSQL, pgvector build only
+make test-pg-shipped           # PostgreSQL, shipped build only
 make fmt                       # Format code (go fmt)
 make lint                      # Run linter (auto-fix)
 make lint-ci                   # Run linter (CI, no auto-fix)
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,17 @@ BUILD_TAGS := fts5 sqlite_vec
 # Omitting sqlite_vec compiles those out and the target gives false confidence.
 PG_TEST_TAGS := fts5 sqlite_vec pgvector
 
+# The only packages that build a different test binary under BUILD_TAGS than
+# under PG_TEST_TAGS. That is not just the packages carrying pgvector-gated
+# files: a package whose own sources are identical still links different code
+# when something in its dependency closure changed, so this is the reverse
+# dependency closure of the tag-sensitive packages, not the tag-sensitive
+# packages themselves. Every package outside this set compiles byte-identically
+# in both configurations, so test-pg-both runs just these in the shipped-build
+# configuration. Verified by `make pg-shipped-only-check`, which re-derives the
+# closure from `go list`.
+PG_SHIPPED_ONLY_PKGS := ./cmd/msgvault ./cmd/msgvault/cmd ./internal/api ./internal/mcp ./internal/scheduler ./internal/vector/chunkmatch ./internal/vector/embed ./internal/vector/hybrid ./internal/vector/pgvector
+
 OPENAPI_ARTIFACTS := api/openapi.yaml pkg/client/openapi.yaml pkg/client/generated
 WEB_INSTALL_STAMP := web/node_modules/.msgvault-install-stamp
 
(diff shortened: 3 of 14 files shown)


danshapiro:
Good catch — that's a real bug, and I've fixed it in `5260dca` (folded into the warm-pool commit rather than stacked on top, so the series stays bisect-safe).

You're right that the PID check is only meaningful inside one namespace, and the consequence is worse than a lost speedup: a warm schema keeps its name for the whole life of the test that claims it, so the schema a foreign sweeper would drop is a **live** one, not a spare. The test loses its tables mid-run.

I went with your "never sweep foreign owners" option rather than the advisory lock. A lease needs a holder, and warm schemas sit unclaimed in the buffer with no session attached to them — that's exactly the window the sweep exists to clean up, so the lock would have to be taken by the creating worker and then deliberately outlive its own connection close.

Each name now carries an owner token derived from the boot id and the PID namespace (`msgvault_warm_<owner>_p<pid>_<suffix>`). The sweep skips any candidate whose token isn't its own, so it only ever judges liveness for PIDs its own `/proc` can resolve. Foreign orphans are left for the namespace that created them to reclaim — a leaked schema costs a name until that binary runs again; a wrong verdict corrupts a running test.

Covered by `TestSweepWarmSchemasNeverDropsForeignNamespaces`, which builds a foreign-owned schema whose PID this process genuinely cannot resolve and asserts it survives, alongside an own-namespace orphan that is still reclaimed so the assertion can't pass vacuously. Verified red before green: with the token guard removed, the foreign schema does land in the dropped list.

danshapiro:
All three are real. The second one turned out to be the important one — it is not a rough edge on the design, it is an argument against the design, and I have changed the design rather than worked around it.

## 2. Background migration interferes with log-sensitive tests

CI proved this before I could: `test-postgres` failed on
`TestIdentityDiscoveryDoesNotReadContentTables`, which captures `slog.Default()`
and asserts the whole buffer never mentions `message_raw`. A pool worker's
`InitSchema` had written that table underneath it.

My first fix was to scope that one assertion by `request_id`. That was wrong,
and a second full-package run showed why: a *different* upstream test,
`TestGenericAPIMessagePathsDoNotHydrateIdentityMatches`, failed the same way on
`account_identities`. Six files in `internal/store` replace the default logger,
across twelve call sites, and I have no way to prove I have found the last
victim — the two I know about failed in different runs, not the same one. A
speedup that requires editing other people's assertions to accommodate its own
noise is not a speedup worth having.

Suppressing the pool's logging is not reachable either: SQL logging goes through
package-level `slog` calls with process-wide atomic configuration, so a worker
cannot quiet itself without a per-`Store` logger — a production change to make a
test helper convenient.

So the workers are gone. A fixture that finds the buffer empty now builds
`warmPoolBatch` schemas concurrently, synchronously, in its own setup, and the
fixtures after it claim theirs for free. The invariant that buys is stronger
than any per-test patch: **the pool's statements execute in exactly the window
where that fixture's own `CREATE SCHEMA` and `InitSchema` already executed**, so
a test body cannot observe anything it would not have observed without the pool.
No background goroutine exists to overlap anything.

Both failing tests are untouched and pass. `TestWarmPoolIssuesNoStatementsOnceATestBodyIsRunning` pins the invariant: it drains the buffer so the fixture
really does trigger a refill, captures the process logger with `FullTrace` on
after setup, and asserts no DDL appears. Verified red first — reinstating a
single background `go p.warmOne()` makes it fail.

The cost is the overlap-with-test-body half of the win; the concurrency half is
what mattered. Re-measured, same host, same container, toggled only by
`MSGVAULT_TEST_PG_WARM_POOL`: `internal/store` went 350 s with the pool off to 256 s with it on, and 261 s on a second pass — all green, and both previously-failing tests untouched.

## 1. Non-Linux warm-pool schemas cannot be reclaimed

Correct, and the random-token fallback was the bug. On a host with no `/proc`
the token was unique per process *and* `processAlive` always answers "alive", so
nothing could ever be swept — every run stranded its buffer permanently.

`deriveWarmOwner` now reports unusable when neither the boot id nor the pid
namespace is readable, and `claimWarmSchema` declines to run at all in that
case. Those platforms lose the speedup and keep the exact behaviour they had
before this PR. Creating garbage you have no mechanism to collect is worse than
being slow.

`warmOwnerFor` is split out so the rule is testable without a host that lacks
`/proc`.

## 3. Connection limits enforced only per test binary

Right, and I am not claiming to have solved it. The bound is per binary, and
`go test ./...` runs up to `-p` binaries at once; a real fix needs a
cross-process budget, which is a bigger thing than this PR should carry.

What I have done instead: the commit message no longer implies the aggregate is
bounded, the constant's comment states the per-binary limitation explicitly and
names both levers (`MSGVAULT_TEST_PG_WARM_POOL_BATCH`, and `-p`), and the knob
can only ever *lower* the batch — a limit an environment can raise is not a
limit. Worth noting the failure mode is not silent: a refill that cannot connect
gives up after `warmPoolMaxFailures` and fixtures create their own schemas,
which is the pre-PR behaviour.

The synchronous batch also narrows the exposure in practice. Connections are now
held only during a refill inside fixture setup, not by four workers looping for
the life of the binary.

danshapiro:
Both correct, and the second one was already failing CI on Windows by the time this review landed — `TestParseWarmSchemaName/accepts_a_name_this_package_generated`, exactly as described. Fixed in `a402a3db`, folded into the warm-pool commit.

## `deriveWarmOwner` accepted either identifier

Right, and the `||` put the original bug back on any host where only one of the
two is readable. A boot id is common to every container on a machine; a
namespace inode number is common to every machine that numbers its namespaces
alike — the initial namespace has the same well-known id everywhere. Either
alone lets two processes that cannot see each other's `/proc` agree on a token,
which is the precise condition the token exists to rule out.

`warmOwnerFrom(bootID, namespace)` now requires both and reports false
otherwise. `TestWarmOwnerRequiresBothNamespaceIdentifiers` covers each single-value
case separately, plus the two collisions that motivate the rule: same host with
different namespaces, and same namespace id on different hosts, must not share a
token.

## The parser test assumed a host with `/proc`

This one I shipped as a regression: removing the random-token fallback (your
previous review's finding 1) left `warmOwnerToken` empty where the pool is off,
and that test was building its fixture from it.

Split in two. The round-trip property is now asserted against a synthetic owner
token, so it means the same thing on every platform. The generated-name subtest
keeps its own assertion and branches on `warmOwnerUsable` rather than skipping —
where there is no owner token it asserts the name is *not* sweepable, because an
ownerless name reaching the sweeper would itself be a bug. Verified against a
build with `deriveWarmOwner` forced to report unusable, which is the Windows and
macOS shape.

Every other test in the file is gated behind a PostgreSQL URL, which is why this
was the only one those lanes reached.

danshapiro:
Both halves are right, and I took the second option you offered — inject synthetic owner state — rather than skipping. Fixed in `7c11f42c`, folded into the warm-pool commit.

## The sweep tests read global ownership

`sweepWarmSchemas` now takes the owner token as a parameter alongside the
liveness function it already took, and returns nothing for an empty owner. That
removes the platform dependence rather than guarding it:

- The two cross-namespace safety tests now stage schemas under synthetic owner
  tokens and inject liveness, so neither needs a `/proc` for a pid to be absent
  from. Both previously carried a `runtime.GOOS != "linux"` skip; both now run
  everywhere. That is a net gain in coverage, not a workaround for the report.
- Synthetic tokens also stop a test competing with a concurrent run of the same
  binary for the same schemas, which the real token made possible on a shared
  server.
- `TestSweepWarmSchemasDropsOnlyDeadOwners` stays on the real token and the real
  `processAlive` — that pairing needs one wired-up test. Where there is no owner
  token it asserts the sweep is a no-op instead of passing over the platform,
  since a sweep that still ran there would be judging pids it cannot resolve.

## The claim test reached past the production gate

Correct, and the consequence was worse than a failed assertion: calling
`warmPoolFor(...).claim()` directly would mint `msgvault_warm__p<pid>_<suffix>`
schemas that `dropOwnedSchema` cannot parse, so the test leaked precisely the
garbage the ownership rule exists to prevent.

It now goes through `claimWarmSchema`, which is where the gate lives, and
asserts both outcomes: a schema where the pool runs, and a refusal where it
cannot. Its cleanup uses `dropOwnedSchema` rather than raw SQL, so it exercises
the same validated path production uses.

Verified by forcing `deriveWarmOwner` to report unusable and running the whole
package against PostgreSQL — the host-with-no-`/proc`-but-a-PG-URL combination
you identified, which did fail before this change and passes now. Also green
unforced, and `internal/store` is green on the same tree.

danshapiro:
Correct, and this one was a coverage hole rather than a gate weakness — the shipped lane really was skipping tests it should have run. Fixed in `a1fa35c2`, folded into the lanes commit.

Measured on this tree: the tag changes **9** packages' test binaries, not 4.
Only 4 carry pgvector-gated files. The other 5 —
`./cmd/msgvault`, `./internal/api`, `./internal/mcp`, `./internal/scheduler`,
`./internal/vector/chunkmatch` — have byte-identical sources under both tag sets
but link one of those 4, so their test binaries differ anyway. Comparing
per-package source lists could not see them, exactly as you describe, and
`test-pg-both` was dropping all five from the shipped-build configuration.

`pg-shipped-only-check` now runs in two steps: the `go list -deps -test` diff
under both tag sets to find what the tag touches directly, then a walk of the
test dependency graph (`go list -test -f '{{.ImportPath}}|{{join .Deps " "}}'`)
adding every package that links one of them. `PG_SHIPPED_ONLY_PKGS` is now the
closure, and the check passes against it. Cost went from ~1.5 s to ~2 s.

Verified to fail red on a stale list where the *only* missing entry is a reverse
dependent (`./internal/api`) — the case the old check would have passed.

I chose the closure over comparing build action identities because it needs no
build: two `go list` calls answer it in a couple of seconds, and the failure
message names the packages to add. Action identities would be the more exact
instrument if the closure ever proves too coarse.

The cost is smaller than the widening suggests. Narrow lane over the 9 packages:
**305 s**, versus 286 s over the previous 4, against 881 s for the full
shipped-tag lane. `cmd/msgvault/cmd` dominates at 294 s and the other eight run
alongside it. PR body and the two docs updated with the corrected figures.

Worth stating plainly: the previous revision's claim that every other package
compiles a byte-identical test binary was false for those five, and the PR body
asserted it. That is now accurate.

wesm:
Thanks @danshapiro, I took the liberty of rewriting the wall of text PR description =)

danshapiro:
That makes one of us who read it... thank you and apologies for the
verbotsity!

My Book <http://hotseatbook.com>   |   My Twitter
<http://www.twitter.com/danshapiro>   |   My Turtles
<http://www.robotturtles.com/>   |   My, it'd be nice to work together
<http://glowforge.com/jobs>

PR #556, opened 2026-08-07
fix(release): recover publication from existing tags

wesm:
## Why

The `v0.19.0` tag-triggered release built macOS and Windows successfully but never published because both Linux jobs failed inside their bare Ubuntu container: ARM64 lacked `unzip` for Bun setup, and AMD64 later invoked a Node executable the container never installed. Since publication depends on every platform artifact, the release job was skipped.

Retagging or moving `v0.19.0` would weaken release provenance, while rerunning the failed workflow would reproduce the same container failures. The release path therefore needs to be self-contained and able to rebuild an unchanged existing tag deliberately.

## Behavior

- Linux release jobs install Bun's archive dependency, trust the container checkout, and run the web-asset validator with the Bun runtime they already provision.
- The workflow accepts an owner-dispatched existing `v*` tag, checks that tag out in every job, and derives version and commit metadata from the checked-out source rather than the dispatch event SHA.
- Manual retries default to `publish=false`, running the complete six-platform build, manifest comparison, archive validation, smoke test, and checksum generation without creating a GitHub Release. Publication requires a deliberate second dispatch with `publish=true`.
- Automatic publication from future tag pushes keeps the existing behavior, while the release action always targets the selected tag explicitly.

## Review boundaries

This changes only the release workflow. It does not move `v0.19.0`, reuse partial artifacts from the failed run, or publish workstation-built binaries. Once merged, first dispatch `v0.19.0` with publication disabled. Only after that exact pipeline succeeds should the same tag be dispatched with publication enabled.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,7 @@ permissions:
 
 env:
   RELEASE_TAG: ${{ github.event_name == 'workflow_dispatch' && inputs.tag || github.ref_name }}
+  RELEASE_REF: refs/tags/${{ github.event_name == 'workflow_dispatch' && inputs.tag || github.ref_name }}
 
 jobs:
   build-web:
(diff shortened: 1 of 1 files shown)


wesm:
Addressed in `bdd67f43`. Every platform checkout now uses a fully qualified `refs/tags/...` ref, and the peeled-commit validation resolves that same qualified tag ref. A conflicting `v0.19.0` branch can no longer satisfy either boundary.

PR #555, opened 2026-08-07
docs: document the 0.19.0 release

wesm:
## Why

msgvault 0.19.0 adds several major archive sources, API surfaces, identity and profile workflows, and storage changes. The public documentation did not yet collect those changes into a complete release record, and a few existing pages still described obsolete IMAP flags, daemon commands, and provider coverage.

A source-backed release pass keeps operators and integrators from missing new workflows or following stale instructions.

## What changed

- Publishes comprehensive 0.19.0 release notes with contributor acknowledgements.
- Adds the missing guide for source identities, durable person profiles, Fastmail alias discovery, and typed attributes.
- Aligns the README and public guides with the shipped CLI, configuration, API schema, analytics cache, Docker vector support, vCard maintenance, and provider coverage.

## How to use it

Start with the 0.19.0 changelog for the release overview. For identity and profile workflows, follow the new People, Profiles, and Source Identities guide. This PR changes documentation only; it does not alter runtime behavior.

PR #557, opened 2026-08-07
fix(release): stop Windows daemons cleanly

wesm:
Windows ARM64 release builds currently complete initialization and cache generation, but the shutdown smoke then waits for the entire grace period. The daemon has exited; the older Kit dependency only checks whether Windows can open the retained process object, which remains possible while a handle is held.

This upgrades msgvault to the current Kit v0.18.1, which includes the Windows process exit-status fix in its normal release history. The upgrade also exposed two msgvault-side compatibility assumptions: thumbnail rows do not store their own logical size, and the attachment durability test assumed an exact directory-sync count. Maintenance now resolves an unknown thumbnail size from DB-recorded loose candidates, including legacy noncanonical paths, with the canonical location as a fallback. The durability test enforces publication ordering without constraining legitimate cleanup syncs.

The release workflow treats the canonical Windows ARM64 daemon stop as a required, two-minute-bounded smoke check and uploads its isolated state on failure. Nix metadata advances to 0.19.1 with the updated dependency closure. This PR remains scoped to the release unblocker; the larger 0.19.0 documentation update remains isolated in #555.


wesm:
Review follow-up: no code change is needed for the unresolved-size case. In Kit v0.18.1, `prepareCandidate` returns `ErrContentMismatch` for `Size < 0`; `packCandidates` handles every non-cancellation, non-limit candidate read error by incrementing `BlobsCorrupt` and continuing to the next candidate. That error is not returned from maintenance, so a stale thumbnail cannot block valid blobs later in the pass. Keeping the unresolved candidate rather than omitting it also preserves the corrupt diagnostic signal.

PR #559, opened 2026-08-07
fix(slack): advance --limit 1 sweeps past certified overlap

### LINKED ISSUE #558: `internal/slack`: `--limit 1` sweep parks permanently once its watermark lands within 10 minutes after midnight
### Summary

A `sync slack --limit 1` sweep can park permanently: once the persisted sweep watermark (floor) lands at, or within 10 minutes after, a midnight in the user's timezone, every subsequent `--limit 1` run re-searches the same already-certified day, makes no durable progress, and never advances the watermark. This violates the sweep's own documented contract that repeated limited runs converge (`sweep.go:44`).

This is also the root cause of the known near-midnight flake of `TestLimitOneSweepConverges` and `TestSweepLimitOneConvergesPastTruncatedDayOverlap`: those tests derive fixtures from `time.Now()`, so when a full-suite run happens to execute a few minutes after UTC midnight, the persisted floor lands in the pathological band and the tests' (correct) assertions fail. The tests are not over-asserting — production genuinely fails there.

### Mechanism (as of ca3c0cf1)

- The day walk's first day is a pure function of the persisted floor: `floor − sweepLagMargin` (10 min, `sweep.go:16`), day-truncated (`sweep.go:227`, `:235-236`, `:490-492`). `imp.now()` only bounds the far end (`:75`, `:237-238`).
- The budget charges one unit per searched day, check-before-charge (`:256-259`); the truncation exemption (`:241-255`) is the only free day.
- The watermark commits only via `tsLess(floor, min(nextBoundary, ceiling))` (`:230-234`).
- Consequence: when `floor ∈ [midnight, midnight + 10min]`, subtracting the lag margin steps back into the *previous* day, that day consumes the entire `--limit 1` budget, and the watermark comparison never advances. The sweep re-certifies the same day forever.

### Reachability

No wall-clock coincidence is needed. Ordinary mid-day operation can commit the floor to exactly `midnight(D+1)` (a day boundary), which is inside the parked band. In probe runs the watermark stayed frozen from +1 minute through +8 days 6 hours of simulated repeated runs.

Escapes: a single `--limit 2` run, or a manual `--full` reset. The 7-day canonical thread audit eventually re-archives missed replies but never advances the sweep floor, so a parked `--limit 1` configuration silently degrades.

`--limit` is a real production flag (`cmd/msgvault/cmd/sync_slack.go:119`) wired to this budget.

### Evidence status

Honest caveat: this is proven by static analysis of the sweep code plus deterministic probe reproductions driving the importer through its `now` seam — not reproduced against a live Slack workspace.

Ready-made regression tests exist on my fork (currently `t.Skip`ped, they fail against HEAD by design and will pass once this is fixed): [`internal/slack/sweep_parking_test.go`](https://github.com/danshapiro/msgvault/blob/main/internal/slack/sweep_parking_test.go). The same branch makes the whole package's tests deterministic via an injected clock and adds pinned midnight-straddling subtests, so the flake can't mask a future regression. Background write-up: [`docs/plans/2026-08-07-slack-sweep-utc-flake.md`](https://github.com/danshapiro/msgvault/blob/main/docs/plans/2026-08-07-slack-sweep-utc-flake.md). Happy to turn any of it into a PR.

### Suggested fix shape (unvalidated)

Mirror the existing truncation exemption: do not budget-charge a day whose `nextBoundary <= floor` — the walk then reaches the first genuinely-uncertified day before spending the budget, and `--limit 1` regains its guaranteed-progress property.

wesm:
A Slack reply sweep with `--limit 1` could spend its only budget unit re-searching an already-certified overlap day whenever its watermark sat at or shortly after midnight. Repeated scheduled runs then parked forever instead of converging on new reply coverage.

This keeps the ten-minute overlap searchable for late indexing, but charges a day only when its next boundary can advance the persisted floor. Exact-midnight, near-midnight, and ordinary day-behind backlogs can now progress without changing the CLI or persisted state.

Closes #558.


wesm:
Addressed in 3f576f00. Free-overlap discovery now reserves the first uncertified-day charge while recording hits as durable thread debt, so `--limit 1` cannot spend that unit on an immediate canonical fetch. Added a deterministic regression with a late-indexed reply strictly between `queryFloor` and `floor`; it asserts the first limited run advances the watermark and repeated minute-spaced runs archive the reply and move it out of the overlap.

wesm:
Addressed in 43489cbf. Each `sweepRange` now captures whether the shared budget had capacity at entry and, after allowing persisted-truncation skips, returns before any unmarked-day search when it did not. A deterministic regression calls two channel-scoped ranges with the same `--limit 1` budget: the first spends it, and the second is asserted to issue no free-overlap search. Ranges that begin with capacity retain the late-index overlap behavior.

wesm:
Addressed in c322f134. A non-context failure on a finite free-overlap search now consumes one shared-budget unit; chargeable days remain single-counted because they pay before the request. The regression runs two consecutive channel gap ranges with `failSearch` and one shared `--limit 1` budget, then asserts exhaustion and exactly one recorded discovery failure, proving the second range stopped before search.

PR #564, opened 2026-08-08
fix(cache): bound relationship index builds

### LINKED ISSUE #560: build-cache: DuckDB builder policy is hard-coded at 2GB and cannot build the identity index on larger archives
On v0.19.1, `msgvault build-cache --full-rebuild` fails deterministically on an archive with a few million messages:

```
Error: build identity index: build relationship_activity: Out of Memory Error: could not allocate block of size 256.0 KiB (1.8 GiB/1.8 GiB used)

Possible solutions:
* Reducing the number of threads (SET threads=X)
* Disabling insertion-order preservation (SET preserve_insertion_order=false)
* Increasing the memory limit (SET memory_limit='...GB')
```

Every other dataset (participants, conversations, messages, etc.) builds in seconds; only `relationship_activity` inside the identity index exceeds the budget, and the cleanup then reports `drop relationship_build_*: TransactionContext Error: Current transaction is aborted`.

The cause looks like `BuilderPolicy` in `internal/duckdbutil/connection.go`, which hard-codes `MemoryLimit: "2GB"`, 2 threads, and `MaxTempDirectorySize: "8GB"`. As far as I can tell there is no flag, config key, or environment variable to override it — `build-cache --help` only offers `--full-rebuild`, and `preserve_insertion_order` is already disabled by the policy.

The host has plenty of free RAM, but the web UI is left permanently on "Analytical cache unavailable" (the daemon stays on sql-fallback), and the only way out today is patching the source and rebuilding.

Would you take a change that makes the builder policy configurable (e.g. an `[analytics]` config section) or scales it with available system memory? Locally, raising the limits (16GB memory / 64GB temp) is enough to get past it. Happy to test a patch or send a PR.

salmonumbrella:
Fixes #560.

- Stream relationship-activity fan-out and logical chat reductions so large group conversations do not build fan-out-sized hash tables.
- Scope activity work by occurrence year, including direct recipients and conversation memberships.
- Add optional `analytics.builder_memory_limit`, `analytics.builder_threads`, and `analytics.builder_temp_limit` settings while keeping existing defaults.

Output schemas, readers, and incremental behavior remain unchanged.

### DIFF
--- a/cmd/msgvault/cmd/build_cache.go
+++ b/cmd/msgvault/cmd/build_cache.go
@@ -342,6 +342,7 @@ func runBuildCacheLocal(fullRebuild, auto bool) error {
 func runBuildCacheLocalMode(mode buildCacheMode) error {
 	dbPath := cfg.DatabaseDSN()
 	analyticsDir := cfg.AnalyticsDir()
+	builderOverrides := analyticsBuilderOverrides(cfg.Analytics)
 
 	// The Parquet cache is a SQLite -> DuckDB ETL; feeding a postgres:// DSN to
 	// the SQLite driver inside buildCache fails immediately with a confusing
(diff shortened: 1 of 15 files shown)


wesm:
Thank you. I'll merge this, test it locally, and then cut v0.19.2

PR #565, opened 2026-08-08
feat(store): add structured person profile primitives

salmonumbrella:
## What changed

- add structured, historized person names, contact points, addresses, dates, categories, media, and keys
- add an open communication-service catalog plus reviewable scoped identity observations and match candidates
- expose typed profile, history, patch, and service-catalog APIs on SQLite and PostgreSQL

## Why

Msgvault needs one durable person profile with multiple emails, phone numbers, service-scoped usernames, and source evidence. Independent rows preserve per-value history, provenance, normalized lookup, and stable vCard property identity. 

Similar names or contact points create review candidates rather than an irreversible merge. This continues roadmap issue #534.

## Usage

Use `GET /api/v1/persons/{id}/profile` to read the curated profile,
`PATCH /api/v1/persons/{id}/profile` with the current revision to apply an
atomic typed change, and `GET /api/v1/persons/{id}/profile/history` to inspect
prior values. Communication services are available under
`/api/v1/communication-services`.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1397,6 +1397,70 @@ components:
       required:
         - stats
       type: object
+    CommunicationService:
+      additionalProperties: true
+      properties:
+        aliases:
+          items:
+            type: string
+          type:
+            - array
+            - "null"
+        created_at:
+          format: date-time
+          type: string
+        default_scope_kind:
+          type: string
+        display_label:
+          type: string
+        id:
+          format: int64
+          type: integer
+        is_active:
+          type: boolean
+        is_system:
+          type: boolean
+        normalization:
+          type: string
+        normalization_version:
+          format: int64
+          type: integer
+        profile_url_template:
+          type: string
+        scope_policy:
+          type: string
+        slug:
+          type: string
+        updated_at:
+          format: date-time
+          type: string
+        uri_scheme:
+          type: string
+      required:
+        - id
+        - slug
+        - display_label
+        - aliases
+        - scope_policy
+        - normalization
+        - normalization_version
+        - is_system
+        - is_active
+        - created_at
+        - updated_at
+      type: object
+    CommunicationServicesResponse:
+      additionalProperties: true
+      properties:
+        services:
+          items:
+            $ref: "#/components/schemas/CommunicationService"
+          type:
+            - array
+            - "null"
+      required:
+        - services
+      type: object
     ConversationResponse:
       additionalProperties: true
       properties:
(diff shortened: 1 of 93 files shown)


salmonumbrella:
@wesm all you

wesm:
Thanks, I'll rebase this and review

wesm:
Summary of the review work that landed on this branch (the branch has since been rebuilt by the author — spot-checking the new head, all of these changes are incorporated).

We reviewed the PR end to end and triaged three rounds of reviewer findings, fixing the ones that held up:

**Data-integrity fixes**

1. **Contradictory provider IDs are no longer silently dropped.** Re-recording an observation whose address now maps to a *different* provider user ID used to be a silent no-op. It now closes the old row and records the new binding as a fresh observation, so both facts survive in history, and conflicts that the change invalidates are cleaned up.
2. **Automatic match acceptance requires a real stable ID.** A candidate could be system-accepted just by carrying the `stable_provider_id` basis label, with no recorded value (or a blank one). Non-user acceptance now requires a non-blank normalized value recording which stable ID matched.
3. **Malformed scopes can no longer fragment identity keys.** A scope kind without a value (or the reverse) now fails validation, blank/padded scope strings are stored as absent instead of as empty strings, and candidate upserts now run the same service-scope validation as observations and contact points (they previously skipped it entirely).
4. **User decisions survive participant merges.** When a merge collapsed a user-accepted candidate with an observation-generated conflict, the decision's who/when/notes were discarded and a later cleanup could quietly reopen the match as undecided. Decision metadata is now preserved, and a new `pre_conflict_state` column lets cleanup restore the accepted/rejected decision once the conflicting observations go away, instead of demoting to an undecided candidate.
5. **Cross-kind observation pairs no longer prop up conflicts.** Conflict generation only ever pairs observations of the same address kind, but cleanup accepted mismatched pairs (e.g. a username on one side, a social handle on the other) as support — so a resolved conflict could stay open forever. Support now requires the pair to share one kind.

**Findings we assessed and did not change**

- Deleting candidates when their observation/person endpoints are removed is required — the endpoints are polymorphic with no foreign keys, so keeping the rows would leave dangling references. Full review-history preservation would need an audit/tombstone design (possible follow-up).
- Keying candidates by address kind (as suggested) would break the flow where an observation conflict promotes a manually created candidate, and would split review of what is semantically one shared handle. The same-kind pair requirement above fixes the real half of that finding without the schema cost.

**Known follow-ups left for later:** conflict generation currently runs for every address kind including `language` (two people sharing "en" would generate a conflict — not reachable yet since observation recording isn't API-exposed); minor cosmetics in the service-catalog POST (decode error message, 200 vs 201); a test-only variable living in the migration file; two undropped temp tables in subset creation.

All changes were made test-first (each new test failed on the exact claimed behavior before the fix), with the full suite, `go vet`, and gofmt clean.

PR #566, opened 2026-08-08
feat(mcp): support protocol 2026-07-28

salmonumbrella:
## What changed

- Move msgvault MCP to the official Go SDK v1.7.0.
- Give 2026-07-28 clients the initialize-free `server/discover` path. Streamable HTTP is sessionless and does not issue `Mcp-Session-Id`; older protocol versions remain a compatibility fallback and do not change the modern path.
- Add strict JSON Schema 2020-12 input and output contracts, structured results with JSON text fallback, opaque attachment resources, cache metadata, trace propagation, and private internal-error isolation.
- Harden HTTP with origin, protocol-header, request-size, rate, concurrency, and read-only-by-default controls.
- Expand raw modern and legacy transport coverage across every advertised tool shape, and make the daemon-stop progress test deterministic.

## Why

The previous server used a 2025-era stateful MCP stack. It could not give current agents sessionless discovery, per-request protocol metadata, structured output, resource caching, and the 2026-07-28 HTTP contract without msgvault owning a second protocol implementation. The official SDK now provides the modern path while keeping older clients usable.

## Usage

Run `msgvault mcp` for stdio with the full local tool set. Run `msgvault mcp --http 127.0.0.1:8080` for stateless Streamable HTTP; HTTP stays read-only by default. Add `--http-allow-writes` only for trusted clients that need `export_attachment` or `stage_deletion`.

### DIFF
--- a/cmd/msgvault/cmd/mcp.go
+++ b/cmd/msgvault/cmd/mcp.go
@@ -18,6 +18,7 @@ var mcpForceSQL bool
 var mcpNoSQLiteScanner bool
 var mcpHTTPAddr string
 var mcpHTTPAllowInsecure bool
+var mcpHTTPAllowWrites bool
 var serveMCPHTTPWithOptions = mcpserver.ServeHTTPWithOptions
 
 var mcpCmd = &cobra.Command{
@@ -66,7 +67,11 @@ Add to Claude Desktop config:
 			if err != nil {
 				return usageErr(cmd, err)
 			}
-			return serveMCPHTTPWithOptions(ctx, opts, normalized, cfg.Server.APIKey)
+			return serveMCPHTTPWithOptions(ctx, opts, mcpserver.HTTPOptions{
+				Addr:        normalized,
+				APIKey:      cfg.Server.APIKey,
+				AllowWrites: mcpHTTPAllowWrites,
+			})
 		}
 		return mcpserver.ServeWithOptions(ctx, opts)
 	},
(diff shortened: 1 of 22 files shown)


salmonumbrella:
Addressed the `get_stats` finding in a0c287e. Vector statistics remain best effort: partial `StatsView` data is returned with archive and account statistics, while the underlying vector error is logged privately. Added a regression test that failed on the prior behavior. Focused MCP tests, the full `internal/mcp` package, repo vet, and lint (`0 issues`) pass.

salmonumbrella:
Addressed the daemon error-translation finding in a2ded20. MCP now converts only a fixed whitelist of known daemon request/search codes to static tool-error messages. Raw daemon messages never cross the wire; unknown codes still return generic `-32603` and are logged privately. Regression coverage runs real daemonclient adapters for body, hybrid, and similar searches. Focused adapter tests, the full `internal/mcp` package, repo vet, and lint (`0 issues`) pass.

salmonumbrella:
Fixed the resource-read limiter gap in `90b51cbc`. `resources/read` now shares the same process-wide rate and concurrency policy as `tools/call`. Saturated reads return a stable `-32000` JSON-RPC busy error before the attachment reader runs. Added raw modern HTTP regression coverage for both token-burst exhaustion and concurrent reads. Focused tests, the full tagged `internal/mcp` package, repository-wide vet, and lint (`0 issues`) pass.

salmonumbrella:
Pushed final CI portability corrections in `7c8ea0d2`. The raw stdio peer now uses portable bounded I/O instead of `os.File.SetDeadline`, which Windows anonymous pipes do not support. The Nix `vendorHash` is updated to the exact fixed-output hash reported after the official SDK dependency change. Focused stdio/resource tests, the full tagged `internal/mcp` package, repository-wide vet, and lint (`0 issues`) pass. Roborev found no issues on the prior head; this final head is ready for another review.

wesm:
Thank you

PR #567, opened 2026-08-08
feat(oauth): request read-only Gmail access with --readonly

quetza:
Adds a `--readonly` flag to `add-account` that requests Gmail read access (`gmail.readonly`) instead of the default read + modify pair.

## What changed

- `add-account --readonly` authorizes an account with read-only Gmail access. Sync, search, and the TUI work as normal; deletion prompts to re-authorize with write access.
- An account that already has write access can't be narrowed in place, so `--readonly` alone fails and points at `--readonly --force`, which revokes the old grant at Google, deletes the stored token, and re-authorizes with the narrower scopes. Revoking at Google means copies of the old token (backups, other machines) lose write access too.
- If revocation fails, nothing is changed. If msgvault has no working credential to revoke with — the stored token is already dead, or missing for a registered account — it asks you to revoke access at [myaccount.google.com/permissions](https://myaccount.google.com/permissions) and confirm before continuing.
- Any Gmail scope other than `gmail.readonly` counts as write access, including scopes msgvault never requests that could appear in a token created by another tool.
- Calendar and Drive grants survive narrowing, and re-authorization paths (expired tokens, `add-calendar`) no longer quietly restore write access to a narrowed account.
- Recovery hints and headless instructions keep `--readonly`, so following them doesn't undo the narrowing. A plain `add-account` on a read-only account warns before requesting write access.
- The deletion pre-flight accepts `mail.google.com` as sufficient for trashing; it previously demanded `gmail.modify` even from accounts with broader access.

## Why

Sync only reads mail, but `add-account` always requested `gmail.modify`, so there was no way to run msgvault with a read-only grant. Restricting scopes in the Google Cloud Console doesn't help — that page is a declaration for verification review, not a limit on what gets granted — so changing what the client requests is the only lever.

## Usage

```console
msgvault add-account you@gmail.com --readonly          # new account
msgvault add-account you@gmail.com --readonly --force  # narrow an existing account
```

Default behavior without `--readonly` is unchanged.

## Limitations

- `--readonly` is rejected for service accounts; their scopes come from the domain-wide delegation grant in the Admin Console.
- Narrowing needs a browser. On a headless server, `--headless --readonly` prints steps to narrow on a browser machine and copy the token across.

Also expands the Google Cloud Console setup walkthrough in `docs/guides/oauth-setup.md`.

### DIFF
--- a/cmd/msgvault/cmd/addaccount.go
+++ b/cmd/msgvault/cmd/addaccount.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"go.kenn.io/msgvault/internal/oauth"
@@ -17,10 +19,17 @@ var (
 	forceReauth                 bool
 	oauthAppName                string
 	noDefaultIdentityAddAccount bool
+	readonlyGrant               bool
 )
 
 // addAccountUse is the usage string for the add-account command.
 const addAccountUse = "add-account <email>"
+
+// addAccountGrantDecidedFlag records that the grant decision was already made
+// by the frontend CLI before it proxied to the daemon. The decision gates
+// authorization, so re-running it in the subprocess — after the frontend has
+// already authorized — would refuse a token this run just minted.
+const addAccountGrantDecidedFlag = "grant-decided"
 
 // errGmailSourceNotFound is returned by findGmailSource when no Gmail
 // source is registered for the given identifier. Wrapped via fmt.Errorf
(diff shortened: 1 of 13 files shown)


wesm:
will review thank you

wesm:
I can take the wheel from here to get this merge-ready, thank you

wesm:
Our clankers are competing with each other so i'm going to open a new PR

wesm:
Your force-push erased a bunch of my work so I'll complete the feature in #594

wesm:
I haven't forgotten about this, things got busy, I will get this feature included in 0.20.0

PR #575, opened 2026-08-09
fix: harden daemon jobs and meeting imports

### LINKED ISSUE #539: sync-beeper hangs indefinitely after 13 chats on a WhatsApp account (Signal/Matrix unaffected)
## Summary

`msgvault sync-beeper` reliably stops making progress after processing 13 of 425 WhatsApp
chats. The CLI never returns; the first occurrence sat idle for **13.5 hours** before I
noticed. No error is printed and no timeout fires.

The same command scoped to the other two Beeper accounts (`signal`, `matrix`) completes
normally, so the problem appears specific to the WhatsApp bridge account rather than to
the Beeper integration as a whole.

## Environment

| | |
|---|---|
| msgvault | 0.18.0 (Homebrew bottle; `commit: unknown`) |
| Go | go1.26.5 |
| Platform | darwin/arm64 |
| macOS | 26.6 |
| Beeper Desktop | 4.3.0 |
| Beeper accounts | `matrix`, `signal`, `local-whatsapp_ba_conr…` |
| Archive size | ~74k messages, 12 GB SQLite DB |

Daemon runs persistently via a LaunchAgent (`msgvault serve`, pinned `api_port = 8788`,
`api_key` set).

## Steps to reproduce

```bash
msgvault add-beeper          # registers matrix, signal and the WhatsApp bridge account
msgvault sync-beeper
```

## Expected

The backfill walks all 425 chats and returns, or fails with a diagnosable error.

## Actual

Output stops after chat 13 and never resumes:

```
Syncing Beeper account local-whatsapp_ba_conrF33mMUTSRh7A-6NNV9ijpU0
  chat 1/425 (WhatsApp): 2 messages
  chat 2/425 (WhatsApp): 54 messages
  chat 3/425 (WhatsApp): 6 messages
  chat 4/425 (WhatsApp): 4 messages
  chat 5/425 (WhatsApp): 7 messages
  chat 6/425 (WhatsApp): 2 messages
  chat 7/425 (WhatsApp): 1 messages
  chat 8/425 (WhatsApp): 0 messages
  chat 9/425 (WhatsApp): 0 messages
  chat 10/425 (WhatsApp): 0 messages
  chat 11/425 (WhatsApp): 0 messages
  chat 12/425 (WhatsApp): 0 messages
  chat 13/425 (WhatsApp): 0 messages
<no further output; killed manually after 150s>
Error: decode CLI run stream: terminated signal received
```

Note that chats 8–13 already report `0 messages`, which looks suspicious given these are
active chats that had never been synced before.

## Diagnostics while hung

**The CLI process is idle, not busy.**

```
  PID ELAPSED  %CPU %MEM STAT COMMAND
49004   05:00   0.0  0.1 SN   msgvault sync-beeper
```

**Its only socket is to the local daemon** — there is no connection to Beeper:

```
msgvault 49004  6u  IPv4  TCP 127.0.0.1:59085->127.0.0.1:8788 (ESTABLISHED)
```

**The daemon has no connection to Beeper either.** Beeper Desktop is running and
listening the whole time:

```
$ lsof -nP -iTCP -sTCP:LISTEN | grep -i beeper
Beeper  9715  101u  IPv4  TCP 127.0.0.1:23373 (LISTEN)

$ lsof -nP -p <daemon-pid> | grep TCP | grep -v ':8788'
<empty>
```

So the work appears to stall before (or without) any request reaching the Beeper API.

**The daemon keeps the CLI request open indefinitely:**

```
level=WARN msg="http request in progress" method=POST path=/api/v1/cli/run
      request_id=msgvault-74 elapsed=13h31m46.545s
```

**Beeper Desktop stays healthy throughout.** Its local API answered other clients normally
while the sync was stuck, so the bridge itself is not wedged.

## What does work

Scoping the sync to the other two accounts completes cleanly in seconds:

```bash
msgvault sync-beeper --account signal --account matrix
# → signal: 314 messages, matrix: 10 messages, outcome=ok
```

## Reproducibility

Reproduced 3×, hanging at the same point every time:

1. Fresh CLI invocation
2. After restarting the daemon (`launchctl kickstart -k …`) to rule out stale in-process state
3. With the sync scoped to only the WhatsApp account (`--account local-whatsapp_…`)

## Impact

Only **14 of 425** WhatsApp chats are archived. The messages that did land come almost
exclusively from a handful of large group chats; virtually all 1:1 conversations are
missing. `last_sync_at` stays `NULL` for all three Beeper sources, so no cursor is ever
recorded and re-running restarts from the same place.

## Possibly useful details

- Killing a stuck run needs care: the process shows up as
  `msgvault --no-log-file sync-beeper`, so `pkill -f "msgvault sync-beeper"` does not match it.
- Beeper sources are not covered by the `[[accounts]]` schedule in `config.toml`, so they
  never sync automatically — only the IMAP email account does. Not a bug as such, but it
  means a hung backfill goes unnoticed.
- Happy to run further diagnostics (`-v`, `SIGQUIT` goroutine dump, etc.) if that helps.``

### LINKED ISSUE #574: Scheduled syncs starve the operation gate: gated endpoints unreachable when the scheduler never idles
## Summary

On a daemon with enough configured sources that the sync scheduler never goes idle, gated mutating endpoints become permanently unreachable. `POST /api/v1/import/meeting` returns `503 operation_in_progress` on every attempt, indefinitely — not as a transient backoff, but as a steady state with no way out short of reducing sync frequency.

## Reproduction

Nine configured sources on staggered five-minute cron schedules, spread one or two per minute:

```toml
[beeper]      schedule = "0-59/5 * * * *"
[[accounts]]  schedule = "1-59/5 * * * *"
[[accounts]]  schedule = "2-59/5 * * * *"
[[gcal]]      schedule = "3-59/5 * * * *"
[[gcal]]      schedule = "4-59/5 * * * *"
# ...nine total, two sharing some minutes
```

Each sync takes roughly 40 seconds, so a new one begins as the previous ends.

Polling `GET /api/v1/health` once per second across a full five-minute cycle, `operation.busy` was never `false`:

```
02:50:49  busy=true  "a scheduled sync"
02:51:32  busy=true  "a scheduled sync"   <- new started_at, 43s later
02:52:12  busy=true  "a scheduled sync"   <- new started_at, 40s later
```

Every POST to the import endpoint is held for about ten seconds (`operationGateWaitLimit`) and then rejected:

```
HTTP/1.1 503 Service Unavailable
{"error":"operation_in_progress","message":"a scheduled sync has been running for 6s"}
```

Sustained retries over several minutes never got through. A client-side queue of meetings has been undeliverable for a week.

## Expected

The yield contract in the code should cover exactly this case:

- `SerialOperationGate.BeginRequestWorkContext` increments `requestWaiters` while an API request is queued
- `SerialOperationGate.HasRequestWaiters` reports it
- `labeledWorkTracker.ShouldYield` (`cmd/msgvault/cmd/work_tracker.go`) surfaces it to scheduled work
- `internal/scheduler/scheduler.go` polls `ShouldYield()` every `yieldPollInterval` (5s) and cancels the running job with `ErrYieldedToWaiter`

With a 5-second yield poll and a 10-second request budget, a queued import should acquire the gate on the first yield.

## Observed

It never does. Throughout sustained request pressure the scheduled sync's `started_at` kept its regular ~40s rhythm, with no sign of a shortened, yielded run.

I could not determine from reading the source alone which link fails. Candidates, in the order I would check them:

1. The scheduled-sync path is not wrapped in the yield context, or the tracker it receives does not implement `YieldChecker`, so `ShouldYield` is never consulted.
2. The yield fires, but the job does not release the gate within the requester's remaining budget.
3. The gate is released, but with a new scheduled job due every minute background work re-acquires it. `BeginLabeledWorkContext` is a plain semaphore acquire with no priority for request waiters.

## Why this is worth fixing

`internal/api/meeting_import_test.go` asserts the 503 under gate contention, so gating the endpoint is clearly intentional and I am not arguing against it. The problem is that the yield which makes the gate survivable does not appear to reach a saturated scheduler. Any deployment with enough sources to keep the scheduler continuously busy loses every gated mutating endpoint, and no configuration restores it except cutting sync frequency — which trades away the freshness the schedule was configured for in the first place.

A `HasRequestWaiters`-aware admission order on the gate, so a queued request is served before the next background job, would close this regardless of which of the three candidates above turns out to be the cause.

## Version

Observed on v0.19.1. Source read from current `main`. Not addressed in v0.19.2, whose notes cover cache-build progress, relationship index bounding, checksum compatibility, and Fastmail pod-scoped JMAP API URLs.

### LINKED ISSUE #576: auto_build_cache = false is ignored by the post-sync and post-import cache rebuild
## Summary

`[analytics] auto_build_cache = false` is honoured by the startup build but ignored by the post-sync and post-import rebuild path. On a daemon with several scheduled sources, that means the setting does not actually stop automatic cache builds — it only delays the first one until the first sync completes.

## The two call sites

Startup gates on the setting (`cmd/msgvault/cmd/serve.go:781`, v0.19.2):

```go
(staleness.NeedsBuild && c.Analytics.AutoBuildCache)
```

The post-sync path does not (`cmd/msgvault/cmd/build_cache.go:2152-2166`, v0.19.2):

```go
func rebuildCacheAfterScheduledSync(ctx context.Context, identifier string) error {
	dbPath := cfg.DatabaseDSN()
	if store.IsPostgresURL(dbPath) {
		return nil
	}
	analyticsDir := cfg.AnalyticsDir()
	staleness := cacheNeedsBuild(dbPath, analyticsDir)
	if !staleness.NeedsBuild {
		return nil
	}
	logger.Info("rebuilding cache after sync", ...)
```

`cfg.Analytics.AutoBuildCache` is never consulted. `grep -rn AutoBuildCache` at v0.19.2 finds it only in `serve.go`, `internal/api/settings.go` and `internal/config/config.go` — not in `build_cache.go`.

## Observed

With `auto_build_cache = false` and `engine = "sql"` set in `config.toml`, and the daemon confirming the setting at startup:

```
level=INFO msg="using live SQL analytics engine" engine=sql
```

every scheduled sync still logs:

```
level=INFO msg="rebuilding cache after sync" identifier=gcal:<redacted> \
  reason="analytics cache schema is stale" full_rebuild=true
```

The same function is reached from meeting import, which logs `identifier=meeting_import:<source>`.

## Why it matters

The documented way to opt out of automatic cache builds is this key. An operator who sets it because full rebuilds are too expensive on their archive still gets a full rebuild after every sync. On an install with nine sources on staggered five-minute schedules that is a rebuild attempt roughly every minute, each one holding the global operation gate for its duration.

It also means the key cannot be used as a mitigation while a cache build problem is being investigated, which is the situation where an operator is most likely to reach for it.

## Suggested fix

Consult the same condition in `rebuildCacheAfterScheduledSync` that `serve.go` uses, so the setting means the same thing on both paths. If the intent is that a stale *schema* must always be rebuilt regardless of the setting, that is defensible but undocumented — `docs/configuration.md` describes the key without that exception, and the startup path does not apply it either.

## Version

Observed on v0.19.2; source read from the v0.19.2 tag. Behaviour is identical on v0.19.1.

### LINKED ISSUE #577: Meeting import returns 500 after the meeting is already persisted, when the post-import cache rebuild fails
## Summary

`POST /api/v1/import/meeting` persists the meeting, then triggers an analytics cache rebuild, and returns `500 internal_error "Meeting import failed"` if that rebuild fails. The meeting is **not** rolled back. A client that trusts the status code re-imports a meeting that already landed, indefinitely.

## Observed

Four meetings imported against a large SQLite archive. Every request returned:

```
HTTP/1.1 500 Internal Server Error
{"error":"internal_error","message":"Meeting import failed"}
```

Server side, in order:

```
level=INFO  msg="rebuilding cache after sync" identifier=meeting_import:<source> \
            reason="analytics cache schema is stale" full_rebuild=true
level=ERROR msg="cache build failed" error="build-cache subprocess: exit status 1; \
            output: ... Error: build identity index: build relationship_activity: \
            Out of Memory Error ..."
level=ERROR msg="meeting import failed" source=<source> external_id=<id> error_class=internal
level=INFO  msg="http request" method=POST path=/api/v1/import/meeting status=500 duration=38.5s
```

All four meetings are nevertheless present and searchable afterwards. `GET /api/v1/search` returns each one with `message_type=meeting_transcript`, the correct `sent_at`, and `source_message_id=meeting:<external_id>`. The import itself succeeded every time; only the follow-on cache build failed.

## Why this is worse than a slow import

The endpoint is explicitly designed for spooling clients — re-posting is safe because the importer keys on `external_id` and replaces in place. A spooling client is therefore built to retry on failure, and a status code is the only signal it has. Reporting a post-write side effect as the write's own failure inverts that contract: the queue never drains, and each retry costs another full cache rebuild attempt.

In our case a client-side spool of four meetings sat undelivered for a week while every meeting in it had in fact been imported on the first attempt.

## Suggested fix

Treat the post-import cache rebuild as a side effect, not part of the import's success condition. Either:

- return `200`/`202` once the meeting is committed and log the cache failure separately, optionally surfacing it in the response body or on `/api/v1/health`; or
- if the rebuild genuinely must be part of the transaction, roll the meeting back on failure so the status code and the stored state agree.

The first seems right: the cache is derived data and is rebuilt on the next sync anyway.

## Related

- The rebuild fires even with `[analytics] auto_build_cache = false` — filed separately.
- The OOM in the underlying build is a separate matter and was addressed for our archive by #564; this report is about the status code, which would misreport any cache-build failure, not just an OOM.

## Version

Observed on v0.19.1. The relevant path is unchanged in v0.19.2: `rebuildCacheAfterScheduledSync` returns an error to its caller on build failure, and meeting import surfaces that as `error_class=internal`.

salmonumbrella:
## What changed

- Serve queued API mutations before background operation-gate waiters without dropping or cancelling queued background work.
- Bound provider-directed retry delays for Beeper, Slack, Teams, and Granola, with safe fallback for malformed or overflowing `Retry-After` values.
- Honor `analytics.auto_build_cache = false` after scheduled syncs and meeting imports, and warn when a ready DuckDB cache will not refresh automatically.
- Keep persisted meeting imports successful when the derived cache refresh fails, and log that failure separately.

## Why

Busy schedulers could starve user-triggered writes, oversized retry delays could make a sync look hung, and disabled cache builds still ran after syncs. A cache refresh failure could also turn a successful meeting import into a false 500 and trap spooling clients in retries.

## Usage

No usage change.

Closes #574.
Closes #539.
Closes #576.
Closes #577.

### DIFF
--- a/cmd/msgvault/cmd/build_cache.go
+++ b/cmd/msgvault/cmd/build_cache.go
@@ -2150,6 +2150,11 @@ func globalConfigFlagArgs() []string {
 // to keep DuckDB's bundled SQLite library out of a long-lived daemon's
 // address space (issue #379).
 func rebuildCacheAfterScheduledSync(ctx context.Context, identifier string) error {
+	if !cfg.Analytics.AutoBuildCache {
+		// AutoBuildCache opts out of automatic daemon rebuilds, even when startup
+		// selected a usable DuckDB cache; engine = "sql" is the live-data choice.
+		return nil
+	}
 	dbPath := cfg.DatabaseDSN()
 	if store.IsPostgresURL(dbPath) {
 		return nil
--- a/cmd/msgvault/cmd/cache_refresh_test.go
+++ b/cmd/msgvault/cmd/cache_refresh_test.go
@@ -332,6 +332,34 @@ func TestRebuildCacheAfterWriteReturnsError(t *testing.T) {
 	require.ErrorContains(err, "refresh analytics cache")
 }
 
+func TestScheduledCacheRefreshSkipsWhenAutoBuildCacheDisabled(t *testing.T) {
+	require := require.New(t)
+	assert := assert.New(t)
+	tmpDir := t.TempDir()
+
+	savedCfg := cfg
+	t.Cleanup(func() { cfg = savedCfg })
+	cfg = &config.Config{
+		HomeDir: tmpDir,
+		Data:    config.DataConfig{DataDir: tmpDir},
+		Analytics: config.AnalyticsConfig{
+			AutoBuildCache: false,
+		},
+	}
+
+	builds := 0
+	oldRunBuild := runBuildCacheSubprocess
+	runBuildCacheSubprocess = func(context.Context, bool, bool) error {
+		builds++
+		return errors.New("unexpected scheduled cache build")
+	}
+	t.Cleanup(func() { runBuildCacheSubprocess = oldRunBuild })
+
+	err := rebuildCacheAfterScheduledSync(context.Background(), "disabled")
+	require.NoError(err)
+	assert.Zero(builds, "disabled auto_build_cache must not start a cache build")
+}
+
 func TestRepairEncodingReturnsCacheRefreshError(t *testing.T) {
 	require := require.New(t)
 	tmpDir := t.TempDir()
@@ -373,7 +401,13 @@ func TestScheduledCacheRefreshFailurePreservesCompletedSyncRun(t *testing.T) {
 
 	savedCfg := cfg
 	t.Cleanup(func() { cfg = savedCfg })
-	cfg = &config.Config{HomeDir: tmpDir, Data: config.DataConfig{DataDir: tmpDir}}
+	cfg = &config.Config{
+		HomeDir: tmpDir,
+		Data:    config.DataConfig{DataDir: tmpDir},
+		Analytics: config.AnalyticsConfig{
+			AutoBuildCache: true,
+		},
+	}
 
 	sentinel := errors.New("scheduled cache sentinel")
 	oldRunBuild := runBuildCacheSubprocess
--- a/cmd/msgvault/cmd/meeting_import_e2e_test.go
+++ b/cmd/msgvault/cmd/meeting_import_e2e_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log/slog"
 	"net/http"
@@ -120,4 +121,50 @@ func TestMeetingImportAPIToStoreUpdatesCanonicalMessage(t *testing.T) {
 	assert.Contains(current.Body, "Replacement summary.")
 	assert.Contains(current.Body, "Speaker 2: replacement transcript")
 	assert.Empty(current.To)
+}
+
+func TestMeetingImportReturnsSuccessWhenCacheRefreshFails(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+
+	st := testutil.NewTestStore(t)
+	cacheErr := errors.New("synthetic cache refresh failure")
+	adapter := &storeAPIAdapter{
+		store: st,
+		meetingImporter: meetingimport.NewImporter(st, meetingimport.Hooks{
+			AfterSourceSetup: func() error { return nil },
+			RefreshCache: func(context.Context, string) error {
+				return cacheErr
+			},
+		}),
+	}
+	srv := api.NewServer(
+		&config.Config{Server: config.ServerConfig{}},
+		adapter,
+		nil,
+		slog.New(slog.DiscardHandler),
+	)
+
+	req := httptest.NewRequest(
+		http.MethodPost,
+		"/api/v1/import/meeting",
+		bytes.NewBufferString(meetingImportE2EBody),
+	)
+	req.Header.Set("Content-Type", "application/json")
+	resp := httptest.NewRecorder()
+	srv.Router().ServeHTTP(resp, req)
+
+	require.Equal(http.StatusCreated, resp.Code, "body: %s", resp.Body.String())
+	var created api.MeetingImportResponse
+	require.NoError(json.NewDecoder(resp.Body).Decode(&created))
+	assert.Equal(meetingimport.StatusCreated, created.Status)
+	assert.NotZero(created.MessageID)
+
+	var persisted int
+	require.NoError(st.DB().QueryRow(st.Rebind(`
+		SELECT COUNT(*)
+		FROM messages
+		WHERE id = ? AND source_message_id = ? AND message_type = ?
+	`), created.MessageID, "meeting:42", meetingimport.MessageType).Scan(&persisted))
+	assert.Equal(1, persisted)
 }
--- a/cmd/msgvault/cmd/serve.go
+++ b/cmd/msgvault/cmd/serve.go
@@ -480,8 +480,12 @@ func runServe(cmd *cobra.Command, args []string) error {
 		AfterSourceSetup: func() error {
 			return runPostSourceCreateMigrations(s)
 		},
-		RefreshCache: rebuildCacheAfterScheduledSync,
-	})
+		RefreshCache: func(_ context.Context, label string) error {
+			// The import is already durable. Keep the refresh independent of the
+			// client, but let daemon shutdown stop it before the store closes.
+			return daemonCacheRefreshError(ctx, rebuildCacheAfterScheduledSync(ctx, label))
+		},
+	}).WithLogger(logger)
 	storeAdapter := &storeAPIAdapter{
 		store:                 s,
 		attachmentMaintenance: attachmentMaint,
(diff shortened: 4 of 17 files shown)


wesm:
rebasing

Issue #550, opened 2026-08-06
Embed chat conversations as time-gap windows; chunk meeting transcripts at speaker turns with per-chunk context


salmonumbrella:
The vector corpus has exactly one embedding unit: a single message. The worker feeds `Preprocess(subject, body)` into `ChunkText`, which only ever splits a message that overflows the window — nothing is ever grouped *up*. That unit fits email well (a message plus its `Subject:` prefix is a self-contained document), but it is the wrong unit for the two other content families the archive now holds: chat messages (Beeper sources) and meeting transcripts (Granola). This proposes fixing both — sessionize chats into conversation windows, make speaker turns the chunk boundary for transcripts — with contextualized chunk embeddings (`voyage-context-4`) as the target model path: window/note as the document, member message/turn-pack as the chunk, on the existing generation machinery. Part of the #534 delivery sequence.

### Problem: chat messages embed as isolated fragments

- A chat message is embedded as bare body text. `BodyTextForEmbedding` returns the body, subject is empty for chats, and nothing prepends the sender or timestamp — so the vector for a message doesn't even encode who said it.
- Typical chat turns are a few words. "sounds good, Tuesday works" produces a vector with no retrievable meaning; the meaning lives entirely in the adjacent turns of the conversation.
- The query shapes this archive exists for (#534 — "what did X and I discuss/decide about Y") span multiple turns. The BM25 leg of hybrid search can still rescue exact tokens, but the vector leg contributes almost nothing for chat sources, which defeats the point of semantic search over the most conversational part of the corpus.

### Problem: transcript chunks lose their meeting and split speaker turns

One Granola note is one message: `buildBody` renders title, time range, attendee names, the AI summary, then the speaker-labeled transcript. Long notes overflow `ChunkText`'s window, which cuts at paragraph/sentence/word soft breaks. Two defects follow:

- Cut points ignore speaker turns, so a turn can straddle a chunk boundary even though turns are already line-structured in the rendered body.
- Every chunk past the first loses the header. A window of raw transcript lines embeds with no indication of which meeting it came from, when it happened, or who was in the room — exactly the metadata a "context before I meet X" query keys on.

### Proposal: conversation-window embedding units for chat sources

- Sessionize each `conversation_id` by silence gap (e.g. 30 minutes, capped at a day boundary and at a rune budget aligned with `ChunkText`'s window). A window is the ordered member messages rendered as `Display Name (HH:MM): text` lines under a one-line header (chat name or participant list + date).
[...]

PR #589, opened 2026-08-09
feat(vector): add contextual chat and transcript embeddings

salmonumbrella:
## What changed

- Add a `voyage-contextual` embedding API format for Voyage Context 4 while keeping the existing OpenAI-compatible path as the default.
- Embed chats as gap-, day-, and size-bounded windows, and embed meeting notes and speaker turns as contextual documents with exact source-span ownership.
- Add durable change journals, document publication ledgers, and atomic publication/CAS behavior for SQLite and PostgreSQL so edits, moves, deletes, retries, and crashes converge safely.
- Wire contextual rebuild, incremental scheduling, activation, hybrid search excerpts, and a synthetic retrieval evaluator through the production paths.

## Why

The current embedding path treats each message or chunk in isolation. That removes the conversation or speaker context needed to retrieve short and ambiguous answers. This adds grouped contextual embeddings without changing the legacy default. 

## Usage

Set `VOYAGE_API_KEY`, then configure:

```toml
[vector]
enabled = true
backend = "sqlite-vec"

[vector.embeddings]
api_format = "voyage-contextual"
endpoint = "https://api.voyageai.com/v1"
api_key_env = "VOYAGE_API_KEY"
model = "voyage-context-4"
dimension = 1024
```

`api_format = "voyage-contextual"` selects the contextual document endpoint. If `api_format` is omitted, msgvault keeps using the regular OpenAI-compatible embeddings path.

Build and atomically activate a new generation:

```bash
msgvault embeddings build --full-rebuild --yes
```

Refs #550


salmonumbrella:
@wesm I am not sure about this one. Goal is better retrieval of chat messages—needs testing to see if it's worth. 

cc: @danshapiro @jesserobbins @mariusvniekerk would love your input. This [Voyage blog post](https://blog.voyageai.com/2026/06/29/voyage-context-4/) served as inspiration.

wesm:
looking

wesm:
Reviewed the full diff, ran build/vet and the affected test suites with the project build tags (all green; PG-specific tests not run locally — needs a `make test-pg-both` pass before merge).

**Overall:** The core claims hold up under tracing — publication is atomic in both backends, I couldn't construct a lost-mutation schedule in the journal lifecycle, span ownership partitions cleanly, and the legacy OpenAI path is behavior-identical for existing users. Test quality is well above the codebase norm. Two things should be fixed before merge; the rest is non-blocking.

**Should fix before merge**

- **Rejected documents block activation forever.** `context_worker.go:889` — a doc the API rejects as too large gets `ResetEmbedGen`, so its members stay uncovered and the generation can never converge. Packing caps bytes (100KB) but Voyage limits tokens, so a token-dense doc under the byte cap retries on every backstop indefinitely, paying an API call each time. Needs a terminal "rejected" coverage state or truncation fallback.
- **Chat body selection mismatch drops content.** `assemble.go:432` picks `body_html` when `TRIM(body_text)` is blank, but `body.go:9` prefers `body_text` whenever non-empty — so a row with whitespace `body_text` + real HTML silently owns zero documents. Align the predicates; this row shape is untested.

**Non-blocking, worth addressing**

- Eval harness allows 480KB documents (`arms.go:621`) vs production's 100KB (`serve_vector.go:24`), so its numbers describe a more generous policy than users get; it also freezes `MaxChunkRunes`/`ChatGap` as literals asserted only against themselves.
- ~600 lines of publisher/CAS logic near-duplicated between `sqlitevec/documents.go` and `pgvector/documents.go`, and the participant-digest SQL duplicated byte-for-byte (`embed_gen.go:305`, `assemble.go:379`) — drift in either digest copy means CAS misses every time and scopes republish forever with no error. A test guards it today; sharing the code would guard it structurally.
- Latent PG livelock: `CAST(p.updated_at AS TEXT)` renders per-session `TimeZone`/`DateStyle`; any future per-connection GUC divergence makes digests permanently mismatch. Use `to_char` with an explicit format.
- `embed_job.go` runs the archive-wide convergence check twice per tick for an unconverged build, and a transient failure of the first check skips the backstop that tick.
- A benign coverage-CAS miss (`context_worker.go:906`) fails the whole tick instead of skipping to the next scope; the coalesced-fanout resume discards its subcursor (bounded restarts, untested); the bisect path discards paid-for embeddings on a transient left-half error.
- Every SQLite message persist now dirties the `embedding_change_clock` row even with the journal disabled (`messages.go:815` + 2 sites) — deliberate, but it costs all users a page write per persist; deserves a comment.

**Minor**

- `NewEngine(client any)` erases type safety (nil-dispatch panic instead of constructor error).
- Four worker functions at 123–167 lines encode the trickiest cursor/budget semantics; extraction would help future review.
- Run budget passed via `context.WithValue`; a few bare `return err` in a file that otherwise wraps; `vectorFeatures.Worker` is a dead alias introduced by the same PR that renamed it; the `go build`-in-test eval integration test runs on every `make test` — consider a `testing.Short()` guard.

Also verified: no SQL injection, migrations idempotent on populated DBs, journal ships disabled so upgraded archives pay nothing until opt-in, test data PII-clean, no committed secrets, API key never logged.

wesm:
looking

wesm:
Pushed 782407fd addressing the review above.

**Fixed**
- Size-rejected singleton documents now retry with progressively halved, rune-safe chunk text before giving up, so a token-dense document can no longer block activation forever (counted in `RunResult.Truncated`; test added).
- Chat body_html fallback now uses one blank-text predicate on both sides — the SQL trims the full ASCII whitespace set (plain `TRIM` only strips spaces, which turned out to be a second variant of the same bug) and the Go scan mirrors it exactly (test added).
- PG participant digest renders via `to_char(... AT TIME ZONE 'UTC', ...)` instead of GUC-dependent `CAST(timestamptz AS TEXT)`, and the expression is now a shared `store` constant used by both the coverage CAS and the assembler, so the copies can't drift.
- Scheduler runs one convergence check per tick; a transient check failure falls through to the backstop instead of skipping it, and the post-backstop re-check only runs when the backstop actually ran.
- Eval harness document cap pinned to the production 100,000-byte limit.
- Removed the dead `vectorFeatures.Worker` alias; `testing.Short()` guard on the go-build eval integration test; documented the clock-page write cost in `UpsertMessage`.

**Deliberately not changed**
- Coverage-CAS-miss error propagation: four existing tests pin this as a designed invariant (the error keeps the tick non-converged and the journal watermark pinned); it self-heals next tick, so softening it wasn't worth re-deriving the crash invariants.
- The sqlitevec/pgvector publisher duplication, worker function-length extractions, and `NewEngine(any)` signature — maintenance-shaped refactors better done as follow-ups than inside this PR.

Verified with build, vet, and the SQLite-backed suites for vector/store/scheduler/cmd/eval (all green; scheduler also 10× under `-race`). PG-tagged suites still need a `make test-pg-both` run.

wesm:
Pushed 0c1df827 for the roborev finding: the truncation fallback now carries the accepted truncated chunk texts into publication — each chunk's source span shrinks to the embedded prefix (same tail-layout math as `trimOwnedChunkText`) and is flagged `truncated`, so excerpts can't cover text the vector never saw. Test extended to pin the stored spans. The earlier `nix-build` failure was a transient bun tarball download error (`FailedToOpenSocket` fetching `kit-ui`; same pin passed on the previous commit) — re-run triggered.

wesm:
CI status: `run / test` is green as of 2bb44b90 (the earlier failure was a testifylint violation in the new truncation test, fixed). The remaining `nix-build` failure is a repo-wide regression hitting every fork PR since #602 — bun cannot resolve the `github:kenn-io/kit-ui` dependency offline in the nix sandbox. Root-caused and fixed in #604 (verified green on a sandboxed `ubuntu-latest` runner); once that merges, a rebase or re-run here should go fully green.

PR #592, opened 2026-08-09
feat(people): add typed temporal relationships

salmonumbrella:
## What changed

- Add seeded and user-defined relationship types with forward and reverse labels, symmetric-type support, presentation metadata, and optional vCard `RELATED` mappings.
- Store one canonical, temporal edge per person relationship with partial start and end dates, history, notes, provenance, confidence, optimistic concurrency, and active-duplicate protection.
- Render the correct label from either person's view without mirrored rows, and keep ended relationships available as history.
- Resolve imported `RELATED` values only by exact person UID; unresolved values enter a review queue instead of creating people automatically.
- Expose the model through equivalent SQLite and PostgreSQL stores, the daemon HTTP/OpenAPI clients, and new relationship CLI commands. A graph UI remains out of scope.

## Why

Msgvault can preserve rich person profiles, but it cannot yet represent how two people are connected or how that connection changes over time. Copying labels onto profiles loses direction, history, and source evidence, while guessing from imported names or contact values risks linking the wrong people.

## Usage

```sh
msgvault relationship-type list
msgvault person relationship add 12 partner 34 --from 2024-03
msgvault person relationship list 12
msgvault person relationship end 57 2026-08-09
msgvault person relationship reviews --status pending
```

Refs #534

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1579,6 +1579,30 @@ components:
         - scope_policy
         - normalization
       type: object
+    CreatePersonRelationshipRequest:
+      additionalProperties: false
+      properties:
+        end_date:
+          type: string
+        notes:
+          type:
+            - string
+            - "null"
+        relationship_type_slug:
+          type: string
+        source_person_id:
+          format: int64
+          type: integer
+        start_date:
+          type: string
+        target_person_id:
+          format: int64
+          type: integer
+      required:
+        - source_person_id
+        - target_person_id
+        - relationship_type_slug
+      type: object
     CreatePersonRequest:
       additionalProperties: false
       properties:
@@ -1587,6 +1611,38 @@ components:
           type: integer
       required:
         - participant_id
+      type: object
+    CreateRelationshipTypeRequest:
+      additionalProperties: false
+      properties:
+        color:
+          type:
+            - string
+            - "null"
+        description:
+          type:
+            - string
+            - "null"
+        forward_label:
+          type: string
+        icon:
+          type:
+            - string
+            - "null"
+        is_symmetric:
+          type: boolean
+        reverse_label:
+          type: string
+        slug:
+          type: string
+        vcard_related_type:
+          type:
+            - string
+            - "null"
+      required:
+        - slug
+        - forward_label
+        - reverse_label
       type: object
     CreateRequest:
       additionalProperties: false
@@ -3664,6 +3720,17 @@ components:
         label:
           type: string
       type: object
+    PatchPersonRelationshipRequest:
+      additionalProperties: false
+      minProperties: 1
+      properties:
+        end_date:
+          type: string
+        notes:
+          type:
+            - string
+            - "null"
+      type: object
     PatchPersonRequest:
       additionalProperties: false
       properties:
@@ -3673,6 +3740,23 @@ components:
             - "null"
       required:
         - display_name
+      type: object
+    PatchRelationshipTypeRequest:
+      additionalProperties: false
+      minProperties: 1
+      properties:
+        color:
+          type: string
+        description:
+          type: string
+        forward_label:
+          type: string
+        icon:
+          type: string
+        reverse_label:
+          type: string
+        vcard_related_type:
+          type: string
       type: object
     PatchSavedViewRequest:
       additionalProperties: false
@@ -4447,6 +4531,112 @@ components:
         names:
           $ref: "#/components/schemas/PersonNamePatchRequest"
       type: object
+    PersonRelationship:
+      additionalProperties: true
+      properties:
+        confidence:
+          format: double
+          type: number
+        created_at:
+          format: date-time
+          type: string
+        created_by:
+          type: string
+        end_date:
+          $ref: "#/components/schemas/PartialDate"
+        forward_label:
+          type: string
+        id:
+          format: int64
+          type: integer
+        is_symmetric:
+          type: boolean
+        notes:
+          type: string
+        relationship_type_id:
+          format: int64
+          type: integer
+        reverse_label:
+          type: string
+        revision:
+          format: int64
+          type: integer
+        source:
+          type: string
+        source_person_id:
+          format: int64
+          type: integer
+        source_ref:
+          type: string
+        start_date:
+          $ref: "#/components/schemas/PartialDate"
+        status:
+          type: string
+        target_person_id:
+          format: int64
+          type: integer
+        type_slug:
+          type: string
+        updated_at:
+          format: date-time
+          type: string
+        updated_by:
+          type: string
+        vcard_identity:
+          $ref: "#/components/schemas/VCardIdentity"
+      required:
+        - id
+        - source_person_id
+        - target_person_id
+        - relationship_type_id
+        - type_slug
+        - forward_label
+        - reverse_label
+        - is_symmetric
+        - status
+        - source
+        - vcard_identity
+        - created_by
+        - updated_by
+        - revision
+        - created_at
+        - updated_at
+      type: object
+    PersonRelationshipView:
+      additionalProperties: true
+      properties:
+        counterpart_display_name:
+          type: string
+        counterpart_label:
+          type: string
+        counterpart_person_id:
+          format: int64
+          type: integer
+        counterpart_vcard_uid:
+          type: string
+        direction:
+          type: string
+        relationship:
+          $ref: "#/components/schemas/PersonRelationship"
+      required:
+        - relationship
+        - direction
+        - counterpart_person_id
+        - counterpart_label
+        - counterpart_vcard_uid
+      type: object
+    PersonRelationshipsResponse:
+      additionalProperties: true
+      properties:
+        relationships:
+          items:
+            $ref: "#/components/schemas/PersonRelationshipView"
+          type:
+            - array
+            - "null"
+      required:
+        - relationships
+      type: object
     PersonSearchHTTPResponse:
       additionalProperties: true
       properties:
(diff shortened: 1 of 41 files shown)


wesm:
Reviewed the store/schema layer, API layer, CLI, and generated artifacts. Verified locally: build, full test suite with `fts5 sqlite_vec` tags, `make openapi-check` (generated client and specs are byte-reproducible), and all CI checks green. The core design held up well — canonical edge normalization, optimistic concurrency, partial-date comparison, and SQLite/PG parity all checked out. One bug worth fixing before merge.

**Major**

- **Seed reconciler can permanently break `InitSchema`** (`relationship_types.go:344` + `relationship_type_seed.go:209`): `UpdateRelationshipTypeContext` allows editing `vcard_related_type` on *system* types, but the seed reconciler rewrites that column on every `InitSchemaContext`. Confirmed by execution: clear the seeded `friend` mapping via the API, create a custom type claiming `vcard_related_type = "friend"`, and every subsequent startup fails with a UNIQUE violation inside `InitSchemaContext` — no API path to recover. Even without the collision, accepted edits to system-type mappings are silently reverted (with a revision bump) on next open. Suggest rejecting `VCardRelatedType` edits on `ownership = 'system'` rows.

**Minor**

- **PATCH relationship-type is fail-open on miscased keys** (`internal/api/person_relationships.go:235`): Go decodes JSON field names case-insensitively, so `{"Forward_Label":"x"}` passes `DisallowUnknownFields` and the min-one-field gate but leaves all update pointers nil — returns 200 with a revision bump while changing nothing. The sibling `handlePatchPersonRelationship` fails closed on the same input.
- **Seeding races across processes** (`relationship_type_seed.go:149`): SELECT-then-INSERT without conflict handling; two processes initializing the same PostgreSQL DB can race, failing the loser's `InitSchemaContext`. Transient, and mirrors the existing attribute-seed pattern.
- **`person_relationships.status` is unconstrained** (`schema.sql:786`): fully derivable from `end_year` but no CHECK ties them together, despite the table comment promising integrity for writers that bypass Go. Add the CHECK or drop the column.
- **Date asymmetry in the spec** (`api/openapi.yaml:1585`): request `start_date`/`end_date` are bare strings with no pattern/description, while responses use `PartialDate` objects — fetched dates can't round-trip into a PATCH body, and accepted formats are only discoverable from 400 error text.
- **CLI `relationship-type create` omits `--description`/`--vcard-related-type`** (`person_relationship.go:569`): the API accepts both on create and `update` exposes them, so a described type takes two commands and two revisions.

**Nits**

- `relationship_types.go:332`: dead nil-check on an `any` holding `sql.NullString`; `%v` renders conflict errors as `{friend true}` instead of `friend`.
- `person_relationships_test.go:146` hard-codes the seed count (19); the next seeded type breaks an unrelated HTTP test.
- `status`/`direction`/`source` response fields lack `enum` in the spec, unlike the review `status` query param.
- Test gaps: "end date before stored start date" via End/Patch is never exercised; no CLI test covers `relationship reviews` or delete confirmations.
- `relationshipIfMatch` duplicates `personIfMatch` modulo the tag prefix; `relationship end` takes the date positionally while `add` uses `--until`.

wesm:
looking

wesm:
Addressed both roborev findings in 775ae52e:

- **Symmetry drift**: restoring `is_symmetric` on a drifted symmetric seed now also restores the seed labels when they were pulled apart out-of-band, so reconciliation can no longer fail the symmetric-label CHECK and abort `InitSchema`. Label updates on seeded types additionally validate against seed-defined symmetry (not just the stored flag). Regression test covers drift → rejected API label update → direct-SQL label drift → successful repair.
- **RELATED orientation**: imports now store the matched person as the typed source (`RELATED;TYPE=parent` naming bob on alice's card records bob as alice's parent, per RFC 6350 §6.6.6), in both automatic resolution and review acceptance. Added orientation tests for an asymmetric auto-resolved import and an accepted review.

wesm:
Addressed the roborev finding in f0c1438b: automatic RELATED resolution now looks up the occurrence's staged review inside the linking transaction before creating an edge.

- A **rejected** occurrence returns its review and never auto-links, even after the person or type mapping later becomes resolvable.
- A **pending** review is claimed as accepted and pointed at the new edge (`accepted_relationship_id`) atomically with the edge write.
- An **accepted** review returns its recorded edge; if that edge was deleted since (`accepted_relationship_id` is `ON DELETE SET NULL`), re-import does not resurrect it.

Regression tests cover rejection preserved across re-import, pending → accepted transition with idempotent re-import, and the deleted-edge case; both were confirmed failing before the fix.

wesm:
Addressed the roborev finding in 12d27a3e: automatic resolution now inserts an already-accepted review row (`status = accepted`, `matched_person_id`, `accepted_relationship_id`, `reviewed_by`) in the same transaction as the edge, using the occurrence unique index's `ON CONFLICT DO NOTHING` for concurrent-import safety.

Every decided occurrence now has exactly one ledger row, so deleting an automatically imported relationship is as durable as deleting a human-accepted one — the surviving accepted row with its `ON DELETE SET NULL`-cleared edge pointer blocks resurrection on re-import. Table comments in both schemas updated to describe the ledger contract.

Regression test covers auto-link → ledger row assertions → edge deletion → re-import returns the tombstoned review and creates no edge; confirmed failing before the fix.

wesm:
Addressed the roborev finding in 4747b771. Relationship data now rides the existing `IncludeProfiles` opt-in rather than a new flag — relationships are person-profile data and that flag is already the documented gate for profile values and their provenance; default subsets still exclude everything relationship-related.

Under `--include-profiles`, subset copies now:
- mirror the source's relationship type catalog into the destination by `universal_id` (upsert over the destination's own seeds, with local ids and `inverse_type_id` remapped, and vCard mappings mirrored without transient unique collisions),
- copy all edges — active and history — whose endpoints both crossed the subset boundary, and
- preserve the `RELATED` decision ledger, clearing `matched_person_id` / `accepted_relationship_id` when their referent didn't cross the boundary, exactly as the live schema clears them on deletion — so deletion tombstones survive and re-import into the subset cannot resurrect a relationship deleted in the source.

Older source databases without the relationship tables are skipped via the same `sourceTableExists` guard the profile copier uses. Behavioral tests cover the profile-subset copy (user-defined type reconciliation, seeded-type edge remapping, tombstone honored on re-import into the subset) and the default-exclusion guard; the preservation test was confirmed failing before the fix.

wesm:
Addressed both roborev findings in 0cda601d by restructuring `resolveMatchedRelatedValue` so the ledger row is established before the edge is written:

- **Aborted-transaction duplicate recovery**: the edge INSERT now runs under a savepoint (`SAVEPOINT`/`ROLLBACK TO`/`RELEASE`, dialect-neutral). PostgreSQL aborts a transaction whose statement raised a unique violation, so recovery rolls back to the savepoint before looking up the existing active edge — a second occurrence resolving to an already-linked pair now reuses the edge on both backends.
- **Occurrence race**: a first-seen occurrence inserts its accepted ledger row up front via the occurrence unique index (`ON CONFLICT DO NOTHING RETURNING`); losing the insert race re-reads the winner's committed row and honors its status (settled decisions stand, a pending row is claimed). A pending review is now claimed **before** edge creation, so its row lock serializes concurrent resolvers, and a claim loser re-reads and honors the committed decision. No committed state shows an edge alongside a pending or rejected review.

New tests cover second-occurrence edge reuse with its own accepted ledger row, and concurrent first imports converging on one edge with exactly one ledger row. Both defects only manifest on PostgreSQL (SQLite doesn't abort on constraint errors and serializes writers), so the SQLite runs guard behavior while the `test-postgres` CI lane exercises the actual failure modes.

wesm:
Addressed both roborev findings; final state is 2bac4e69 (0cda601d restructure + a nilnil lint fix). `resolveMatchedRelatedValue` now establishes the ledger row before writing the edge:

- **Aborted-transaction duplicate recovery**: the edge INSERT runs under a savepoint (dialect-neutral `SAVEPOINT`/`ROLLBACK TO`/`RELEASE`). PostgreSQL aborts a transaction whose statement raised a unique violation, so recovery rolls back to the savepoint before looking up the existing active edge — a second occurrence resolving to an already-linked pair now reuses the edge on both backends.
- **Occurrence race**: a first-seen occurrence claims its accepted ledger row up front via the occurrence unique index (`ON CONFLICT DO NOTHING RETURNING`); a lost insert race re-reads the winner's committed row and honors its status. A pending review is claimed **before** edge creation so its row lock serializes concurrent resolvers, and a claim loser honors the committed decision. No committed state can pair an edge with a pending or rejected review.

New tests: second-occurrence edge reuse with its own accepted ledger row, and concurrent first imports converging on one edge with exactly one ledger row. Both defects only manifest on PostgreSQL, and the `test-postgres` CI lane is green on this code.

wesm:
Addressed both roborev findings in 1a4bdc75:

- **Deferred-transaction snapshot busy**: `ResolveRelatedValueContext`'s two write paths (matched resolution and review staging) now retry the whole transaction on busy/serialization errors with a short growing backoff, following the `StartSyncContext` precedent. A deferred SQLite transaction that read the ledger before its first write can fail its lock upgrade with `SQLITE_BUSY_SNAPSHOT` when a concurrent resolver commits first; a fresh attempt takes a new snapshot and observes the winner's decision. Both operations are idempotent by construction (ON CONFLICT occurrence claims, duplicate edge reuse), so re-execution is safe. `IsBusyError` matches the primary `SQLITE_BUSY` code, which covers the `BUSY_SNAPSHOT` extended code; the PG side maps to lock/deadlock SQLSTATEs. The backoff matters: I reproduced the failure under heavy machine load, where five immediate retries all landed inside the winning transaction's still-open window.
- **Accept vs existing edge**: accepting a pending review whose asserted relationship already exists as an active edge now attaches the review to that edge via the same savepoint-protected add-or-reuse helper automatic resolution uses, instead of failing with a duplicate error and stranding the review pending. Non-duplicate edge failures (e.g. unknown type) still roll the claim back atomically — that test now proves rollback with an unknown type slug.

New tests: accept-attaches-to-existing-edge, and a direct contract test for the retry helper (transient busy retries, other errors pass through, persistent busy stops at the cap) since the snapshot race can't be provoked deterministically through the public API. The concurrent-import test now passes repeatedly under load.

PR #593, opened 2026-08-09
feat(people): add organizations and employment records

salmonumbrella:
## What changed

- Add first-class organizations with versioned profiles, typed custom attributes, lifecycle and merge operations, and reviewable duplicate suggestions.
- Add current and historical employment records with primary-employment constraints and read-time person and vCard projections.
- Keep SQLite and PostgreSQL behavior aligned, and expose the model through daemon APIs, generated clients, and `organization` and `employment` CLI commands.

## Why

People can have several current or historical roles, but copied company and title fields cannot preserve that history or represent one organization consistently across people. First-class records make employment changes queryable without losing source evidence.

## Usage

```sh
msgvault organization create "Example Corp" --domain example.com
msgvault employment add --person 42 --organization 7 --title "Engineer" --start 2024-01 --primary
```

Refs #534


wesm (changes requested):
Thanks — the core engineering here is solid (dual-backend concurrency, merge lock ordering, partial dates are all handled carefully, and the tests exercise real DB/HTTP paths). But there are two data-loss-class bugs to fix, and the PR is carrying roughly a third more code than the feature needs. Requesting changes.

**Blocking bugs**

- `employment set` builds the PATCH body from changed flags only, but the server PATCH is full-replace (`cmd/msgvault/cmd/employment.go:63`). Fixing a title on a past employment nulls role, department, and dates, and flips it back to `is_current=true`. `organization set` already does the correct read-merge-write; `employment set` should too. There's no test for `set`.
- `PATCH /organizations/{id}` un-retires on any field-only update: `Retired` is a non-pointer `bool` with full-replace semantics (`internal/api/organizations.go:40`). Renaming a retired org silently reactivates it. The CLI works around this internally, which shows the footgun is live for other clients. Make it `*bool`.
- Employment writes have no SQLite busy-retry, unlike attribute writes (which this PR generalized `retryAttributeWrite` for). Concurrent writes for one person surface raw `SQLITE_BUSY` as an untyped 500, and both concurrency regression tests skip on SQLite, so the default backend has no coverage here (`internal/store/employments.go:423`).

**Behavior surprises to resolve**

- An update with `is_primary` omitted can silently auto-promote that employment to primary, changing the person's derived company as a side effect of a typo fix (`employments.go:453`).
- `is_current=true` can coexist with a past `end_date`; nothing validates consistency (`employments.go:397`).
- Profile GET opens a write transaction with `FOR UPDATE` where the person analog uses a read snapshot; on Postgres, reads serialize behind merges (`organization_profile.go:161`).
- An invalid `primary_domain` is silently dropped instead of erroring; the profile-identifier path errors for the same condition (`organizations.go:523`).

**Scope: please trim**

- `organization_names` copies 13 person-name columns the code never reads (organizations now have `honorific_prefixes`), and `organization_schema_parity_test.go` enforces the copy. Strip the unused columns and the test that freezes them in.
- The duplicate-suggestions + merge workflow (~1,000 lines across store/API/CLI) goes beyond "organizations and temporal employment associations" — the roadmap lists same-name suggestions as bonus research. Suggest a follow-up PR.
- Dead surface: `UpdateOrganizationContext`/`RetireOrganizationContext`/`UnretireOrganizationContext` have no non-test callers; `PrimaryCurrentEmploymentsContext` is uncalled; `SupersedeOrganizationAttributeValueContext` is wired in the serve adapter but no route calls it (so org attributes can't be cleared over HTTP); `PUT /organizations/{id}/profile` has no consumer; `resolved_by` is never written and tests assert it's always nil.
- `organization_attributes.go` is a ~90% copy of `person_attributes.go`, and the copies have already drifted (different error sentinels and locking for the same operation). The `retryAttributeWrite` extraction in this PR shows the parameterization works — please share the rest.
- CLI: about 10 of the 22 subcommands aren't needed for the advertised workflow (org attributes, the duplicates trio, merge, and `organization employments`, which duplicates `employment list --organization`).

**Smaller items**

- `organization duplicates resolve` derefs `resp.JSON200.Status` without the nil guard every other subcommand has (`organization.go:476`).
- `getCLIOrganization` has a fallback for "older daemons" that can't exist — this PR introduces the endpoint — and the only test for the delete/retire path exercises that dead fallback.
- Org endpoints reuse `decodePersonRequest`, so malformed org bodies say "Invalid person request".
- Flag inconsistency: `--end` on add/set vs `--end-date` on end.
- Merge leaves open duplicate suggestions pointing at the merged org, unconditionally nulls `employments.address_id`, and allows merging a live org into a retired one.

With the bugs fixed and the trims above, this drops from ~10.7k to roughly 6-7k hand-written lines without losing anything the description promises.

wesm:
looking

wesm:
Pushed ec139bde addressing the review (note: the PR head had moved to 7c31adf4, which already added the org-attribute DELETE route and profile history endpoint — the fixes are rebased on that).

**Fixed**
- `employment set` now read-merge-writes; unspecified flags come from the fetched record, and `--person`/`--organization` no longer need re-stating. Covered by a new CLI test.
- `PATCH /organizations/{id}` `retired` is now a nullable boolean; omitting it preserves lifecycle state (regression test added for the rename-reactivates case).
- Employment writes retry on SQLite busy errors via the shared helper (renamed `retryContendedWrite`), with a concurrency test on the SQLite path.
- Updates no longer auto-promote to primary when `is_primary` is omitted; `is_current=true` + `end_date` is rejected; invalid `primary_domain` errors instead of vanishing; profile GET uses a read snapshot; merging into a retired org is rejected; org endpoints no longer emit "Invalid person request".

**Trimmed**
- Duplicate-suggestion workflow removed end to end (store, schema, API, CLI, ~1,900 lines with generated code) — better as a follow-up PR.
- The 13 person-only name columns dropped from `organization_names`; the parity test now asserts they are absent rather than required.
- Dead surface removed: `UpdateOrganizationContext`/`RetireOrganizationContext`/`UnretireOrganizationContext`, batch `PrimaryCurrentEmploymentsContext`, the bare-Organization CLI fallback (and the test that encoded it), and the redundant `organization employments` subcommand. `--end-date` unified to `--end`.

**Kept, deliberately**
- `merge` stays: `merged_into_id` redirects are woven through the schema, employment locking, and regression tests; removing it would be a much larger change than the review's intent.
- `PUT /organizations/{id}/profile` stays: on the current head the profile envelope is the GET representation with a history endpoint, so this is the model's only write path rather than dead surface.
- The `organization_attributes.go`/`person_attributes.go` unification is deferred: beyond the now-shared retry helper, the remaining diff includes real behavior differences (owner locking, error sentinels) and per-entity public types, so collapsing it means touching shipped person behavior — follow-up material.

Net effect: −2,153 lines against the previous head. `go vet`, the testify-helper check, and the store/api/cmd/vcardmap/daemonclient suites pass locally (the only failures are two FTS5 environment-dependent tests that fail identically on main).

wesm:
Pushed 85f2271d addressing all four roborev findings — each verified against the code before fixing, and each has a regression test:

- **Person deletion guard**: `DeletePersonContext` now also counts active `organization_attribute_values` record references (`value_record_type = 'person'`), so a person referenced by an organization attribute cannot be deleted; superseded references still allow deletion. Test mirrors the existing person-side guard test.
- **Historical employment resurrection**: `UpdateEmploymentContext` preserves the stored `is_current` when both `is_current` and `end_date` are omitted; supplying an end date still ends the employment. Test updates a historical record's title and asserts it stays historical.
- **`employment set` field preservation**: the merge body now carries `address_id`, `source_ref`, and `confidence` from the fetched record. `address_id` is cleared only when `--organization` changes (the address belongs to the previous org), and `source_ref`/`confidence` only when `--source` changes (they describe the previous provenance). The preservation test now asserts all three round-trip.
- **Organization media content**: added `GET /organizations/{id}/profile/media/{media_id}/content` with `ReadOrganizationMediaDataContext`, mirroring the person profile media endpoint (same headers, 404 for URI-only values). Test round-trips inline logo bytes and covers the URI-only and missing-ID 404s.

OpenAPI specs, Go client, and web schema regenerated; store/api/cmd suites pass locally including the artifact-staleness checks.

wesm:
Pushed d741dfa2 addressing both findings — verified against the code first:

- **Envelope metadata round-trip**: confirmed `organizationEnvelopeMatches` compares `type_label`, `type_tokens`, `confidence`, and `active_from`, none of which `OrganizationEnvelopeBody` could express — so any row carrying them was superseded and reinserted stripped of metadata on every PUT. The body now exposes all four (the doc comment states the invariant: the body must cover every field the reconciliation compares). The new HTTP test PUTs a name with all four fields, then makes an unrelated update (adds a category) re-sending the name, and asserts the row's envelope ID is unchanged with metadata intact.

- **Values stranded on merge redirects**: confirmed the redirect is immutable after merge (both attribute write paths reject merged organizations), so active profile rows and attribute values left on the losing organization could never be corrected, and an active person record reference would block that person's deletion permanently. `MergeOrganizationsContext` now supersedes the losing organization's active rows across all six profile tables (reusing `supersedeOrganizationRowsTx`) and its attribute values before marking it merged. I chose supersede over migrate deliberately: blindly moving the losing organization's facts onto the survivor risks planting wrong or duplicate data there, while superseded rows remain fully readable as history on the redirect and anything that belongs on the survivor can be curated there explicitly — same philosophy as the existing former-name retention. The test merges an organization holding a profile and a person record reference, then asserts the redirect has no active values, the history remains readable, and the referenced person is deletable.

OpenAPI specs, Go client, and web schema regenerated; store/api/cmd suites pass including artifact-staleness checks.

wesm:
Pushed 28cdbffb addressing both findings plus the CI failure:

- **Subset completeness**: employment data now crosses the subset boundary under the existing opt-ins. `IncludeProfiles` copies the employments of copied people, the organizations they reference, and those organizations' profile rows (contact points go through the communication-service remap). `IncludeAttributes` copies organization attribute definitions and values with the same `universal_id` mapping and person record-reference boundary as person attributes. Default subsets still exclude all of it — employment history is the same sensitivity class as structured profiles. Behavioral tests cover both the opted-in copy and the default exclusion. One boundary note: employments cannot reference merged organizations, so the copied organization set needs no merge-redirect closure.
- **Survivor deletion**: `DeleteOrganizationContext` now counts inbound `merged_into_id` redirects and returns a typed `ErrOrganizationInvalid` ("N merged organization(s) redirect here") instead of letting `ON DELETE RESTRICT` surface as a 500. Test extends the existing redirect-immutability test.
- **CI**: the `run / test` lint failure was a `wastedassign` on the `retired` declaration — fixed. The `run / nix-build` failure is not from this PR: it's the sandboxed web build issue fixed on main by #607 (which landed about an hour after that run started); the new CI run against the updated merge picks up the fix.

wesm:
Pushed 67cd031a addressing the media-retention finding — verified first: the reconciliation hashed only the request's inline bytes, and since reads expose media metadata without the bytes, any GET-derived PUT computed an empty hash, mismatched the stored row, and superseded it with a URI-only copy.

The fix is the retention-identifier approach: `OrganizationMediaBody` and `OrganizationMediaInput` now carry `content_hash`. A media row sent with `content_hash` and no `data` matches the stored row carrying that hash and keeps its bytes — and since GET already returns `content_hash` on every media row, a GET-derived payload round-trips naturally. Two guardrails close the ambiguity between omission and removal: a retention hash that matches no active row is rejected as `invalid_organization` (never inserted as a hash-without-content row), and `data` sent with a mismatching `content_hash` fails validation. Omitting both still means URI-only, so explicit removal keeps working.

The HTTP regression test writes URI-plus-inline media, re-sends the GET-shaped row alongside an unrelated category addition, and proves the row is retained (same envelope ID, `has_data` still true, bytes still downloadable via the content endpoint); a stale hash returns 400.

OpenAPI specs, Go client, and web schema regenerated; store/api/cmd suites pass including artifact-staleness checks.

wesm:
Pushed 7188c613. Confirmed: the media validation required `uri` or `data`, so inline-only media — which reads back as `content_hash` with no bytes and no URI — was rejected on a GET-derived PUT. A non-empty `content_hash` now satisfies the requirement (the retention path then matches it against the stored row as before, and a hash matching no active row is still rejected at insert). The new HTTP test round-trips inline-only media through an unrelated update and proves the row and its downloadable content survive.

Store-side validation only — no API schema change, so no regeneration was needed.

wesm:
Pushed 8deee645 (+ a gofmt follow-up). Confirmed the limitation: the retention hash only survived an exact match — any edit to media_type, uri, or an envelope field superseded the active row and the hash-only replacement then failed at insert for lacking data, so metadata edits required re-uploading the bytes.

As suggested, retention hashes are now resolved against the active media rows before reconciliation: the stored bytes are loaded into the input, so when an edit supersedes and reinserts the row, the replacement carries the content. A hash with no active match is still rejected as `invalid_organization` (the insert-time guard remains as defense in depth). Fingerprints stay consistent because the stored hash is by construction the hash of the loaded bytes.

The HTTP regression test edits an inline media row's type and URI sending only the retention hash, and proves the replacement row keeps `has_data` and its downloadable content.

Store-side only again — no API schema change, no regeneration needed. Store/api suites pass.

wesm:
Pushed 8c54234f. Confirmed: the `IncludeIdentity` closure's reference edges only walked `person_attribute_values`, so a person referenced by an included person's employer (an organization attribute) never entered the identity set — and the org-attribute copy's dangling-reference filter then silently dropped the value.

The reference-edge CTE now adds a second edge: employments of included people → their organizations' person-valued attribute values → the target's participants. It's gated on the source schema having the employment tables (older sources skip it) and on both `IncludeProfiles` and `IncludeAttributes`, since organizations only cross the subset boundary through employment references with attributes opted in.

The regression test builds an off-message person referenced by an included person's employer: with all three opt-ins the target person and the referencing attribute value survive the copy; without `IncludeIdentity` the target stays excluded and the reference is dropped rather than dangling, matching the person-attribute policy.

wesm:
Force-pushed d14e7aa8: all review follow-ups squashed into the single feature commit and rebased onto current main (15c8f2cf), authored as the original PR author with co-author trailers.

Also fixes the last roborev finding: `employment set` with `--not-current` or `--end` on a primary employment now demotes it (matching the server-side end operation) instead of sending an invalid historical primary, and an explicit `--primary` combined with `--not-current` or `--end` is rejected as a usage error. Both cases have CLI tests.

Rebase notes — main's typed person relationships (#592) landed on the same seams, resolved by keeping both features:
- subset export now copies relationships *and* employment/organization data under `IncludeProfiles`
- the serve adapter carries both route sets, with the wiring tests for each
- `APISchemaVersion` moves to 1.40.0 (main took 1.39.0 for relationships); contract assertions updated
- a duplicate `formatCLIPartialDate` (added independently by both branches) deduplicated

OpenAPI specs, Go client, and web schema regenerated against the merged surface. On the rebased tree: `go vet` clean, and the store, api, cmd, vcardmap, and daemonclient suites pass (the only failure is the FTS5 environment test that fails identically on main).

wesm:
Pushed 4731c7c1 addressing both findings and the CI lint failure:

- **Value duplicates with distinct identity**: confirmed the prepare-time duplicate checks for addresses, contact points, and media keyed only on business value, and the reconciliation's business-key map was single-valued — so an imported profile carrying the same value under two PROP-IDs could not round-trip. The duplicate check now appends an envelope discriminator (durable vCard identity, ordinal, TYPE label/tokens), and reconciliation keeps a list of current rows per business value, matching each desired row to an unclaimed counterpart — preferring the one whose full envelope matches so every row keeps its own identity. The regression test round-trips two identical addresses under distinct PROP-IDs and two identical media blobs under distinct ordinals, asserting stable row IDs across an identical replacement. Names, identifiers, and categories intentionally keep strict uniqueness, matching their unique indexes.
- **Merge busy-retry**: `MergeOrganizationsContext` now runs under the shared bounded busy/deadlock retry (`retryContendedWrite`), same as employment and attribute writes. Each attempt is a fresh transaction that re-reads both roots, so a concurrent commit produces a typed revision conflict rather than a raw `SQLITE_BUSY`. One deviation from the suggested fix: revisions are *not* reloaded server-side across attempts — they're the caller's compare-and-swap intent, and refreshing them would silently merge over a concurrent change.
- **CI lint (goconst)**: the merge with the relationships feature pushed the `"ETag"` and `"int64"` literals in `internal/api` over the threshold; both are now package constants.

Store and api suites pass, vet clean.

wesm:
Pushed 7dec1310. Confirmed: `employment list --organization` printed the organization ID on every row — the one value the caller already supplied — with no way to identify the employee. The counterpart column now follows the scope: person-scoped listings show `ORGANIZATION`, organization-scoped listings show `PERSON`. The new CLI test lists an organization with two employees and asserts the header and per-row person IDs.

On the CI failure: `run / test` died in the `govulncheck` step because `vuln.go.dev` returned 403 Forbidden while fetching the vulnerability database — transient infrastructure on Google's side, not a code issue. This push retriggers the run.

wesm:
Pushed b055a8a6. Confirmed: both employment list endpoints accept `limit`/`offset` but the CLI never sent them, so listings silently capped at the server's default page of 200. `employment list` now has `--limit` and `--offset`, forwarded to both the person- and organization-scoped requests — same flag shape as `organization list`. The organization-scoped test now asserts the parameters reach the server.

wesm:
Pushed 456f75fa addressing both findings — verified first: `OrganizationAttributeValueInput` supports `active_from`/`active_until` and the person-side API body already exposes them, but the organization body dropped both; and the body had `ordinal` while the CLI offered no way to set it.

- `SetOrganizationAttributeBody` now carries `active_from` and `active_until`, forwarded to the store. The HTTP test covers a backdated write (stored `active_from` matches the request, not the write time) and a fully historical write (`active_from`+`active_until` in the past) that lands as history without becoming the current value.
- `organization attribute set` gains `--ordinal` (validated non-negative, forwarded through the body), so multi-valued attributes can be updated in place and `--expected-value-id` targets the intended ordinal instead of conflicting against a fresh one. The CLI test asserts the ordinal reaches the request body.

OpenAPI specs, Go client, and web schema regenerated; api/cmd suites and the artifact-staleness checks pass.

wesm:
Pushed b7f06be6 for the CLI finding, and the CI failure is a repo-wide issue with a separate fix:

**CLI dates and reactivation**: confirmed — an empty `--start`/`--end` produced a non-nil empty string the server rejects, and nothing could flip `is_current` back on. `employment set` now has `--clear-start`/`--clear-end` (omit the date from the full-replace body) and `--current` (reactivate an ended employment). Guard rails: empty `--start`/`--end` fails fast pointing at the clearing flag, `--current` conflicts with `--not-current`, and `--current` with a retained end date asks for `--clear-end` instead of sending a state the store rejects. Tests cover the reactivation body and all four rejection cases.

**CI**: this `govulncheck` failure is not from the PR — Go 1.26.6 shipped stdlib security fixes (GO-2026-6089/6090/6091/6218), so every branch building with 1.26.5 now fails the vuln check, including main's next run. Fix opened separately as #614 (go.mod toolchain bump + nix flake pin); once that merges, this PR's CI goes green on re-run.

wesm:
I'm dropping the SQL enum CHECKs in favor of validating in Go (SQLite can't alter constraints like these without rebuilding the table), would have liked to have caught these earlier and prevented them from getting into the codebase at all

wesm:
Pushed 3a679fbd: dropped the seven-value `source IN (...)` enum CHECK from all twenty tables per backend (person, organization, employment, relationship, and identity tables alike — this PR's tables mirrored the pattern, so the drop is repo-wide for consistency).

Rationale: a schema CHECK is a compatibility ceiling on SQLite — `CREATE TABLE IF NOT EXISTS` never updates constraints on existing archives and `ALTER` can't modify them, so extending the provenance vocabulary would mean rebuilding every table on every archive. The kinds vocabularies and the communication-service catalog already avoid database enums for exactly this reason, and provenance is validated at the Go boundary (`ParseProvenance`/`Provenance.Valid` plus the API enum tags). Kept: `NOT NULL` and the cross-field invariants (confidence range and declared-source scope, partial-date shape, active windows, merge self-reference).

The four tests that pinned the frozen vocabulary now prove the intended contract instead: the database accepts future provenance values while the store boundary still rejects them.

One legacy caveat: `person_attribute_values` shipped its CHECK in v0.19.1, so archives created on v0.19.1–v0.19.3 keep the frozen vocabulary on that single table until a rebuild migration; everything else lands open.

wesm (approved):

wesm:
Pushed 91e70103. Confirmed: the `create-subset` flag help and runtime warnings described only person data, understating what the opt-ins now export.

- `--include-profiles` help and warning now name employment history and the referenced organizations' profiles, contacts, and media; `--include-attributes` now says person *and organization* attribute values.
- For auditing, `CopyResult` gains `Organizations`/`Employments` counts, printed in the summary when `--include-profiles` is set. The employment subset test asserts both counts.

PR #596, opened 2026-08-10
feat(beeper): add reviewable identity candidates

salmonumbrella:
## What changed

- Capture Beeper account, bridge, service, scope, stable provider ID, phone, email, and username observations with provenance and history.
- Create evidence-backed identity candidates without merging people by username, phone, email, display name, or shared conversation alone. Only the same stable provider or Beeper user ID can confirm a match automatically.
- Preserve unknown bridge services, durable review decisions, conflicts, and accepted-link replay diagnostics across repeated imports.
- Add authenticated list, accept, and reject endpoints with generated Go and web clients, with equivalent SQLite and PostgreSQL behavior.

## Why

Beeper can expose the same address on different services, accounts, and workspaces, and usernames can be renamed or reused. Treating those values as one identity can merge different people, while dropping them loses useful evidence. This keeps the importer lossless and makes ambiguous matches reviewable before they change the identity graph.

## Usage

Run Beeper syncs as before. Observation capture and candidate generation happen during import. Review candidates through the authenticated API:

```text
GET  /api/v1/identity/match-candidates?state=candidate
POST /api/v1/identity/match-candidates/{id}/accept
POST /api/v1/identity/match-candidates/{id}/reject
```

Refs #534


wesm:
looking

PR #597, opened 2026-08-11
feat(vector): support account-scoped embedding builds

elviskahoro:
Add account and collection scoping to embedding builds, including durable `[vector.embed.scope] accounts` configuration for daemon jobs. Scope becomes part of generation identity, so changing the selected accounts requires an explicit rebuild and cannot mix vectors across policies.

Coverage and generation reporting now honor the resolved scope, and the CLI documents repeatable `--account` / `--collection` flags.

Verified with the SQLite-vector end-to-end two-account build path and a local scoped build against a configured embedding endpoint: only the selected account was embedded, and its generation activated successfully.

### DIFF
--- a/cmd/msgvault/cmd/embed.go
+++ b/cmd/msgvault/cmd/embed.go
@@ -12,6 +12,8 @@ var (
 	embedFullRebuild            bool
 	embedYes                    bool
 	embedBackstop               bool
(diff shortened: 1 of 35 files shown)


elviskahoro:
Superseded by a fresh PR containing the scope-safety fixes.

PR #598, opened 2026-08-11
feat(vector): support account-scoped embedding builds

elviskahoro:
Add account and collection scoping to embedding builds, including durable `[vector.embed.scope] accounts` configuration for daemon jobs.

Scope is part of generation identity, preventing vectors from different account policies being mixed. Explicit empty collections fail closed, durable scope configuration resolves identifiers rather than unstable source IDs, and daemon jobs verify account-to-source mappings before embedding.

Coverage and generation reporting honor the resolved scope across SQLite and pgvector. The CLI documents repeatable `--account` and `--collection` flags.

### DIFF
--- a/cmd/msgvault/cmd/embed.go
+++ b/cmd/msgvault/cmd/embed.go
@@ -12,6 +12,8 @@ var (
 	embedFullRebuild            bool
 	embedYes                    bool
 	embedBackstop               bool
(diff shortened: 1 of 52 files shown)


wesm:
looking

wesm:
fixing the CI

PR #602, opened 2026-08-12
fix(nix): embed the production Web UI

### LINKED ISSUE #601: Homebrew ships msgvault without the embedded web UI
## Summary

`msgvault serve` installed from Homebrew starts the API fine but 404s for every web UI route. Looks like it's built from the git tag tarball, which only contains `internal/web/dist/stub.html`, so the binary embeds the stub instead of the Vite output.

`docs/web-ui.md` currently states that the UI "is embedded in every release binary," which is true for the GitHub release artifacts but not for this install path. Unfortunately not error is thrown, the UI just never loads, and nothing in the logs points at the cause.

## Reproduce

```console
$ brew install msgvault
$ msgvault serve
...
  API server: http://127.0.0.1:61268

$ curl -s http://127.0.0.1:61268/
{"error":"not_found","message":"No route matches GET /"}
```

Can be easily verified with:

```console
$ node scripts/check-web-assets.mjs --binary /opt/homebrew/bin/msgvault
web asset validation failed: binary does not embed release asset bytes: .vite/manifest.json
```

For comparison, the official v0.19.3 darwin/arm64 release binary passes:

```console
$ node scripts/check-web-assets.mjs --binary ./msgvault   # from the GitHub release tarball
validated 8 release web assets
```

I also rebuilt v0.19.3 locally with `make web-embed` before `go build`, and the UI loads and populates correctly against a real archive.

## Suggested fix

Would you consider publishing a source tarball with `web/dist` already staged into `internal/web/dist/`, so packagers could just point their `url` at that instead of the auto-generated GitHub archive?

`release.yml` already produces the `release-web-dist` artifact, so it should be a relatively small change. I also noticed this is true for the conda build as well, so this could fix both packagers with a one-line URL change each.

If you'd rather not ship prebuilt assets in a source artifact, the Homebrew formula could be modified to add `depends_on "bun" => :build` and run the frontend build in the formula. I checked, and bun's bottles cover the same six platforms msgvault bottles for, so that is viable. I am happy to open that PR against homebrew-core, but I wanted to check your preference first, since the tarball route seems simpler.

mariusvniekerk:
Nix installs currently compile the checked-in browser stub instead of the production Web UI. The package builds successfully, but users cannot use the browser interface that source and release builds provide.

This change makes the Nix derivation build the locked frontend dependency graph with Bun 1.3.14 before Go compilation. It validates the staged release assets and embeds them in the final binary. The Bun integration stays out of the separate Go module derivation, and the flake continues to evaluate for Linux and macOS on both supported architectures.

Issue #601 also affects the conda-forge package, which builds from the same generated source archive. The included design and implementation plan record that external feedstock work and its binary-level regression check. This pull request changes only the in-repository Nix package.

Closes #601


wesm:
removing the superpowers docs

PR #620, opened 2026-08-14
feat(media): add scoped attachment policies and excluded-media purge

### LINKED ISSUE #618: Add per-account and conversation-scoped attachment policy, with a purge path for excluded media
## The situation

Attachment capture is all-or-nothing per provider. The `[beeper]` table exposes
`media` and `max_media_mb`, and both apply to every connected network at once.
`[slack]` has the same pair at the same workspace-wide scope. Discord and Teams
have no media setting at all.

That scope is too coarse once an archive covers real chat networks. On a large
Beeper archive, roughly 99% of attachment rows, and a similar share of attachment
bytes, come from group chats rather than direct messages. Almost all of it is
large public rooms and channels: memes, forwarded video, sticker packs. The
direct-message attachments — the ones with lasting personal value — are well
under one percent of the volume.

Participant count separates those two populations far better than the group/DM
distinction does. In that same archive a few dozen large rooms hold the large
majority of all attachment bytes, while many thousands of small groups hold a
modest fraction. Small groups are where family and close-friend media lives, so
a plain "drop group attachments" rule throws away the good content along with the
noise. A participant threshold keeps it.

Both predicates are already available at import time. Beeper reports chat type as
`single` or `group`, and the importer already maps that onto its own
`direct_chat` / `group_chat` conversation types. Beeper also returns an explicit
participant total that stays correct when the participant list itself is
truncated, and the importer already re-fetches truncated membership. In practice
the resulting `participant_count` is populated for every Beeper conversation.

`sync-beeper --account <id> --no-media` skips media for one account on one run.
It does not solve this for two reasons.

The daemon scheduler reads config, not per-run flags, so every scheduled sync
uses the single global `media` value. A flag on a manual run cannot express an
ongoing policy.

`--no-media` also writes no pending markers, so a deliberate skip is
indistinguishable from an attachment that was never seen. `backfill-beeper-media`
will not revisit it, and there is no way to ask later what was intentionally not
kept.

`accounts` and `exclude_accounts` do not help either. They gate whether an
account is archived at all, and cannot express "archive this network's messages
but not its media".

## Ask

Add attachment policy at a finer scope than the provider table:

- a per-account `media` and `max_media_mb` override, so one noisy network can
  skip media while the rest of the table keeps it;
- a conversation-scoped policy — `media_scope` of `all`, `direct`, or `none` —
  applied at the point where media download is already gated;
- a participant threshold, so groups above a configured size skip media while
  smaller groups keep it;
- the same scope vocabulary for Slack (channel vs DM) and Discord (guild vs DM),
  and `media` / `max_media_mb` extended to Discord and Teams, which have neither;
- a typed skip marker recording why an attachment was not stored — policy scope,
  participant threshold, per-account policy, size cap, or fetch failure —
  distinct from a failed download, so `backfill-beeper-media` can honour a later
  policy change instead of retrying a deliberate skip forever;
- a purge path that reclaims already-stored media a changed policy would no
  longer collect, with a dry run and an explicit confirmation.

Sketch of the config shape:

```toml
[beeper]
media = true
media_scope = "all"          # all | direct | none
media_max_participants = 20  # 0 = unlimited
max_media_mb = 500

[beeper.accounts_config.<accountID>]
media = false
```

Two things constrain the purge path. `internal/deletion` is Gmail-specific today
and deletes at the source; its staged manifest model is the right shape to
generalise, but local-archive-only deletion is a different and safer operation
than source deletion and should stay clearly separated from it. Attachments are
also content-addressed and deduplicated, so one blob can be referenced by both a
group message and a direct message — purge has to drop occurrences and delete a
blob only once nothing references it, then reclaim bytes through the existing
repack path rather than a second mechanism.

## Delivery boundary

This issue owns collection policy and reclamation of media excluded by it. It
does not change how attachments are stored, addressed, or packed, and it does not
touch retrieval. Document extraction and search remain in #608, visual attachment
embeddings in #609, and person-scoped browsing in #612.

One related default is worth deciding alongside this: `max_media_mb` of 100 MiB
silently drops large video, leaving pending markers that every backfill retries
and re-fails. Raising that cap without a scope policy makes the group-chat volume
problem worse, so the two settings are better designed together than separately.

Refs #608, #609, and #612.

Happy to send the PR.

salmonumbrella:
## What changed

- Add a shared attachment policy for Beeper, Slack, Discord, and Teams: `media`, `max_media_mb`, `media_scope` (`all`, `direct`, `none`), and `media_max_participants`, with per-account overrides through `accounts_config` (Beeper account IDs, Slack team IDs, Teams emails) and the existing Discord `guilds` map.
- Record a typed outcome on every attachment occurrence — `pending`, `stored`, `skipped`, or `failed`, plus a skip reason — so a deliberate policy skip is distinguishable from a failed download, and backfills retry only eligible items.
- Add `purge-excluded-media`, which removes already-stored media that current policy excludes. `--dry-run` previews occurrence, blob, and logical byte counts; deletion requires `--yes` or interactive confirmation.
- Keep SQLite and PostgreSQL behavior aligned, preserve the 100 MiB Beeper/Slack/Teams and 50 MiB Discord caps as defaults, and leave configurations with omitted policy keys allow-all.

## Why

Attachment capture is all-or-nothing per provider, so archiving a busy chat network means storing every meme and forwarded video from large public rooms in order to keep the direct-message media that actually matters. `--no-media` on a manual run cannot express an ongoing policy because the daemon scheduler reads config, and it writes no marker, so a deliberate skip is indistinguishable from an attachment that was never seen.

## Usage

```toml
[beeper]
media_scope = "all"
media_max_participants = 12
max_media_mb = 100

[beeper.accounts_config.accountid]
media = false
```

```sh
msgvault purge-excluded-media --dry-run   # preview what current policy excludes
msgvault purge-excluded-media --yes       # remove it
```

Closes #618

### DIFF
--- a/cmd/msgvault/cmd/backfill_beeper_media.go
+++ b/cmd/msgvault/cmd/backfill_beeper_media.go
@@ -3,9 +3,11 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"io"
 	"time"
 
 	"github.com/spf13/cobra"
+	"go.kenn.io/msgvault/internal/beeper"
 )
 
 var backfillBeeperMediaAccounts []string
@@ -14,13 +16,13 @@ func newBackfillBeeperMediaCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "backfill-beeper-media",
 		Short: "Retry pending Beeper attachment downloads",
-		Long: `Retry pending Beeper attachment downloads.
+		Long: `Retry eligible Beeper attachment downloads.
 
-Attachments that failed to download during sync-beeper (Beeper Desktop asset
-temporarily unavailable, over the size cap at the time, transient errors)
-leave a pending marker. This command re-fetches those messages from Beeper
-Desktop and downloads their media into the attachment store. Idempotent:
-already-downloaded attachments are content-addressed and skipped.
+This command retries unfinished downloads and policy exclusions that are now
+allowed, such as media whose configured size cap was raised. It re-fetches
+eligible messages from Beeper Desktop and downloads their media into the
+attachment store. Already-downloaded attachments are content-addressed and
+skipped.
 
 Examples:
   msgvault backfill-beeper-media
@@ -53,10 +55,7 @@ Examples:
 						rebuildCacheAfterWrite(dbPath),
 					)
 				}
-				_, _ = fmt.Fprintf(cmd.OutOrStdout(),
-					"%s: %d messages checked, %d attachments downloaded, %d still pending (%s)\n",
-					accountID, sum.MessagesProcessed, sum.AttachmentsDownloaded, sum.AttachmentsPending,
-					sum.Duration.Round(time.Second))
+				writeBeeperMediaBackfillSummary(cmd.OutOrStdout(), accountID, sum)
 				if sum.Errors > 0 {
 					_, _ = fmt.Fprintf(cmd.OutOrStdout(), "  %d errors — see sync run items; re-run to retry\n", sum.Errors)
 				}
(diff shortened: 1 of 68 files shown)


wesm:
looking

wesm (approved):
will merge pending CI

PR #621, opened 2026-08-15
feat(vcard): add lossless native resource envelopes

salmonumbrella:
## What changed

- Add a versioned native vCard resource envelope: immutable original bytes, an exact stored body that is the SHA-256 hash and ETag input, and an ordered property-occurrence tree with stable identities (group, PROP-ID, PID, ALTID) so editing one managed property never disturbs unrelated properties, parameters, or vendor extensions.
- Render canonical vCard 4 with import-only 2.1 normalization and a pure vCard 3 compatibility view; properties v3 cannot express stay in the canonical tree and return in later v4 renders.
- Persist envelopes and retired person-UID aliases with revision compare-and-swap and projection-fingerprint conflict detection, keep SQLite and PostgreSQL behavior aligned, and include both in profile subset copies.
- Project typed People data — names, contact points, addresses, dates, media, categories, primary employment, registered relationships, and portable scalar attributes — from one repeatable-read snapshot; everything unmapped remains ordered residue.

## Why

Updating one field msgvault owns must not normalize, reorder, or discard the rest of a contact's card. An exact, independently versioned envelope keeps unknown properties and vendor extensions intact through typed-profile edits, and gives the future CardDAV layer the ETag, revision, and UID-continuity contract it needs.

## Usage

No usage change.

Refs #534


wesm:
looking

wesm:
continuing on this

Issue #612, opened 2026-08-13
Add a cross-channel attachment gallery for each person


salmonumbrella:
## The situation

Msgvault stores attachment occurrences with their owning messages, and the Files workspace can browse archive-wide files. A person page still has no direct way to answer “what files did this contact send me?” across email, chats, and every other linked channel.

Chatwoot provides a useful interaction model: its contact attachment endpoint aggregates message-owned attachments across all conversations, and its contact Media tab separates visual media from other files while retaining a jump to the source message.

## Ask

Add a person-scoped attachment gallery over Msgvault’s durable person and identity-cluster model:

- show every live attachment sent by the person across linked identities, sources, conversations, and channels;
- separate visual media from documents and other files, with bounded pagination and chronological ordering;
- retain exact attachment-occurrence, message, conversation, source, date, role, and direction provenance;
- deep-link every result to the source message instead of copying attachment ownership onto the person;
- default to attachments received from the person, with an explicit option to include files sent to them or exchanged in group conversations;
- reuse the existing Files workspace and person/relationship filters instead of building a second attachment store;
- support filename and metadata filtering without hosted processing;
- when the document or multimodal indexes are enabled, allow semantic search within that same person-scoped result set;
- preserve identity-cluster behavior so linking or splitting identities updates gallery membership without duplicating blobs.

The API and generated clients should expose the same person-scoped, direction-aware contract used by the web person detail. SQLite and PostgreSQL behavior must remain equivalent.

## Delivery boundary

This issue owns person-scoped browsing, direction semantics, and the Media/Files presentation. Document extraction and search remain in #608. Visual attachment embeddings and image queries remain in #609. The attachment occurrence stays message-owned in all three features.

Refs #534, #608, and #609.

Reference implementations:
- https://github.com/chatwoot/chatwoot/pull/14391
- https://github.com/chatwoot/chatwoot/pull/14393

salmonumbrella:
Completed by #622 and #668 (both merged). #622 added the direction-aware person Media/Files gallery with exact message-owned provenance and source deep links. #668 added the shared person-scoped metadata, document, and visual retrieval contract across CLI, MCP, and API. Closing.

Issue #609, opened 2026-08-13
Add opt-in semantic search for visual attachments (images, videos)


salmonumbrella:
## The situation

Msgvault stores original image and video attachments, but their visual content is not searchable. A screenshot, photo, diagram, or short video can contain the thing I am looking for while message text and filename search return nothing.

## What works today, and why it is not enough

Existing text embeddings can search message text and attachment metadata. They do not represent pixels or motion, cannot accept an image query, and cannot safely identify which live attachment occurrence produced a match.

## Ask

Add an independently enabled Voyage multimodal lane for authoritative standalone visual attachments:

- use `voyage-multimodal-3.5` with bounded owning-message text and one original media item per document;
- support text and image queries through attachment search and the Files workspace;
- start with JPEG, PNG, WebP, and direct-input MP4 only after authenticated provider-contract checks pass;
- support animated GIFs only if the provider check proves motion-sensitive behavior;
- classify attachment roles from source evidence and fail closed for inline, thumbnail, sticker, avatar, UI-asset, and unknown roles;
- preserve exact message and attachment-occurrence provenance instead of resolving bare content hashes;
- keep hosted processing disabled by default with separate consent, privacy disclosure, quotas, and status;
- provide matching SQLite/sqlite-vec and PostgreSQL/pgvector lifecycle, filtering, stale-result suppression, and crash-recovery behavior;
- fuse filename/metadata and visual ranks with rank-based fusion without changing ordinary message-search ranking.

PDF/document extraction, OCR, audio, transcription, FFmpeg, keyframes, transcoding, and generated attachment text are out of scope. PDF semantic search is tracked separately in #608.

Implementation depends on the final contextual vector interfaces from #589. Keep this issue open until text/image attachment retrieval, backend parity, privacy gates, and the full verification matrix are complete.

All committed fixtures and public artifacts must be synthetic and free of sensitive real-world data.

salmonumbrella:
Draft implementation is open in #611. It includes the attachment authority/lifecycle foundation, independent consent and reconciliation, the Voyage multimodal adapter and contract harness, and crash-safe two-phase worker publication.

The draft intentionally waits on #589 for concrete SQLite/pgvector backends, runtime/scheduler wiring, search transports, and Files rank fusion.

Issue #608, opened 2026-08-13
Add opt-in semantic search for every authenticated Mistral document format


salmonumbrella:
## Problem

Msgvault stores attachment bytes, but their contents are not searchable. Relevant evidence can be trapped inside PDFs, Office files, spreadsheets, EPUBs, structured text, source code, or email containers while message and filename search return nothing.

## Proposed direction

Add an opt-in document indexing lane with separate consent for each hosted provider:

- classify attachment occurrences with authoritative role and source-part provenance;
- send eligible standalone documents to the stateless Mistral OCR endpoint using private inline data URLs;
- retain normalized source units, bounded chunks, citations, and lexical indexes locally while discarding raw provider JSON and full provider Markdown;
- generate document vectors with the configured Voyage text embedding profile;
- expose dedicated lexical and semantic document search with exact attachment, message, source-unit, extraction, and vector-generation provenance;
- hide stale results immediately on replacement, deletion, role change, or message lifecycle change.

## Format scope

The target is every non-visual attachment format accepted by the pinned stateless Mistral OCR endpoint, not PDF alone. The authenticated capability probe must cover:

- PDF, DOCX/DOC, ODT, RTF;
- PPTX/PPT;
- XLSX/XLS, ODS, CSV, Numbers;
- EPUB;
- TXT, Markdown, RST, LaTeX;
- JSON/JSONL, XML, YAML;
- representative supported source-code formats;
- EML and MSG.

Mistral explicitly documents OCR document inputs including PDF, DOCX, and PPTX, while its broader upload products list additional families. Documentation is discovery evidence, not upload authority: each media type enters Msgvault only after a synthetic authenticated probe proves inline stateless input, stable bounded source units, safe failure behavior, and usable citations. Passing formats ship with family-specific caps and quality fixtures. Rejected formats remain visible as provider-unsupported; Msgvault must not silently create a Mistral File or Library as a fallback.

Standalone images, GIFs, and video stay in the separate Voyage multimodal issue. Audio remains a transcription concern. Archives, executables, encrypted inputs, and unknown containers fail closed.

## Delivery boundary

This lands as a reviewable stack: consume the shared attachment-role foundation, add the common Mistral client/schema and lexical indexing, enable every passing format family, then add required Voyage vectors and semantic/RRF retrieval. Lexical-only operation is a useful fallback but does not complete this issue. Files-workspace integration can follow independently.

[...]

PR #622, opened 2026-08-15
feat(web): add person attachment galleries

salmonumbrella:
## What changed

- Add a person-scoped file search that spans every linked identity, source, conversation, and channel, selectable by direction: files sent by the person, sent to them, or exchanged in a shared group conversation.
- Keep the attachment occurrence message-owned. Each row carries exact provenance — matched cluster members, participant roles, and the owning message, conversation, and source — and deep-links to the source message rather than copying ownership onto the person.
- Expose the contract at `POST /api/v1/people/{id}/files/search` and regenerate the Go and web clients from the OpenAPI spec. SQLite and PostgreSQL behavior stays equivalent, and there is no schema or migration change.
- Add a Media presentation to the person's Files view that separates visual media from other files, throttles thumbnail loading through a shared bounded queue, and keeps the existing archive-wide Files response contract untouched.

## Why

A person page could not answer "what files did this contact send me?" without falling back to the archive-wide Files workspace and filtering by hand, which cannot follow one person across linked identities and channels. Scoping the existing file search to an identity cluster answers it directly, and because the attachment occurrence stays with its message, linking or splitting identities changes gallery membership without moving or duplicating blobs.

## Usage

Open a person in the web UI and use the Files view. Results default to files received from that person; the direction controls add files sent to them or exchanged in group conversations, and the Media/Files toggle switches between the thumbnail gallery and the existing table.

Semantic search inside the person-scoped result set is deliberately out of scope here — it depends on the document and multimodal indexes in #608 and #609.

Refs #612


wesm:
looking

PR #616, opened 2026-08-14
feat(search): index document attachments with Mistral

salmonumbrella:
## What changed

- Add an opt-in Mistral OCR document-indexing lane with authenticated format capabilities, explicit retention/training consent, bounded processing, and private inline uploads.
- Classify attachment occurrences across importers with stable source-part provenance, then reconcile edits, replacements, deletions, retries, and rebuilds for SQLite and PostgreSQL.
- Store normalized source units and chunks locally, expose dedicated lexical document search and status through CLI, HTTP/OpenAPI, and MCP, and return exact attachment, message, and source provenance.
- Keep raw provider JSON and full provider Markdown transient, disclose derived plaintext in full backups, and leave attachment-document embeddings to #617.

## Why

Msgvault stores attachment bytes, but evidence inside documents is not searchable. This adds a consent-gated lexical foundation that can find document content without silently uploading unsupported formats or returning stale results after attachment lifecycle changes.

The authenticated Mistral capability matrix covers all 26 non-visual document formats supported by this lane:

- Documents: PDF, DOCX, DOC, ODT, and RTF
- Presentations: PPTX and PPT
- Spreadsheets: XLSX, XLS, ODS, Apple Numbers, and CSV
- Ebooks: EPUB
- Text and markup: TXT, Markdown, reStructuredText, and LaTeX
- Structured data: JSON, JSONL/NDJSON, XML, and YAML
- Source code: Go, Python, and JavaScript
- Email: EML and Outlook MSG

Each format still has to pass the authenticated private-fixture probe before Msgvault will send that format to Mistral or include it in the recorded consent profile.

## Usage

Set `MISTRAL_API_KEY`, enable `[attachments.documents]`, and choose explicit retention and training postures. Then probe a private synthetic format matrix and record consent for that exact capability manifest:

```sh
msgvault documents probe-mistral --fixtures <fixture-directory> > mistral-capabilities.json
msgvault documents consent-mistral --capabilities mistral-capabilities.json --yes
msgvault documents build --capabilities mistral-capabilities.json --yes
msgvault documents search "shipping damage"
```

Document embeddings remain unavailable in this PR; search is lexical and filename-ranked. The semantic/vector phase is tracked in #617.

Refs #608 and #617

PR #635, opened 2026-08-18
feat(documents): index attachments with shared Docbank processing

wesm:
Msgvault can now extract and search standalone PDF attachments through an explicit, capability-gated Mistral workflow. It imports Docbank as a Go library; it does not start a Docbank daemon, open a Docbank vault, or share application data with a Docbank installation.

Operators create a private synthetic fixture set, probe the configured endpoint and model, review the resulting manifest, and record consent for that exact policy before a build can upload a document. The first capability contract authorizes PDF only. Formats without an enforceable pre-upload provider-unit bound remain blocked.

Docbank owns deterministic normalization, private staging, format detection, capability evidence, and the Mistral transport. Msgvault keeps attachment and message provenance, claims and retries, reconciliation, run budgets, configuration, consent records, storage, backups, local maintenance, and search serving. Provider extraction is manual-only: the daemon reconciles local state and cleans derivatives, but it never starts an upload on its own. When the daemon owns the archive, it handles an explicit build or resume command under its writer lock.

The continuation also makes oversized candidates reach a durable terminal state without starving later work, keeps document-search pagination on one stable ranked candidate set, rejects stale reconciliation updates during concurrent changes, and invalidates pagination when attachment cascades remove searchable occurrences.

The earlier branch described an EU policy but serialized Mistral global endpoint. Because this feature has not shipped, this continuation records the actual EU endpoint instead of preserving consent bytes for the wrong destination.

This continues #616 from commit 73d6c0b33f74c1fd072a7c0258f1cf1e80054698 and preserves the original contributor commit and authorship.


wesm:
Incorrect. Merging

Issue #367, opened 2026-06-03
Idea: a small retrieval-quality eval for msgvault (+ a finding) — worth pursuing?


fmasi:
Hi Wes — thanks for msgvault, I've enjoyed digging through the code.

I've been teaching myself RAG by building a little email-RAG side project (https://github.com/fmasi/mailrag), and while messing with retrieval I started benchmarking against msgvault. A few things turned up that feel more useful to a real project than to my hobby repo, so I figured I'd share them.

I wanted it reproducible by someone other than me, so I ran msgvault's fts/vector/hybrid modes against the public TREC 2010 Legal Track (Enron) qrels — real human relevance judgments, so anyone can re-run it. Two observations, and I might just be holding it wrong:

- There's no way to measure retrieval quality today, which felt like the main gap. I hacked together a tiny `msgvault eval` locally (recall@k, nDCG, MAP over a qrels file) just to see, and it works. Happy to share it.
- A reranking stage helps a lot on the real qrels: P@10 went from about 0.23 to 0.40 on the verbose queries.

(I also hit one clean bug while doing this, filed separately as #366: hybrid mode errors on FTS5 special characters in the query.)

Why I bothered: on my own ~20k-email work mailbox, the same stack (hybrid, reranking, thread expansion, contextual summaries) took coverage@3 from 45% to 84%, and recall@1 from 36% to 70%. Big caveat: those 360 eval queries are LLM-generated and LLM-judged, which is honestly the whole reason I wanted to re-test on a public set with real human judgments. The private numbers are what I live with day to day; the public run is there so you don't have to take my synthetic set on faith. Method's written up in the repo.

Anyway — would an in-tree retrieval eval (and maybe an optional reranker) be useful to you, or is it out of scope? If you're open to it I'm glad to do the work; I'd just want a steer on the shape you'd prefer (an `internal/eval` package plus a dev command, a separate tool, whatever fits). No worries if it's not a fit.

Thanks for reading.

hiSandog:
A small checked-in eval harness would be valuable if it separates corpus preparation from retrieval scoring and records the exact embedding model/index settings in each run. TREC qrels can provide the first reproducible baseline, but a tiny msgvault-shaped fixture with threads, quoted replies, dates, and participants would catch product-specific regressions that legal-document retrieval may miss. Reporting latency and index size beside recall/nDCG would also keep quality improvements from hiding an operational cost.

salmonumbrella:
@fmasi I'd love to see the eval.

[...]

PR #649, opened 2026-08-19
feat: add `msgvault eval` — retrieval-quality evaluation against relevance judgments

fmasi:
Adds `msgvault eval` to measure whether search changes help users find relevant messages. It compares keyword (`fts`), semantic (`vector`), and combined (`hybrid`) search against a list of queries and relevance ratings you provide.

```sh
msgvault eval --qrels qrels.txt --topics topics.tsv --modes fts,vector,hybrid -n 100
```

`topics.tsv` contains query IDs and search text, separated by tabs. `qrels.txt` contains relevance ratings in the format `query_id 0 document_id grade`; grades of 1 or higher mean relevant.

- Reports how many relevant results were found, how highly they ranked, and query timings, as a table or `--json`.
- Records the model, search settings, and archive/index sizes so runs can be compared.
- Supports message or conversation IDs with `--doc-key`; conversation scoring counts each conversation once.
- Reports skipped queries, missing ratings, and incomplete rankings, and rejects conflicting ratings or ambiguous document IDs.

Uses a local SQLite archive. Vector modes need a configured embedding provider and an existing index. Compare runs using the same query file.

Follow-up to #367.


fmasi:
Fixed all three, pushed in d145c6b.

**High (FTS ranking):** the eval was reading `query.Engine.Search`, which sorts by recency, not `store.SearchMessagesQueryContext`, which orders by BM25 and is what production search actually serves. Switched to the BM25 path. Also picked up the deletion-scope and address-filter differences you flagged along the way.

**Medium (PoolSaturated):** confirmed it was being dropped. A saturated per-signal pool now reports as a distinct shortfall reason from real corpus exhaustion, naming `k_per_signal` as the setting to raise. Chose not to auto-raise it — that's a fusion parameter, not a page size, and changing it silently would make the reported numbers not match what's in the provenance block.

**Medium (topic parse errors):** `search.Parse`'s error wasn't checked. Now parsed once per topic, error surfaced, topic skipped with a note instead of silently evaluating a broader query.

New tests for all three, each verified red first (revert the fix, confirm the new test fails). Full suite green.

fmasi:
Confirmed. attachVector always built an OpenAI-compatible client and ignored api_format entirely. Production's newEmbeddingRuntime (serve_vector.go) switches on Embeddings.EffectiveAPIFormat() to pick between the OpenAI-compatible client and NewVoyageClient for voyage-contextual. Pulled that selection logic into a shared constructor both index-side and eval now call, so they can't drift apart.

The bigger issue was the query-vs-document role. Voyage's contextual API tags each embed call input_type: document or input_type: query. The old code sent neither, since it never reached the Voyage client at all. Both client types already implement the same EmbedQuery interface, so once the right client gets constructed, the query role comes along for free.

Also validates the resolved config now and fails clearly on an unsupported api_format instead of silently falling back. New tests assert the actual wire protocol for both formats: path, input_type, nested vs flat body.

fmasi:
Fixed both, pushed in b380968.

All-zero qrels: confirmed — a topic id present in the qrels file with every judgment at rel=0 produced the same empty relevant set as a topic id that never appears at all, so both got skipped. Added Qrels.HasJudgments to tell the two apart; a topic judged entirely non-relevant now scores normally (contributing real zeros to the macro average) instead of being silently dropped.

Provenance: confirmed too, both halves. Corpus counts now go through the same live-messages predicate search actually uses, so dedup-hidden and source-deleted rows drop out. IndexedVectors is now scoped to the generation_id the run resolved as active, not a whole-table count across every retired generation vectors.db still holds.

While fixing that second one we found one more gap in the same vein and decided to document rather than chase: attachVector resolves the active generation once at setup, but the search engine re-resolves it on every query. If someone reactivates embeddings mid-run against the same archive, the reported vector count can go stale for topics scored after the swap. Narrow enough (you'd have to be rebuilding against the same archive you're benchmarking, at the same time) that a lock or per-query check felt like more machinery than the risk warrants. Noted directly in the command help and the field's doc comment instead.

fmasi:
Fixed all four, pushed through 97e83a5.

Empty parsed queries: confirmed. subject:"" parses without error but to no search criteria, and fts answers that by listing the whole live corpus newest-first. Extended the existing per-topic parse check to also catch Query.IsEmpty(), so it skips and reports instead of silently scoring a date sort.

Unjudged topics: added a count and the topic IDs to diagnostics, both table and JSON. Left out the coverage flag you suggested. Partial judgment coverage is normal while a qrels file is still being built, and every other non-fatal condition on this command goes to diagnostics rather than failing the run. A flag would have been the one exception to that.

Duplicate topic IDs: rejected outright rather than deduped. A repeated qid scores the same judgments twice, so it double-weights that query in the macro average, and if the two occurrences carry different query text there's no non-arbitrary way to pick a winner. LoadTopics now names every offending id and refuses to load.

MAP/MRR labeling: they're now MAP@n/MRR@n, matching what the P@k/R@k fix already did two rounds back. Rankings get truncated to -n before AP and MRR are computed, so a relevant document past that rank is invisible to them the same way it is to recall — the unqualified label was overclaiming what the number covers.

One more we caught ourselves reviewing this batch: the new unjudged-topics note computed coverage as Parsed minus unjudged, which only equals what actually got scored when every judged topic also scores. A topic judged but empty-query, or skipped by every mode, is neither unjudged nor scored, so that arithmetic could print a coverage number that contradicts topics_evaluated in the same report. Fixed to read the same counter topics_evaluated is built from.

fmasi:
Fixed both, pushed through bcaaa90.

Cross-source doc-key collisions: confirmed, and this one's real for the common case, not an edge case. source_message_id and source_conversation_id are assigned by the connected account, not globally, so two accounts in one archive can issue the same id for unrelated documents. I didn't source-qualify the key shape, though — every qrels file this branch tests against, including the TREC set, carries bare ids with no source, so composing the key would break every existing qrels file by scoring a flat zero instead of erroring, which is the same silent corruption one level up. Instead the run now checks up front whether the archive's chosen doc-key actually has a single id space (COUNT(DISTINCT source_id) per id, GROUP BY/HAVING), and refuses to score with a clear error naming the colliding ids if not. An archive with disjoint sources (a mailbox plus a chat account, say) scores normally, since there's nothing to collide.

Duplicate --modes: deduped, not rejected, and deliberately different from how we handled duplicate topic ids last round. Two topic rows sharing a qid can carry different query text, so there's real ambiguity to reject over. fts,fts has exactly one reading, so rejecting it would buy no safety, just friction.

Our own review of this batch caught two more, both minor: the collision check ran its full scan even on a single-source archive, which is the common case and provably can't collide, so it now short-circuits on a cheap distinct-source count first. And the error message's id list was being pre-truncated by hand before being passed to the shared formatter, so the formatter's own truncation logic never actually ran there — simplified to pass the untruncated list through.

fmasi:
Status: every finding through round 5 is fixed and pushed (see the replies above). CI's green apart from test-postgres and test-windows-slow-packages, still running.

One more just landed: per-mode corpus provenance under an account-scoped vector generation. It's real, but restructuring the report to carry per-mode corpus counts is a bigger change than anything so far, and it's the sixth round of findings on this PR.

Before going further I'd like a steer: keep chasing rounds to zero, or is this a good state to pause and look at as-is? Not trying to shortcut anything real, just aware the CI here runs on your account, not mine, and didn't want to keep spending it without checking that's wanted.

fmasi:
Curious about the round count here, so I looked at how this repo's own merges have handled roborev findings: 45 of the last 48 merged clean, and the 3 that didn't either had a documented reason or, in one case, none — which then broke main. That's enough to keep working through rounds properly rather than stopping. Will keep responding to what comes next.

fmasi:
Fixed, pushed through a10867c.

Added VectorMessages and VectorConversations to the provenance block. For vector mode they report the population an account-scoped [vector.embed.scope] generation actually searches. The number comes from intersecting the generation's embedded message ids in vectors.db against main.db's live, scoped, stamped rows. The table only prints the extra line when it differs from the archive-wide corpus, so an unscoped generation, the common case, doesn't get a second line saying the same thing twice.

First pass computed messages through the existing Backend.EmbeddedMessageCount, then conversations through a second, separately-erroring query. Reviewing this batch caught the problem. The two calls can desync on a transient failure in just one of them, printing a self-contradictory line ("0 live messages, N conversations") with nothing in the output explaining why. Both counts now come from one query. It reads COUNT(DISTINCT id) and COUNT(DISTINCT conversation_id) off the same row, so they can only ever go stale together.

Added a test that deliberately stamps an out-of-scope message into the same generation, to pin down the source_id filter itself rather than let embed_gen alone carry the test for the wrong reason.

fmasi:
Fixed both, pushed through c381a40.

Corpus provenance: confirmed. `collectVectorCorpusStats` required `embed_gen = gen`, but `Backend.Search` never checks that column at all. It reads vectors.db purely by `generation_id`. A content change resets `embed_gen` to flag a message for re-embedding, and the stale vector stays searchable right up until the re-embed actually runs. Requiring the stamp undercounted relative to what a run can retrieve. Same mismatch, wrong direction. Dropped the predicate. Membership is now just being in vectors.db for this generation, live, in scope, matching Search's own contract exactly.

SQLite parameter limit: real. Not new to this round, either. `GetMessageSummariesByIDs` binds one parameter per id in the `IN (...)` list, and it's shared with production code (`internal/mcp/handlers.go`), not eval-only. Chunked it into batches of 500 rather than rewriting the query around SQLite's json1 extension. This engine also serves PostgreSQL, and `json_each` means something different there (object keys, not array elements), so a dialect-agnostic fix beat a SQLite-specific one. The existing "reassemble in caller order" contract stays intact across chunks. Other callers depend on it for search rank.

New tests for both. One resets an embedded message's `embed_gen` to NULL after seeding and checks it still counts. The other requests 503 ids in reversed order across the chunk boundary and checks full recall plus exact order.

fmasi:
Fixed, pushed through 2026b2c.

Confirmed. Chunking the base summary query wasn't the whole fix. The label hydration right after it still passed every result into one IN list. Same overflow, one call later. Wrapped that call in the same 500-id loop, so it now hydrates labels chunk by chunk instead of all at once.

Extended the existing chunk test rather than writing a new one. Gave the first message (chunk one) and the last message (chunk two, since the fixture sits at messageSummaryIDChunk + 3) each a label, then checked both come back attached. It's worth being honest about what that proves. 503 ids is nowhere near SQLite's actual 32766-parameter ceiling, so it can't demonstrate the overflow itself. What it does catch is a bug that silently skipped the second label-fetch call, or mis-sliced the chunk boundaries, which is the failure mode this fix could plausibly introduce, even if it isn't the one roborev originally flagged.

fmasi:
Fixed, pushed through 21d4d05.

Confirmed, and it's the same bug in a different package this time. Rounds 7 and 8 chunked the overflow in `internal/query/sqlite.go`, which serves eval's vector and hybrid modes. FTS mode never goes through that code. It calls `Store.SearchMessagesQueryContext` directly, which hydrates recipients and labels through its own separate functions in `internal/store/api.go`, none of which were touched by the earlier rounds. Same unchunked `IN (...)` list, same fix: `batchGetRecipients` and `batchGetLabels` now loop over the ids in batches of 500, merging into a shared map per chunk.

Checked whether anything else in that file has the same shape reachable from eval. `GetMessagesSummariesByIDsContext` binds an unchunked list too, but it's only called from the daemon's API server, not from anything eval touches, so it's staying as is for now.

New test creates enough messages to cross the chunk boundary, puts a recipient and a label on the first message and on the last, and checks both come back hydrated correctly.

fmasi:
Fixed, pushed through 66d5db8.

Confirmed. The stub only kicks in when sqlite_vec is missing, so a binary built with sqlite_vec but not fts5 gets the real eval command anyway. `--modes fts` then falls back to `Store.SearchMessagesQueryContext`'s LIKE branch without ever telling you, and the report keeps calling that mode "fts" as if it scored BM25. `requireFTS5ForModes` checks FTS5 availability before the first topic runs. If it's missing, the whole run stops with a clear error. Extracted it as a plain function instead of wiring the check straight into the command, mainly so it's testable without a real store. There's no way to force FTS5 unavailable across a package boundary here, only from inside the store package's own tests.

Checked whether hybrid needs the same guard. It doesn't. Its BM25 leg runs `messages_fts MATCH` directly inside the vector backend's fused query, and that query references the table unconditionally, so a build without FTS5 fails outright there rather than quietly downgrading.

One thing I looked at and didn't fix: `Store` has a second, narrower fallback that forces the same LIKE branch mid-run if a query that started fine ever errors later, corruption or an extension fault, even on a build where FTS5 was genuinely available at startup. My check can't see that coming, and closing it properly would mean changing `SearchMessagesQueryContext` itself, which several daemon handlers depend on well outside anything eval touches. That felt like real machinery for an edge case rare enough to not be worth it here, so I documented it on the function instead of chasing it.

fmasi:
Fixed, pushed through c86b21f.

Confirmed, and it's exactly the mismatch you'd expect. The corpus query filtered on the raw config fields, `vecCfg.Embed.Scope.MessageTypes`/`SourceIDs`, while both the embed path and eval's own vector search go through `BuildScope()` first. That call lowercases and trims message types and drops non-positive source ids. `message_type` has no case-insensitive collation, so a config value like `"EMAIL"` bound literally into the `IN` clause matched nothing, even on an archive where every row is `"email"` and search would have returned all of them. Same fix as everywhere else this has come up in this PR: build the normalized scope once, then use its fields instead of the config's.

New test sets `message_types = ["EMAIL"]` against an archive seeded with the usual lowercase rows and checks both counts still come back as 2, not 0.

fmasi:
Pushed 665d7f2, merging main back in. Not a review finding, just catching up. Main had drifted about 30 commits since this branch forked, and one of them, the semantic person search feature, touched the same function this PR's own client-sharing fix lives in.

Only one file conflicted, `serve_vector.go`. Both sides had changed the OpenAI and Voyage client construction in `newEmbeddingRuntime`: mine to route the message client through the shared constructors eval and search also use, main's to add a second, person-scoped client alongside it. Kept both. The message client still comes from the shared constructor, the person client is built the way main already had it, just with `BeforeRequest` wired to the person gate. `newConvergenceChecker` took main's newer four-argument signature, since that's what the person feature needs and every call site already expected it.

Full suite, `go vet`, `golangci-lint`, and `testify-helper-check` all pass after the merge. I also had a second pass specifically verify the hand-resolved logic rather than trust "it compiles" alone, since nothing in the existing tests exercises this exact combination.

fmasi:
Fixed both, pushed through 998af0b.

Rank inflation: confirmed, and it's a real one. `DedupeKeys` dropped a hit outright when its doc-key came back empty, which happens whenever a message has no `source_message_id` (or `source_conversation_id`, depending on `--doc-key`). MRR, AP, and nDCG all read off rank position. Removing a slot from the middle of a ranking isn't neutral. Every relevant document below it shifts up one, and the run gets credit for a rank it didn't earn. Gave the hit a key instead, unique to its own position in the list. It still occupies the slot, so nothing shifts, and it can never resolve to relevant since no qrels file names an empty or null-byte-prefixed docid. Two hits missing an id don't collapse into each other either, since each gets its own position-based key. New test checks the actual number. A relevant doc sitting second in the ranking now scores MRR 0.5, not the 1.0 the old drop-then-shift behavior would have handed it.

Context propagation: also confirmed. `InitSchema` and the startup migration both ran on a background context, so Ctrl-C during a long one on a large archive did nothing. Both already had context-aware versions elsewhere in the codebase, so eval just needed to call them. Added `runStartupMigrationsContext` as a sibling to the existing helper rather than changing its signature, since three other commands call it and none of them needed to change.

fmasi:
Fixed, pushed through e71ea2a.

Confirmed, and it's the same shape as a bug this file already guards against on the topics side. `LoadQrels` did a plain map assignment per line, `q[qid][docid] = rel`, so a second line grading the same pair differently just overwrote the first with nothing to show for it. Whichever line happened to load last decided the grade. `LoadTopics`, right below it in the same file, already rejects a repeated query id for exactly this reason: a caller can't repair that kind of ambiguity, only the file can. Gave qrels the same treatment. Read the whole file, then reject it in one message naming every conflicting pair, not just the first. An identical repeat, the same grade twice, still loads fine, since merged or re-exported judgment files do that constantly and there's nothing wrong with it.

New tests cover both sides: two lines grading "301 docA" as 1 and 0 now fail with both values named in the error, and two identical "301 docA 1" lines still load clean.

rodboev:
@fmasi could you rebase this onto main? It's conflicted, and the reranking work in #877 needs the eval to land first. If you're short on time I can open a follow-up PR with your commits rebased.

wesm:
i'm rebasing this now

wesm:
Rewriting the PR description to not be a wall of text.

wesm:
I can't push to this fork, so I'm going to open a PR to supersede.

PR #611, opened 2026-08-13
feat(vector): add opt-in multimodal attachment search

salmonumbrella:
## What changed

- Add an independently enabled visual attachment lane for standalone JPEG, PNG, WebP, and direct-input MP4 attachments, using bounded owning-message context and exact live attachment-occurrence provenance.
- Store generation-scoped visual vectors in SQLite/sqlite-vec and PostgreSQL/pgvector with explicit hosted-processing consent, fenced claims, atomic publication, stale-result suppression, retry, retirement, and crash recovery.
- Add text and image similarity search through the CLI, HTTP API, MCP, and Files workspace, with sender, source, message, date, filename, and MIME filters plus stable pagination.
- Fuse filename and visual results in Files with deterministic reciprocal-rank fusion while leaving ordinary message-search ranking unchanged.
- Keep animated GIF indexing disabled until provider capability is confirmed. PDF extraction, OCR, audio, transcription, transcoding, and generated attachment text remain out of scope.

## Why

Message text and attachment metadata cannot find something that is only visible in a screenshot, photo, diagram, or short video. This adds a separate, consent-gated retrieval lane without mixing visual vectors into message embeddings or losing the source message and attachment occurrence behind each result.

The formats supported by this PR are:

- Images: JPEG/JPG, PNG, and WebP
- Video: direct-input MP4

GIF indexing is not enabled in this PR. Animated GIFs remain blocked until the provider capability is authenticated; PDF, OCR, audio, transcription, and transcoding are also out of scope.

## Usage

Set `VOYAGE_API_KEY`, then enable the lane in `~/.msgvault/config.toml`:

```toml
[vector.multimodal]
enabled = true
```

Restart the daemon, explicitly consent to hosted attachment processing, then search by text or image:

```sh
msgvault multimodal build --yes
msgvault multimodal status
msgvault multimodal search "diagram description"
msgvault multimodal search --image query.png
```

Closes #609. Refs #612. Built on #589.

Draft boundary: the authenticated Voyage contract and representative-corpus evaluation still need to pass before this is ready to merge.

Issue #623, opened 2026-08-15
Expose person-scoped attachment retrieval through MCP and the CLI


salmonumbrella:
## The situation

"What files did this contact send me last week?" is one of the most useful questions an archive can answer, and it is about to be answerable three different ways, none of which an agent or a terminal can reach through one person.

#612 (PR #622) adds `POST /api/v1/people/{id}/files/search`: a person-scoped file search with an explicit `from_person` / `to_person` / `group` direction vocabulary, date bounds through the shared explore predicate, filename and MIME-family filters, and exact attachment-occurrence, message, conversation, source, role, and direction provenance. It widens the scope to the person's whole identity cluster so files owned only by a linked alias are still found. It ships with no MCP tool and no CLI command — it exists to back the web person detail.

#616 adds the `search_documents` MCP tool and a `msgvault documents` command group for extracted document content. #611 adds the `search_visual_attachments` MCP tool and a `msgvault multimodal search` command for visual attachment embeddings. Both are lane-scoped: they search their own index, with their own filters, under their own consent gate.

So once all three land, the archive can answer the question over HTTP, and can answer parts of it semantically, but an agent still cannot ask it once and get everything a person sent.

## What works today, and why it is not enough

Each of the three surfaces is missing a different piece.

`search_documents` accepts `query`, `source_ids`, `message_types`, `attachment_id`, `message_id`, `limit`, and `cursor`. There is no person filter and no date range, so document content cannot be scoped to a contact at all. The only path is to already know the message or attachment ID, which is the answer, not the query.

`search_visual_attachments` is closer: it accepts `sender_person_id` alongside `after` and `before`. But it is sender-only — there is no way to ask for visual attachments sent *to* someone or exchanged in a shared group conversation — and it resolves the person by joining `person_participants`, the durable person root. #612 resolves `/people/{id}` as a participant ID widened through `ClusterMembers`. Those are two different identifier namespaces with two different link semantics, and the schema comment on `person_participants` already notes that unlinking can leave one person spanning split clusters. Handing the same integer to both tools will silently return different populations.

[...]

PR #650, opened 2026-08-20
feat(vector): add capability-gated visual attachment search

### LINKED ISSUE #609: Add opt-in semantic search for visual attachments (images, videos)

wesm:
## What changed

- Add an independently enabled visual attachment lane for standalone JPEG, PNG, WebP, GIF, and direct-input MP4 attachments, using bounded owning-message context and exact live attachment-occurrence provenance.
- Delegate media detection, eligibility bounds, and the Voyage transport to [Docbank](https://github.com/kenn-io/docbank)'s `document/media` and `document/voyage` packages — the same audited provider contract the document lane uses for Mistral. Msgvault keeps provenance, claims, reconciliation, vector storage, consent records, schedules, and search serving.
- Gate every upload on an operator-probed capability manifest: each format, animated GIFs, video, image queries, text-plus-media documents, and multi-item batches all require their own recorded probe evidence. The probe uploads only deterministic synthetic fixtures; no archive content leaves the machine.
- Bind hosted-processing consent to the manifest's policy fingerprint: build records it, and resume, retry, and scheduled passes refuse to run when a re-probed manifest changes it.
- Ride the attachment change journal exactly as #635 shipped it: message subject and body edits are handled by in-place invalidation triggers, a stale-publication sweep, and a commit-time CAS on the owning message's content stamp, so a context edit during a provider request can never publish a vector of the old context.
- Store generation-scoped visual vectors in SQLite/sqlite-vec and PostgreSQL/pgvector with fenced claims, atomic publication, retry, retirement, and crash recovery; search by text or image through the CLI, HTTP API, MCP, and Files workspace with deterministic reciprocal-rank fusion.

## Why

Message text and attachment metadata cannot find something that is only visible in a screenshot, photo, diagram, or short video. This adds a separate, consent-gated retrieval lane without mixing visual vectors into message embeddings, and shares the provider-facing machinery with the document lane so format detection, request shapes, limits, and failure classification have one owner.

## Usage

Set `VOYAGE_API_KEY`, then probe the provider with private synthetic fixtures and review the manifest:

```sh
msgvault multimodal probe --seeds <seed-directory> --out voyage-capabilities.json --yes
```

Enable the lane in `~/.msgvault/config.toml`:

```toml
[vector.multimodal]
enabled = true
capabilities_file = "/path/to/voyage-capabilities.json"
```

Restart the daemon, consent to exactly that capability profile, then search:

```sh
msgvault multimodal build --yes
msgvault multimodal status
msgvault multimodal search "diagram description"
msgvault multimodal search --image query.png
```

Continues #611 from commit d6e8450bf49375878f3ef16344544b4968bfc72d and preserves the original contributor commit and authorship. Closes #609. Refs #612.

Draft boundary: go.mod pins Docbank to the kenn-io/docbank#174 branch (per-format interleaved and mixed-batch probes) and must be re-pinned to main after that PR merges; the Nix `vendorHash` needs a refresh for the dependency bump; and the live Voyage probe against the real endpoint has not yet been run.


salmonumbrella:
i’m building #623 on top of this branch so person-scoped metadata, document, and visual retrieval can land in one sweep. i’m treating durable person IDs as canonical, translating participant IDs through the same resolver, and extending visual search at its public request boundary. i’ll keep rebasing onto #650 and avoid touching reconciliation/provider internals. lmk if you expect the request or result contract to move again.

wesm:
Great, hopefully this will be able to land without too many more cycles

wesm:
Declining both findings on `66ef95d` (recorded previously on f245e04, 6fb8ea1, 52440cd, 66ef95d):

Both reduce to the same residual window — a mutation of the searchable publication set racing an in-flight pagination, bounded to one transient skipped or duplicated row in a listing, self-corrected on refresh. The direct-search cursor additionally narrows this to publications replaced mid-pagination *within* one generation, which only occurs while a build pass is actively running. The proposed fix (a transactionally maintained per-generation revision counter) would need to be bumped in every publication-mutating store method and all invalidation-trigger statements on both backends — disproportionate hardening for an offline archive UI. The existing guards (generation pin, cache revision, pre-ranking publication marker, canonical query vectors, vectorless-cursor rejection) cover every non-sub-second case.

wesm:
Declining the finding on `d0c590e` and merging:

The window is local-only and transitional: after a manifest change reopens reconciliation, search serves already-embedded vectors under the previous eligibility policy until consent is renewed and the re-scan converges. No uploads occur during it (builds are consent-gated), nothing leaves the machine, and no billing happens. The proposed fix — refusing search until consent renewal plus full re-reconciliation — would black out search for the whole re-scan after every manifest change, even though most publications are unaffected, and conflates consent (which governs hosted uploads) with serving local queries. Eligibility convergence during policy transitions is by design incremental.

PR #652, opened 2026-08-20
feat: add exact source selection

### LINKED ISSUE #499: Introduce a first-class exact-source selector for account-scoped operations
## What I was doing

While investigating #498, I traced how account selection works across identity management, scoped queries, collection membership, sync, account mutation, and staged deletion.

The identity failure is one instance of a broader modeling gap: msgvault has source IDs internally, but most user-facing account operations can only submit an identifier or display name.

## What happened

The word account currently covers three different selection contracts:

1. **Exact source** — mutate or inspect one `sources` row.
2. **Account family** — select a primary source plus related sources such as Google Calendar.
3. **All matches** — run every syncable source matching an identifier or display name.

Those contracts are implemented independently and have no shared exact-source selector:

- `identity show/add/remove` and `update-account` reject ambiguous names safely, but provide no way to select one of the source IDs shown in the error.
- `search`, `stats`, `identity list`, and scoped dedup use account resolution that may expand into related Calendar sources.
- Collection membership already accepts numeric source IDs.
- `remove-account --type` handles the common cross-type case, but type plus a duplicated display name can still select the first match.
- Staged deletion may begin with an exact `source_id`, but the manifest retains only an account string and resolves it again at execution time.
- `sync`, `sync-full`, and scheduled sync deliberately fan out across matching syncable source types. Their behavior should not be replaced with exact-one resolution.
- TUI filtering and lower-level API filters already operate on source IDs.

The database already supplies two useful identities:

- `sources.id` is exact within an archive and suitable for interactive CLI/API selection.
- `(source_type, identifier)` is unique and is the more stable form for durable configuration or persisted provenance.

Bare numeric account arguments should not become source IDs globally because valid account identifiers, including phone numbers, can also be numeric.

## Why it matters

Ambiguity rejection is the correct default, but an error without an escape hatch leaves valid account layouts unusable. The same identifier can legitimately represent Gmail, IMAP, mbox, meeting, or chat sources, and display names are intentionally human-friendly rather than unique.

This also affects safety. Exact-source mutations should never depend on first-match ordering, and a destructive workflow should not lose exact source provenance between planning and execution.

A first-class selector would give CLI and daemon requests one consistent way to say exactly this source without changing the existing human-friendly or multi-source behavior where that behavior is intentional.

## Proposed direction

Introduce a shared source-selector contract and make each caller declare its cardinality:

- **ExactOne** for identity mutations, account updates/removal, deletion execution, and explicit exact-source scoping.
- **AccountFamily** for account scopes that intentionally include related sources.
- **AllMatches** for sync commands that intentionally fan out.

The selector should support an explicit source ID for interactive CLI/API use. Durable configuration and persisted artifacts should retain the stable source type plus identifier, optionally alongside the source ID and a display snapshot.

Existing account tokens should remain backward-compatible. Ambiguous tokens without an exact selector should continue to fail rather than choosing a row.

## Acceptance criteria

- [ ] A shared selector/request type can carry an explicit source ID without overloading bare numeric account tokens.
- [ ] Exact selection is available through both CLI and daemon API transports.
- [ ] Callers explicitly choose exact-one, account-family, or all-matches resolution semantics.
- [ ] #498 can use the exact-one path for `identity show`, `add`, `remove`, and scoped `list`.
- [ ] `update-account` and `remove-account` can target an exact source; same-type display-name collisions never select the first row silently.
- [ ] Staged deletion preserves exact source provenance from staging through execution.
- [ ] Existing sync fan-out behavior remains unchanged unless an explicit exact selector is supplied.
- [ ] Existing unambiguous account arguments remain backward-compatible.
- [ ] Documentation distinguishes account tokens, source IDs, and the three cardinality modes.
- [ ] Tests cover cross-type duplicate identifiers, same-type duplicate display names, numeric identifiers, Calendar expansion, and intentional sync fan-out.

## Related issues

- #498 — motivating identity-command bug and likely first adopter
- #278 — original accounts, identities, collections, and deduplication model
- #12 — established `remove-account --type` as a disambiguation precedent
- #164 — introduced account-scoped search
- #200 — source-type visibility in account listings

salmonumbrella:
## What changed

- Add one shared source-selector contract with explicit exact-one, account-family, and all-matches semantics. Bare numeric account tokens stay account tokens; only `--source-id` selects by archive-local ID.
- Expose exact source selection through CLI and daemon transports for sync, full sync, account updates/removal, and staged deletion, and move identity operations onto the same shared selector. Existing unambiguous account arguments and intentional sync fan-out stay compatible.
- Persist staged deletion provenance as a stable source type plus identifier, with the source ID retained as a snapshot. Execution resolves the current source from that stable tuple, scopes local archive updates to it, and rejects ambiguous, cross-source, or unsupported targets before claiming a batch.
- Keep SQLite, PostgreSQL, DuckDB, API, TUI, and MCP behavior aligned. Existing version-1 deletion manifests remain readable and can be bound explicitly at execution; there is no database schema or migration change.

## Why

Msgvault already had exact source IDs internally, but most account-scoped operations could only submit an identifier or display name. Ambiguous layouts therefore either had no safe escape hatch or risked losing exact source provenance between planning and execution. This makes cardinality explicit at each call site while preserving the account-family and fan-out behavior that is intentional.

## Usage

List sources to find the exact ID, then pass it explicitly:

```console
msgvault list-accounts --json
msgvault sync --source-id 42
msgvault sync-full --source-id 42
msgvault update-account --source-id 42 --display-name "Work"
msgvault remove-account --source-id 42 --yes
msgvault delete-staged <batch-id> --source-id 42 --yes
```

`--source-id` is mutually exclusive with the existing account token. New staged-deletion manifests retain their source automatically; the explicit deletion flag is mainly the safe binding path for legacy manifests.

Closes #499

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -584,6 +584,9 @@ components:
           type: boolean
         remote_delete_enabled:
           type: boolean
+        source_id:
+          format: int64
+          type: integer
         "yes":
           type: boolean
       type: object
@@ -610,6 +613,9 @@ components:
             - "null"
         remote_delete_env_var:
           type: string
+        resolved_source_id:
+          format: int64
+          type: integer
         scope_escalation_account:
           type: string
         scope_escalation_body_lines:
@@ -1984,6 +1990,8 @@ components:
         message_count:
           format: int64
           type: integer
+        source:
+          $ref: "#/components/schemas/SourceReference"
         status:
           type: string
         summary:
@@ -2020,6 +2028,28 @@ components:
         - created_by
         - description
         - message_count
+      type: object
+    DeletionTarget:
+      additionalProperties: true
+      properties:
+        message_id:
+          format: int64
+          type: integer
+        source_id:
+          format: int64
+          type: integer
+        source_identifier:
+          type: string
+        source_message_id:
+          type: string
+        source_type:
+          type: string
+      required:
+        - message_id
+        - source_id
+        - source_type
+        - source_identifier
+        - source_message_id
       type: object
     DiscoverError:
       additionalProperties: true
@@ -3620,6 +3650,12 @@ components:
           type:
             - array
             - "null"
+        targets:
+          items:
+            $ref: "#/components/schemas/DeletionTarget"
+          type:
+            - array
+            - "null"
       required:
         - gmail_ids
       type: object
@@ -4124,6 +4160,8 @@ components:
         id:
           type: string
         raw_filter: {}
+        source:
+          $ref: "#/components/schemas/SourceReference"
         status:
           type: string
         summary:
@@ -7919,6 +7957,21 @@ components:
         - signals
         - confirmed_at
       type: object
+    SourceReference:
+      additionalProperties: false
+      properties:
+        id:
+          format: int64
+          type: integer
+        identifier:
+          type: string
+        type:
+          type: string
+      required:
+        - id
+        - type
+        - identifier
+      type: object
     SourceStatus:
       additionalProperties: true
       properties:
@@ -8065,6 +8118,8 @@ components:
           type:
             - array
             - "null"
+        source:
+          $ref: "#/components/schemas/SourceReference"
         status:
           type: string
       required:
@@ -8683,12 +8738,16 @@ components:
     UpdateRequest:
       additionalProperties: false
       properties:
+        account:
+          type: string
         display_name:
           type: string
         email:
           type: string
+        source_id:
+          format: int64
+          type: integer
       required:
-        - email
         - display_name
       type: object
     UpdateResult:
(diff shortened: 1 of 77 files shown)


wesm:
looks useful, reviewing this

wesm:
Updated this PR with the review fixes:

- `delete-staged` now carries the exact source ID through planning and execution, including when Gmail and IMAP share an address.
- TUI and MCP staging reject selections that span sources. The TUI now explains how to filter by account.
- CLI, API, and changelog documentation now covers the new selectors and source details.
- Unused Gmail-ID lookup code was removed, and shared deletion checks were consolidated.
- The pre-sync old-daemon safety check remains. The after-the-fact source acknowledgment was removed.
- MCP and dedup now stop before making changes when source details are missing instead of guessing them.
- The API schema is now 2.4.0, so a 2.3 daemon cannot be mistaken for one that supports exact source selection.

The branch is squashed to one commit with `salmonumbrella` retained as the author. Full tests, lint, generated API checks, web checks, docs checks, and the build pass.

Issue #628, opened 2026-08-16
Add a scheduled people sweep that keeps curated attributes current from the archive


salmonumbrella:
## The situation

A person in msgvault is pegged to a stream of conversations that never stops: email, chats, meetings, and now the attachments they send. The profile, by contrast, only changes when someone edits it. So the archive knows that a contact changed jobs, moved cities, got a dog, or stopped mentioning a partner months before the profile does, and "last time we talked" is whatever the user remembers.

The typed attribute layer (#535) and the structured primitives (#565) were built so that facts about a person are typed, historized, and provenance-aware. What is missing is the thing that keeps them current: a scheduled sweep that looks at what happened with a person since the last pass, updates the fields it can update safely, proposes the rest, and keeps the old values as history. Nobody breaks up with a past partner in a database; the name should still be there, closed with a date.

This issue proposes that sweep, plus the attribute catalog it maintains, so a personal-CRM profile like the one below stays populated without anyone typing:

- Contact: email, phone, social profiles, chat handles
- Profile: role, company, location, born in, birthday, address, membership, religion, politics, personality
- Family: partner, pets, kids
- Fun now, fun growing up, favorite food, favorite place, how we met, ask me about
- Last contacted, contact frequency, primary channel
- Last time we talked

## What works today, and why it is not enough

Three pieces are in place, and one of them already does part of this without any model:

- #599 computes `person_contact_state` from the dated activity spine: first and last contact, last inbound and outbound, interaction count. That is "last contacted" and the basis for "primary channel" and "contact frequency", deterministically, for every person.
- #535 seeds `primary_channel`, `contact_frequency`, `ask_me_about`, and `last_contacted` as attribute definitions with bitemporal values (`active_from`/`active_until` plus `created_at`/`superseded_at`) and a `source`/`confidence`/`actor` envelope. The provenance vocabulary already distinguishes declared sources (`user`, `carddav_import`, `vcard_import`) from `extraction` and `enrichment`, and only the latter may carry a confidence.
- #589 embeds chats as conversation windows and meetings as turn-aware chunks; #611 embeds visual attachments; #616 extracts document text. The corpus a sweep would read is retrievable, per person, by meaning.

[...]

PR #653, opened 2026-08-20
feat(people): add consent-gated inference providers

salmonumbrella:
## What changed

- Add an exact, fingerprinted provider policy and SQLite/PostgreSQL consent history for people-sweep inference.
- Add a bounded OpenAI-compatible structured-output transport behind a fail-closed runner that checks request scope, consent, credentials, and output schema before derived data can be accepted.
- Add `msgvault person provider status|consent|revoke|check` through the daemon CLI path. Only the fixed synthetic `check` forwards the configured provider credential; status, consent, and revoke forward none.
- Keep archive content out of the operator check and keep revoked or disabled policies from reaching the provider.

## Why

The people sweep cannot safely send message or meeting text to a hosted model until the operator can see and consent to the exact destination, model, retention/training assertions, allowed sources, date range, and sensitive-content policy. This makes that boundary durable and auditable before the later extraction pipeline exists.

## Usage

Configure `[people.sweep]` and `[people.sweep.provider]`, then inspect and consent to the exact policy:

```sh
msgvault person provider status
msgvault person provider status --all
msgvault person provider consent --yes
msgvault person provider check
msgvault person provider revoke
msgvault person provider revoke --all
```

Changing a fingerprinted policy requires new consent. Stored grants remain visible and revocable with `--all` even when the provider is disabled or its configuration changes. The feature remains disabled by default.

Refs #628

### DIFF
--- a/cmd/msgvault/cmd/attribute_definition.go
+++ b/cmd/msgvault/cmd/attribute_definition.go
@@ -269,7 +269,7 @@ var attributeDefinitionDeleteCmd = &cobra.Command{
 
 func attributeObjectTypeArg(cmd *cobra.Command) (string, error) {
 	value := strings.TrimSpace(attributeDefinitionObjectType)
-	if value == "" || value == "person" || value == "organization" {
+	if value == "" || value == personValue || value == "organization" {
 		return value, nil
 	}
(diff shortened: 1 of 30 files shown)


wesm:
Merging, we can defer some of the documentation until this whole subsystem settles

PR #656, opened 2026-08-20
feat(people): add semantic person search

salmonumbrella:
## What changed

- Build one deterministic, privacy-bounded semantic document for each curated person and maintain it in a person-owned vector corpus, separate from message embeddings.
- Keep curated-person embedding off by default and require a separate exact consent bound to the provider destination, API format, model, credential variable, retention/training posture, renderer, disclosed profile fields, caller-supplied search-query text, and global durable-person scope.
- Keep SQLite and PostgreSQL behavior aligned for publication, reconciliation, cosine ranking, generation lifecycle, and read-time revision fencing.
- Add authenticated `POST /api/v1/people/search` and `msgvault person search`, returning current durable profiles in semantic rank order without exposing projection text.
- Keep a compatible active person corpus current while another generation builds, and keep disabled, unconsented, revoked, or stale person policy from blocking message embedding and activation.

## Why

Durable person profiles can be found by exact fields today, but not by intent such as “a finance expert in New York.” Operators need that query to return the person directly without routing person IDs through the message index, exposing projection text, or silently sending additional profile data under an older message-embedding configuration.

## Usage

Configure the separate people policy, inspect its exact disclosure, and consent:

```toml
[vector.people]
enabled = true
retention_posture = "zdr"
training_posture = "opted-out"
```

```sh
msgvault person provider status --semantic-embeddings
msgvault person provider consent --semantic-embeddings --yes
msgvault embeddings resume --backstop
```

Then search through the daemon:

```sh
msgvault person search "finance expert in New York"
msgvault person search --limit 10 --json "finance expert in New York"
```

The default limit is 20; `--limit` accepts 1–100. Each search query is embedded by the disclosed provider. Changing any fingerprinted provider policy requires new consent. `[vector.embed.scope]` remains message-only; semantic person consent explicitly covers all durable curated people.

Refs #534

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -6980,6 +6980,43 @@ components:
       required:
         - relationships
       type: object
+    PersonSearchRequest:
+      additionalProperties: false
+      properties:
+        limit:
+          default: 20
+          format: int64
+          maximum: 100
+          minimum: 0
+          type: integer
+        query:
+          minLength: 1
+          type: string
+      required:
+        - query
+      type: object
+    PersonSearchResponse:
+      additionalProperties: true
+      properties:
+        results:
+          items:
+            $ref: "#/components/schemas/PersonSearchResult"
+          type: array
+      required:
+        - results
+      type: object
+    PersonSearchResult:
+      additionalProperties: true
+      properties:
+        person:
+          $ref: "#/components/schemas/Person"
+        score:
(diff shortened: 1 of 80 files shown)


mariusvniekerk:
Very cool.

salmonumbrella:
@wesm ready for you

PR #658, opened 2026-08-20
Retrieve person files across metadata, documents, and visual search

### LINKED ISSUE #623: Expose person-scoped attachment retrieval through MCP and the CLI

salmonumbrella:
## What changed

- adds one shared person-scope contract across attachment metadata, extracted documents, and visual search
- adds `msgvault person files <person-id>` with explicit metadata, documents, visual, and all-lane behavior
- adds the always-available MCP `search_person_files` tool and person filters on document and visual retrieval
- keeps exact attachment, message, conversation, source, matched identity, role, and direction provenance

## Why

The underlying indexes stay independent, but callers should not have to translate a durable person into separate participant filters for each retrieval lane. This gives them one stable person-scoped interface over the Docbank-backed document and visual indexes plus authoritative metadata.

This is stacked on #650 so the visual lane uses the shared visual index landing there.

## Usage

```bash
msgvault person files 42
msgvault person files 42 --lane documents --query "inspection notes"
msgvault person files 42 --lane visual --query "whiteboard diagram"
msgvault person files 42 --lane all --query "signed agreement" --json
```

Closes #623


salmonumbrella:
blocked on #650. once #650 merges, this needs to be rebased onto main and the full CI + RoboRev cycle rerun.

wesm:
Not sure why github closed this, I'll rebase and reopen it

PR #660, opened 2026-08-20
fix(teams): use exclusive cursor for incremental chat sync

### LINKED ISSUE #657: Teams chat sync silently no-ops on every run after the first ($filter uses `ge`, which Graph rejects)
## Summary

`sync-teams` only ever ingests messages on the **first** run for a given
account. Every subsequent run fetches nothing, reports `Teams sync complete!`,
and exits 0 — while every single chat has actually failed.

The cause is one operator. `ListChatMessages` builds its incremental filter
with `ge`:

```go
// internal/teams/client.go:231
url += "&$filter=lastModifiedDateTime%20ge%20" + sinceISO + "&$orderby=lastModifiedDateTime%20desc"
```

Microsoft Graph does not accept `ge` on `lastModifiedDateTime` for
`/chats/{id}/messages`. It returns **400 BadRequest**:

> The entity property 'lastModifiedDateTime' and operationKind
> 'GreaterThanOrEqual' is not allowed in $filter query.

Only `gt` and `lt` are accepted. Since the cursor is non-empty on every run
after the first, every incremental request 400s.

## Reproduction

Against a real tenant, msgvault at `3806380c`:

```
$ msgvault sync-teams me@tenant.com --no-channels     # first run
  Chats:           36
  Messages added:  1166
  Errors:          10

$ msgvault sync-teams me@tenant.com --no-channels     # second run
  Chats:           0
  Messages added:  0
  Errors:          46          <-- 41 chats × 1 + 5 pre-existing
```

Exit status is 0 both times.

The 400 is reproducible directly against Graph, isolating it from everything
else in the pipeline:

| Request | Result |
|---|---|
| `?$top=50` (no filter — the first-run shape) | 200 |
| `?$top=50&$filter=lastModifiedDateTime ge <ts>&$orderby=...` (**msgvault's incremental shape**) | **400 BadRequest** |
| `?$top=50&$filter=lastModifiedDateTime gt <ts>&$orderby=...` | 200 |
| `?$top=50&$filter=lastModifiedDateTime ge <ts>` (no `$orderby`) | **400 BadRequest** |
| `?$top=50&$filter=lastModifiedDateTime gt <ts>` | 200 |
| `?$top=50&$filter=lastModifiedDateTime gt <ts with ns precision>` | 200 |

So `$orderby` is not involved and nanosecond cursor precision is fine. The
operator is the whole problem.

## Fix

Change `ge` to `gt` at `internal/teams/client.go:231`.

This also changes the cursor from inclusive to exclusive: a message whose
`lastModifiedDateTime` exactly equals the stored cursor will be skipped.
Given nanosecond precision on the cursor, exact ties are vanishingly rare —
and an exclusive cursor that works is strictly better than an inclusive one
that 400s. If genuinely inclusive semantics are wanted, the cursor would need
to be rewound by a tick before being formatted into the filter.

Verified on a real tenant — same account, same 41 chats, immediately after
the failing run above:

```
  Chats:           36     (was 0)
  Messages added:  0      (correct: nothing new since the first sync)
  Errors:          10     (was 46; the residual 10 are unrelated, see below)
```

## Why the test suite did not catch this

`TestListChatMessagesUsesInclusiveCursor`
(`internal/teams/client_test.go:173`) asserts the `ge` form against an
`httptest` server that echoes back whatever `$filter` it is given. The mock
accepts a query that real Graph rejects outright, so the test passes while the
feature is 100% broken in production. The test name states the intent
("inclusive cursor"), which is exactly the intent Graph will not honor here.

Any fix should flip that test to `gt` and rename it, so the suite stops
asserting an impossible contract.

## Secondary finding: the failure is completely silent

This bug is only invisible because of a second issue, which may be worth
splitting into its own report.

Every `sum.Errors++` in `syncChats` (`internal/teams/importer.go`) increments a
counter and moves on — no `logger.Warn`, no `logger.Error`, no error detail
retained anywhere. Concretely, in the failing state:

- 100% of chats fail with HTTP 400
- the summary prints `Teams sync complete!` and `Errors: 46`
- the run logs `outcome=ok` and the process exits **0**
- nothing anywhere records *what* failed, on which chat, or with what status

A user has no way to tell "nothing new to sync" from "every request failed".
Automated/scheduled syncs would report success indefinitely while archiving
nothing.

Suggested minimum: log each counted error at WARN with the chat ID and the
Graph status/code, and consider a non-zero exit (or at least a distinct
summary line) when a run's error count is a large fraction of the chats
attempted.

## Note on the residual 10 errors

Unrelated to this bug and **not** a msgvault defect — recorded so the numbers
above reconcile. Five chats in the test tenant return **403
InsufficientPrivileges** on both `/messages` and `/members`, and all five carry
`"chatType": "unknownFutureValue"` — a chat type newer than the Graph v1.0
enum. Two endpoints × five chats = the 10 errors present in every run. They
would, however, be far easier to diagnose if the secondary finding above were
addressed.

## Environment

- msgvault `3806380c` (v0.19.3-32), built from source, `linux/amd64`, Go 1.26.6
- Delegated Graph auth via `add-teams`, single-tenant app registration
- 41 chats, `--no-channels` (the channel paths were not exercised)

exactmike:
Teams chat sync only ingests messages on the first run for an account.
Every later run fails on all chats and reports success.

`ListChatMessages` built its incremental filter with `lastModifiedDateTime ge
<cursor>`. Microsoft Graph does not allow `ge` on that property for
`/chats/{id}/messages` and returns 400 BadRequest:

> The entity property 'lastModifiedDateTime' and operationKind
> 'GreaterThanOrEqual' is not allowed in $filter query.

Only `gt` and `lt` are accepted, and the cursor is non-empty on every run after
the first, so every incremental request failed.

This switches the filter to `gt`. Graph accepts nothing else here, so the
cursor is necessarily exclusive, and a message sharing the cursor's exact
`lastModifiedDateTime` would be skipped on every later sync. Graph timestamps
are millisecond-resolution, so such ties are possible rather than negligible.
The incremental query therefore starts a one-second overlap before the stored
cursor and relies on idempotent persistence to absorb the re-read; the
summary's added-message count is resolved by identity so the overlap does not
inflate it.

`TestListChatMessagesUsesInclusiveCursor` asserted the `ge` form against a mock
that echoes back any `$filter` it is given, so it passed while the request
shape was rejected by the real API. It is renamed and flipped to `gt` so the
suite stops asserting a contract Graph will not honor.

Behavior change for users: `sync-teams` now picks up new chat messages on runs
after the first. No migration or re-auth needed; existing cursors are reused
as-is.

Fixes #657

https://claude.ai/code/session_01BT68iRsMWaYRd9hz7Yw7L4

### DIFF
--- a/internal/teams/client.go
+++ b/internal/teams/client.go
@@ -228,7 +228,12 @@ func (c *Client) ListChannels(ctx context.Context, teamID string) ([]Channel, er
 func (c *Client) ListChatMessages(ctx context.Context, chatID, sinceISO string, limit int) ([]ChatMessage, bool, error) {
 	url := "/me/chats/" + chatID + "/messages?$top=50"
 	if sinceISO != "" {
-		url += "&$filter=lastModifiedDateTime%20ge%20" + sinceISO + "&$orderby=lastModifiedDateTime%20desc"
+		// Graph rejects "ge" on lastModifiedDateTime for /chats/{id}/messages
+		// with BadRequest ("operationKind 'GreaterThanOrEqual' is not allowed
+		// in $filter"); only gt/lt are accepted. The cursor is the newest
+		// lastModifiedDateTime already ingested, so gt is also the correct
+		// boundary.
+		url += "&$filter=lastModifiedDateTime%20gt%20" + sinceISO + "&$orderby=lastModifiedDateTime%20desc"
 	}
 	var out []ChatMessage
 	_, truncated, err := pageThroughLimit[ChatMessage](ctx, c, url, limit, func(p []ChatMessage) { out = append(out, p...) })
--- a/internal/teams/client_test.go
+++ b/internal/teams/client_test.go
@@ -170,7 +170,11 @@ func TestListChatsAndMessages(t *testing.T) {
 	assert.Equal("m1", msgs[0].ID)
 }
 
-func TestListChatMessagesUsesInclusiveCursor(t *testing.T) {
+// Graph rejects "ge" on lastModifiedDateTime for /chats/{id}/messages with
+// BadRequest, so the cursor is necessarily exclusive. A message whose
+// lastModifiedDateTime exactly equals the stored cursor is therefore skipped;
+// the cursor carries nanosecond precision, so exact ties are vanishingly rare.
+func TestListChatMessagesUsesExclusiveCursor(t *testing.T) {
 	assert := assert.New(t)
 	var filter string
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -183,5 +187,5 @@ func TestListChatMessagesUsesInclusiveCursor(t *testing.T) {
 	_, _, err := c.ListChatMessages(context.Background(), "19:x@thread.v2", "2025-01-01T00:00:00Z", 0)
 	require.NoError(t, err)
 
-	assert.Equal("lastModifiedDateTime ge 2025-01-01T00:00:00Z", filter)
+	assert.Equal("lastModifiedDateTime gt 2025-01-01T00:00:00Z", filter)
 }
--- a/internal/teams/importer.go
+++ b/internal/teams/importer.go
@@ -158,6 +158,31 @@ func (imp *Importer) BackfillInlineMedia(ctx context.Context, opts ImportOptions
 	return sum, err
 }
 
+// chatCursorOverlap widens the incremental chat window backwards before
+// querying. Graph only accepts an exclusive "gt" filter on
+// lastModifiedDateTime, so a message sharing the cursor's exact timestamp
+// would be skipped forever; Graph timestamps are millisecond-resolution, so
+// such collisions are possible. Re-reading a small overlap costs a handful of
+// messages per chat and is free of side effects: persistence upserts on
+// (source_id, source_message_id), and maxTime is seeded from the stored cursor
+// so a re-read of older messages cannot move it backwards.
+const chatCursorOverlap = time.Second
+
+// chatQuerySince rewinds a stored chat cursor by chatCursorOverlap. An empty
+// or unparseable cursor is passed through untouched, so a first sync stays
+// unfiltered and a malformed cursor degrades to the previous behavior rather
+// than dropping the filter entirely.
+func chatQuerySince(cursor string) string {
+	if cursor == "" {
+		return ""
+	}
+	t, err := time.Parse(time.RFC3339Nano, cursor)
+	if err != nil {
+		return cursor
+	}
+	return t.Add(-chatCursorOverlap).Format(time.RFC3339Nano)
+}
+
 func (imp *Importer) syncChats(ctx context.Context, sourceID, syncID int64, opts ImportOptions, state *SyncState, sum *ImportSummary) error {
 	chats, err := imp.client.ListChats(ctx)
 	if err != nil {
@@ -205,7 +230,7 @@ func (imp *Importer) syncChats(ctx context.Context, sourceID, syncID int64, opts
 		}
 
 		since := state.ChatCursor(ch.ID)
-		msgs, pageTruncated, err := imp.client.ListChatMessages(ctx, ch.ID, since, opts.Limit)
+		msgs, pageTruncated, err := imp.client.ListChatMessages(ctx, ch.ID, chatQuerySince(since), opts.Limit)
 		if err != nil {
 			sum.Errors++
 			continue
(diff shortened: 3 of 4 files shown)


exactmike:
Good catch — the nanosecond claim in my description was wrong, and I've fixed
the PR body as well.

`RFC3339Nano` only preserves whatever precision Graph sends; it does not create
it. I measured against a live tenant — 246 messages across 11 chats — and
`lastModifiedDateTime` comes back at **millisecond** resolution (1-3 fractional
digits, never more). So ties are possible, and with a strict `gt` filter a tied
message would be filtered out on every later sync rather than just once.

Fixed in `0e5f9f36` along the lines you suggested: the incremental query now
starts one second before the stored cursor. Re-reading that overlap is
side-effect free — persistence upserts on `(source_id, source_message_id)`, and
`maxTime` is seeded from the stored cursor, so re-reading older messages cannot
walk it backwards.

Two notes on the details:

**Existing guard.** `syncChats` only advances the cursor when
`chatComplete && !truncated`, so a partially-read or limit-truncated chat never
moves it. That already ruled out the pagination variant of this bug; what
remained was a message arriving at the boundary timestamp after a complete
read, which the overlap now covers.

**Reporting.** The overlap deliberately re-reads, so "persisted" stopped
implying "new" and `MessagesAdded` reported a nonzero count on every
incremental sync that changed nothing — on my tenant, 36, one per chat. The
commit resolves each batch through `MessageExistsBatch` and counts only
genuinely new messages; a live incremental sync now reports 0 added with 1166
rows and 1166 distinct source IDs. Identity is the only workable test here: a
message sharing the cursor timestamp may be a boundary re-read *or* the exact
tie the overlap exists to recover, and a timestamp comparison cannot separate
the two.

The behavioral test you asked for is in `0e5f9f36`. Its fake Graph server
applies the `$filter` instead of echoing it, so it fails if the overlap is
removed — verified by setting the overlap to zero, which loses the tied message
permanently.

exactmike:
Thanks — I dug into this one and I think the mechanism is right but the
severity is lower than stated. I've added the regression test you asked for,
in `76de1aa6`.

**The scenario is real.** With `--limit`, the overlap is fetched under the cap,
and since tied messages have no deterministic secondary ordering, a capped run
can admit preexisting boundary messages and defer a newly arrived tied one.

**It is not data loss, though, and not indefinite.** `pageThroughLimit` reports
truncation whenever it stops short of the available results, and `syncChats`
only advances the cursor when the read was both complete and untruncated. So a
capped run that defers the tied message also leaves the cursor where it was —
the message stays inside the query window, and any unlimited run picks it up.
"Indefinitely skipped" holds only for an operator who runs `--limit` forever,
and such an archive is incomplete by construction for every message past the
cap, tied or not.

That distinction matters for the fix choice: paging the full boundary before
applying the cap, or adding a secondary cursor, would buy completeness for
capped runs specifically, at the cost of making `--limit` no longer a hard cap
(the boundary group can be arbitrarily large) or of a schema change to the
stored cursor. Neither seems worth it to protect a mode that already does not
progress. Happy to be overruled if you'd rather have it.

The new test asserts what the code actually guarantees: three capped syncs
with the cap smaller than the number of tied messages, asserting the cursor
never advances past the deferred message, then an unlimited sync that archives
it. It also asserts the deferral is actually occurring, so it cannot quietly go
vacuous.

One thing your review made me check that is worth recording: the overlap
re-runs the whole persist path on boundary messages every sync, so I verified
it is cheap. Inline images are reused when a content hash already exists rather
than re-fetched, and `embed_gen` is preserved by the upsert unless subject or
message type changed, so re-reads do not re-embed. A live incremental sync
reports 0 added, 0 attachments, 0 inline images across 36 chats.

exactmike:
This finding was raised on `0e5f9f3`, cleared on `76de1aa` ("No issues found"),
and has come back on `3bb573c`. `3bb573c` is a **test-only amend** of `76de1aa` —
`git diff 76de1aa6 3bb573c3 -- ':!*_test.go'` is empty. It rewrites the fake
Graph server to report a rejected `$filter` from the test body instead of calling
`require` inside its own handler, which is what `testifylint`'s `go-require` rule
was failing CI on. No production line changed between the two reviews.

On the substance, which I do not want to duck: **the behaviour you describe is
real, and the test asserts it deliberately.** With `--limit 1` and three messages
sharing a timestamp, repeated capped runs do not import the late one —
`TestLimitedChatSyncDoesNotStrandTiedMessages` makes exactly that assertion, and
guards it against going vacuous.

Where I still disagree is that this is a defect in this change rather than a
property of `--limit`:

- **Nothing is stranded permanently.** `pageThroughLimit` returns
  `truncated = true` whenever it stops short of the available results
  (`internal/teams/client.go:190-192`), and `syncChats` advances the cursor only
  when `chatComplete && !truncated` (`importer.go:285`). A capped run therefore
  cannot move the cursor past a deferred tie; the tie stays in range and the next
  unlimited run archives it. The test asserts both halves — cursor held, then
  recovered.
- **The suggested fix does not fit inside the cap as it is applied.** `opts.Limit`
  is passed to `ListChatMessages`, so it bounds the fetch, not just the persist
  loop — the overlap rows are already the only rows the client asked for.
  Continuing pagination past them means decoupling the fetch budget from the
  persist budget in the shared `pageThroughLimit`, which every Teams caller uses.
  That is worth doing, but it is a different change from this one, whose
  production diff is a `ge`→`gt` operator fix and a cursor rewind.
- **No steady-state path can reach this.** `ImportOptions.Limit` has exactly one
  setter — the `sync-teams --limit` flag, default `0` = unlimited
  (`cmd/msgvault/cmd/sync_teams.go:142`). Nothing in the daemon or scheduler sets
  it, so scheduled syncs are always unlimited. The docs scope it to a pre-flight
  check: *"We recommend starting with `--limit` or a date range to verify
  everything works before kicking off the full run"* (`docs/setup.md:211`).
  Reaching the stranded case means re-running `--limit N` with N below the number
  of messages tied at the cursor and never once running the full sync the flag
  exists to precede.

Happy to open a follow-up issue for the fetch/persist budget split so it does not
get lost — or to fold it in here if you would rather see it land together.

Issue #634, opened 2026-08-18
Ship recommended defaults and a setup pass that turns on the people and attachment stack


salmonumbrella:
## The situation

By the end of the #534 sequence an archive has four retrieval lanes on three providers, scoped media collection, a typed people layer with a scheduled sweep, and an MCP surface over all of it. Each piece shipped correctly as opt-in, with its own config table, its own key variable, its own consent step, and its own build command. Nothing chooses for the operator, and nothing turns on by itself.

This is what a fresh install has to assemble by hand today, taken from `main` and the open PR heads:

- **Text embeddings** (`[vector]`, `[vector.embeddings]`): `enabled = false`; `endpoint`, `model`, and `dimension` are all required with no default; `api_format` defaults to `openai` and accepts `voyage-contextual`, which pins the model to `voyage-context-4` (#589). `api_format` is not documented anywhere under `docs/`. The documented example is Ollama with `nomic-embed-text`. `voyage-4-large` appears only in the eval harness. `[vector.embed.schedule]` is `cron = ""`, `run_after_sync = false`.
- **Chat windows and meeting turns** ride the same generation through `voyage-contextual` (#589); the contextual endpoint is Voyage-only, so an OpenAI-compatible key gets per-message embeddings and no conversation context.
- **Visual attachments** (#611, `[vector.multimodal]`): `enabled = false`; provider `voyage` is the only legal value; endpoint, `VOYAGE_API_KEY`, `voyage-multimodal-3.5`, and dimension 1024 are already sensible defaults, but a key alone enables nothing, consent is per generation via `multimodal build --yes`, `schedule.cron = ""`, and the PR adds no docs.
- **Documents** (#616, `[attachments.documents]`): `enabled = false`; provider `mistral`, region `eu`, `mistral-ocr-4-0`, `MISTRAL_API_KEY`; `retention_posture` and `training_posture` default to `unknown` and must be set; a schedule requires a `capability_manifest` and a pricing assumption; consent is `documents consent-mistral` against the exact manifest fingerprint; the embeddings sub-table is reserved for #617; no docs. After #633 the extraction and probes come from Docbank, but the msgvault-side config and consent stay.
[...]

PR #599, opened 2026-08-11
feat(activity): add dated contact activity

salmonumbrella:
## What changed

- Adds portable dated activity projection for message, chat, and meeting evidence, with source-aware dates and owner-relative direction.
- Materializes contact recency, cadence, daily aggregates, and authored daily notes for bounded person and day queries.
- Adds REST, OpenAPI, generated client, and `msgvault activity build` surfaces with SQLite and PostgreSQL parity.
- Keeps projection work and contact state consistent across identity changes, message mutations, deletions, and cache builds.

## Why

People and relationship views need reliable recency and cadence without scanning the full archive on every request. The archive also contains mixed date quality and source-specific ownership signals, so this needs one durable, explainable projection instead of separate query-time heuristics.

## Usage

Run the incremental projection:

```sh
msgvault activity build
```

Force a complete archive backstop scan when needed:

```sh
msgvault activity build --backstop
```

The scheduled incremental job uses the `[activity]` configuration and runs at `17 * * * *` by default.

Refs #534

PR #751, opened 2026-09-03
feat(mcp): add get_person_profile

salmonumbrella:
## What changed

- Add one read-only MCP tool, `get_person_profile`, that returns a durable person's overview from local derived state: display name, tracking state, the deterministic contact state (first and last contact, last inbound and outbound, interaction count, inferred channel), the curated `primary_channel`, non-sensitive attributes, current employment with organization names, typed relationships, contact points, dates, and categories.
- Exclude sensitive attributes, private Notes, addresses, and media, and name those exclusions in the response so a client knows what it did not get.
- Compose the read in the daemon client from the existing authenticated person routes behind a new optional `peoplebrowser.ProfileReader` surface. Absent sub-resources degrade to empty values; an unknown person keeps the daemon's `person_profile_not_found` error. No new server route, generated client, or provider call.

## Why

With no hosted lane configured, the MCP server could search people and read Notes or relationship temperature, but nothing answered "who is this", "when did we last talk", or "which network do I reach them on". The activity projection and the typed people layer already hold that state locally; this exposes it through one tool that is on whenever the people backend is.

Refs #634, refs #599.

## Usage

```json
{"name": "get_person_profile", "arguments": {"person_id": 51}}
```

Find the ID with `search_people` first, or create one with `promote_person`. The `contact_state` field is `null` until the activity job has computed a row for the person; `tracked` is `null` when the daemon cannot report tracking state.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,5 @@
 ---
-last_edited: 2026-09-01
+last_edited: 2026-09-03
 title: Changelog
 description: Release history for msgvault
 ---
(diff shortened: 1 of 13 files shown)


salmonumbrella:
CI note: the `run / test-postgres` failure on 1018dd02 is `TestCodexCleanupReturnsBoundedSafeErrorWhenKillFailsAndWaitBlocks` in `internal/peoplesweep` (a wall-clock bound: 773ms observed against a 500ms limit under runner load). That commit only changed a daemonclient test file; the same job passed on the previous commit b8f68a09. I lack rerun rights on this repository, so a maintainer rerun of the failed job is the fix.

PR #759, opened 2026-09-03
feat(cli): stage deletions from search criteria

### LINKED ISSUE #191: CLI command to stage deletions from search results
Currently the only way to stage messages for deletion is through the TUI (select → `d`). It would be useful to have a CLI command that stages deletions from search results, e.g.:

```bash
msgvault stage-delete "from:newsletter@example.com older_than:1y"
msgvault stage-delete --ids 123,456,789
```

This would enable:
- Scripted/automated cleanup (e.g. in cron jobs chained with `delete-staged --trash --yes`)
- Bulk operations on large result sets without manually selecting in the TUI (related: #41)
- Integration with AI agents that can search and triage but can't drive the TUI

The existing `delete-staged` execution flow (stage → review → execute) would stay intact — this just adds a non-interactive way to create the staging manifest.

rodboev:
`msgvault stage-delete` stages active messages for deletion from search criteria, with `--dry-run` output and optional `--source-id` narrowing for searches that span multiple sources. Messages already deleted at the source are excluded, and an exact source selector keeps each pending deletion batch tied to one source. The existing review and execution flow is preserved.

The daemon already provides revision-pinned search, preflight, and staging, but the CLI previously had no non-interactive entry point for that safe path. `stage-delete "from:newsletter@example.com older_than:1y"` stages the matches; for a search spanning multiple sources, stage it once per source with `--source-id 42`. Review the batch with `show-deletion`, then use the existing `delete-staged` flow.

Closes #191

### DIFF
--- a/cmd/msgvault/cmd/stage_delete.go
+++ b/cmd/msgvault/cmd/stage_delete.go
@@ -199,13 +199,25 @@ func runStageDelete(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-// stageDeleteDaemonErr turns the daemon's structured cache-unavailable
-// response into an actionable message instead of a bare API error.
+// stageDeleteDaemonErr turns the daemon's structured rejections into
+// actionable messages instead of bare API errors.
 func stageDeleteDaemonErr(op string, err error) error {
 	var apiErr *daemonclient.APIError
-	if errors.As(err, &apiErr) && apiErr.APIErrorCode() == analyticalCacheUnavailableCode {
+	if !errors.As(err, &apiErr) {
+		return fmt.Errorf("%s: %w", op, err)
+	}
+	switch apiErr.APIErrorCode() {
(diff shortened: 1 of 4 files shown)


rodboev:
Both findings are correct. `stage-delete` now limits selection to active messages, so source-deleted matches can't invalidate an otherwise valid batch. It also accepts `--source-id`, allowing a search that spans sources to be staged once per source.

The query parser, revision-pinned preflight, dry-run behavior, and separate `show-deletion` and `delete-staged` flow stay unchanged.

rodboev:
The List-ID compatibility check now runs before Explore. `stage-delete` rejects `list:` and `list-id:` queries when the selected daemon reports an API schema older than 2.14.0, so an older daemon can't reinterpret the request. Compatible daemons continue through the existing active and optional source-filtered selection flow.

PR #761, opened 2026-09-03
fix(imap): retry connection failures with bounded exponential backoff

### LINKED ISSUE #199: IMAP: retry with exponential backoff on connection failures
## Context

The IMAP client in #193 makes a single reconnect attempt on network errors. For large mailboxes synced over unreliable connections (or servers that throttle aggressive clients), this can cause sync failures that require manual re-runs.

## Proposal

Add configurable retry with exponential backoff for transient network errors:
- 3 attempts with backoff (e.g. 5s, 15s, 45s)
- Only retry on connection-level errors (EOF, reset, timeout) — not auth or protocol errors
- Log each retry attempt at warn level

The retry logic could live in `withConn()` or a wrapper around the connect/reconnect path.

## References

- Yahoo Mail PRD: #93
- IMAP support: #193

rodboev:
IMAP connection establishment now retries transient TCP failures and empty-greeting transport failures in plaintext, implicit-TLS, or STARTTLS mode on a bounded 5s, 15s, 45s schedule. TLS negotiation remains with `crypto/tls`, the pinned IMAP client owns greeting and STARTTLS protocol handling, and LOGIN or XOAUTH2 runs once after establishment succeeds.

The retry boundary records inbound plaintext IMAP progress and, for STARTTLS, separates the completed protocol response from the following TLS handshake transport state, so encrypted TLS handshake records cannot be mistaken for a partial greeting. A fresh connection is opened only before any inbound server greeting byte, or after a completed STARTTLS response with transient TLS transport loss, each retry emits a warning, and context cancellation interrupts connection work or backoff. Established-session reconnects and command replay retain their existing behavior, and there are no usage or configuration changes.

Closes #199


rodboev:
Fixed in `dbeb8270`. The STARTTLS constructor error path in `internal/imap/client.go:484-487` now sends an empty transient greeting through the existing greeting retry classifier before the STARTTLS command is issued. Protocol and truncated-greeting failures remain non-retryable. `internal/imap/retry_test.go:229-249,400-427` adds a production-path regression test covering a dropped initial greeting followed by a protocol rejection.

rodboev:
Fixed in `cabae0fb`. `connectOnce` now honors `dialIMAP`'s explicit retryable result for raw dial, TLS handshake, and STARTTLS setup failures. STARTTLS observation preserves pre-command bytes, so a clean close before the greeting retries while truncated greetings remain terminal; clean and reset after-command cases plus implicit TLS and STARTTLS handshakes are covered by focused tests. Authentication, protocol rejection, and established-session replay behavior remain unchanged.

rodboev:
Fixed in `c7eb8dad`. The connection-stage classifier now treats wrapped `EPIPE` and the native Windows broken-pipe value as transient alongside the existing reset and abort errors, so dropped STARTTLS writes reach the bounded retry loop. The test covers all accepted transport values and keeps connection refusal, cancellation, generic text, authentication, TLS certificate, protocol, and truncated-greeting failures non-retryable.

Issue #803, opened 2026-09-08
Gmail profile 429 can block scheduled sync for 10–20 minutes before incremental sync starts


johnnieball:
### Version

Observed with msgvault v0.17.1 on macOS arm64. The relevant retry pattern also appears to remain in current `internal/gmail/client.go`.

### Problem

When `GET /gmail/v1/users/me/profile` returns HTTP 429, an incremental Gmail sync can occupy the scheduler for 10–20 minutes before failing or yielding. No structured `incremental sync` event is emitted because the profile request happens first.

Example:

    2026-09-07T09:25:00.032+01:00 msg="starting scheduled sync" email=hello@johnnieball.com
    2026-09-07T09:43:04.827+01:00 msg="scheduled sync failed" email=hello@johnnieball.com duration=18m4.783661666s error="hello@johnnieball.com (gmail): incremental sync failed: get profile: max retries exceeded: rate limited (429)"

A manually invoked incremental sync remained at:

    Starting incremental sync for hello@johnnieball.com
    Last history ID: 2639050

until it was terminated after 120 seconds.

An unauthorised control request to the same endpoint returned 401 in 0.088 seconds, so DNS, TLS and basic transport were healthy.

### Cause

`Client.request` retries 429 responses up to `maxRetries = 12` using exponential full-jitter back-off capped at 600 seconds. There is no overall deadline covering the HTTP call and retry sleeps.

`defaultTimeout = 30 * time.Second` exists, but `NewClient` constructs the client with:

    oauth2.NewClient(context.Background(), tokenSource)

and the timeout is not applied to the complete request lifecycle.

The 429 response body and `Retry-After` header are also discarded, which prevents diagnosis of the actual Google quota reason.

### Expected behaviour

A scheduled Gmail operation should fail within a bounded interval when its initial profile request is repeatedly throttled, allowing other scheduled work to continue.

### Suggested smallest change

When `Client.request` receives a context without an earlier deadline, derive one with `context.WithTimeout(ctx, defaultTimeout)` and use it for both `http.NewRequestWithContext` and the retry-back-off waits.

Please add a regression test using a transport that repeatedly returns 429 and assert that `GetProfile` exits within the deadline.

It would also help to preserve the Gmail error reason and honour or report `Retry-After`, without logging credentials or tokens.

PR #806, opened 2026-09-08
fix(gmail): bound requests and OAuth refresh waits

salmonumbrella:
Gmail profile and metadata calls now share a 30-second deadline across HTTP calls and retry backoff, preventing repeated 429 responses from delaying scheduled work for many minutes. Raw message downloads get five minutes. Earlier caller deadlines and cancellation still apply.

The request budget starts after the shared rate limiter grants tokens. Quota pauses of 30 or 60 seconds remain subject to the caller’s context, allowing full sync to wait and fetch the next page after a quota response. The new request timeout does not cap that initial wait, and retries do not restart the budget.

Google OAuth token sources cap each token-endpoint HTTP call at 30 seconds, including service-account exchanges and later refreshes. Shorter timeouts and custom HTTP settings are preserved without mutating the supplied client. This cap is independent of the Gmail request deadline; authentication-style detection can make two calls. Other Google clients using these token managers receive the same cap. Externally supplied token sources retain their own refresh behavior.

No configuration changes are required. This addresses the request and refresh timeout portion of #803. Preserving Gmail quota error reasons and honoring or reporting `Retry-After` remain follow-up work.

Refs #803

### DIFF
--- a/internal/gmail/client.go
+++ b/internal/gmail/client.go
@@ -93,7 +93,13 @@ func (c *Client) Close() error {
 // request makes an HTTP request with rate limiting and retry logic.
 // bodyBytes can be nil for requests without a body.
 func (c *Client) request(ctx context.Context, op Operation, method, path string, bodyBytes []byte) ([]byte, error) {
-	// Share one budget across rate limiting, HTTP I/O and retry backoff.
+	// Quota pauses can exceed the request timeout. Wait using the caller's
+	// context so a previous request's throttle does not exhaust this budget.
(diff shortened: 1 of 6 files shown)


salmonumbrella:
The refresh observation is correct. #803 targets authenticated Gmail 429 retries; token acquisition still uses its creation-time context. The code and PR body now state that boundary explicitly.

The tests now retain the production OAuth transport and verify the bearer token on each retry. Bounding refresh itself needs a separate change to token-source construction; wrapping the contextless `Token()` call in a goroutine would leave a stalled refresh running.

PR #808, opened 2026-09-08
fix(imap): refresh edited outgoing copies after relocation

### LINKED ISSUE #804: imap: body keeps the draft text after a message moves Drafts → Sent
### Summary

When an IMAP message is edited between being saved as a draft and being sent, `message_bodies.body_text` keeps the **pre-edit** text while the row's `source_message_id` and labels advance to the **Sent** copy. The row therefore presents itself as the sent message but carries words that were never sent. Metadata (date, recipients, subject, labels, attachments) is correct throughout — only the body is stale.

Version: **v0.18.0**. I checked `v0.18.0..v0.19.3` and found no commit touching this path, so I believe it is still present at HEAD.

### Reproduction

1. Save a draft in Outlook/O365 (or APPEND one over IMAP), let `msgvault sync` run.
2. Edit the message — even one word — and send it.
3. Delete the draft, run `msgvault sync` again.
4. Read the message back from the store, then `FETCH` the same UID from the Sent folder over IMAP.

Observed on a real row (content redacted):

```
row id 35526
  source_message_id : "Gesendete Elemente|48533"     <- the Sent copy
  rfc822_message_id : <FR2PPFF...@...PROD.OUTLOOK.COM>
  body_text         : "Hallo Herr <SURNAME>, ..."    <- the DRAFT wording

IMAP FETCH of UID 48533 in "Gesendete Elemente"
  body              : "Hallo <FIRSTNAME>, ..."       <- what was actually sent
```

Only one row exists for that Message-ID; querying for message-ids present in both a Drafts and a Sent label returns 0, so there is no second row holding the correct body.

### Scale

I compared `message_bodies.body_text` against a live IMAP `FETCH` for every Sent message in a 7-week window (20 Jul – 8 Sep, single O365 account):

| | |
|---|---|
| compared | 544 |
| bodies matching | 518 |
| **genuinely divergent** | **8** |
| (other 26 flagged were my own comparison artefacts — signature/quote trimming and empty bodies) | |

All 8 share one property: the message existed as a draft and was edited before sending. Messages composed and sent in one pass are unaffected. So it is not time-bounded — roughly 1.5% of sent mail, concentrated on exactly the messages worth editing.

### Notes toward a cause

I did not confirm the mechanism, but two things look relevant:

- `upsertMessageSQL` conflicts on `(source_id, source_message_id)` and its `DO UPDATE SET` list covers metadata only. Bodies live in `message_bodies` and are not part of that statement, so nothing in the upsert path can refresh a body.
- `GetDuplicateGroupMessages` ranks duplicate-group survivors on `has_raw`, `label_count`, `is_from_me` and `has_sent_label`. Body freshness is not a criterion, so a merge across a Drafts/Sent pair has no reason to prefer the sent text. This is adjacent to #461.

### Suggested behaviour

When a message's folder or `source_message_id` changes, treat the stored body as stale and re-fetch it — or, at minimum, expose that the body predates the current location so callers can tell.

### Why it matters

The stored text reads as a complete message, so there is no signal that it is wrong. In my case two messages — one quoting a price, one confirming a date — were archived with earlier drafts, and reading the store rather than the mailbox produced a confidently incorrect account of what had been sent.

salmonumbrella:
## What changed

- Refresh the archived snapshot when IMAP adopts an edited copy from a trusted outgoing mailbox, retaining the message ID and updating the body, raw MIME, recipients, attachments, and search index together.
- Refresh from an edited Sent copy while the original Drafts copy still exists, without letting a later stale Drafts copy replace it.
- Recover surviving outgoing copies after the previous location disappears, including incremental syncs where the survivor itself has not changed and full scans where an All Mail mirror is fetched first.
- Preserve archived content when a matching Message-ID comes from an ordinary received-mail folder or All Mail.
- Keep unrelated mail syncing when a relocation target repeatedly fails, preserving the failed message and its retry location until recovery.
- Increase the Windows CLI shard timeout to accommodate its measured suite runtime.

## Why

Sending an edited draft could move the archived message to Sent while leaving its earlier draft text in the archive. A Message-ID match alone cannot authorize replacement because incoming mail can forge that header.

## Usage

Unambiguous server-advertised `\Sent` and `\Drafts` folders are trusted automatically. For a server that does not advertise these roles, configure that account's Sent folder in `config.toml`. Copy the exact `ACCOUNT` value from `msgvault list-accounts`:

```toml
[sync.trusted_imap_sent_mailboxes]
"imaps://user@example.com@imap.example.com:993" = ["Gesendete Elemente"]
```

Trust applies only to the named account. Only configure its Sent folders, and do not include folders that receive incoming mail through filters or filing rules. Advertised Drafts and received-mail roles cannot be overridden by this setting. A folder advertising both Sent and Drafts is ambiguous and remains untrusted. Other relocations preserve the archived snapshot.

This prevents new stale snapshots on supported relocations; it does not repair historical rows that already lost their old location information.

Closes #804.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -395,11 +395,19 @@ jobs:
         include:
           - name: cli
             package: ./cmd/msgvault/cmd
+            # Windows SQLite I/O makes this package's aggregate shard duration
+            # exceed one hour: the passing shard alone ran 406 tests in 3373.8s
+            # (~56m) and the next shard timed out at 60m while still actively
+            # executing schema setup (no deadlock; tests kept progressing).
+            # Give only this package the larger per-shard budget.
+            timeout: 120m
           - name: sync-vector-embed
             package: ./internal/sync
             package2: ./internal/vector/embed
+            timeout: 60m
           - name: vector-sqlite
             package: ./internal/vector/sqlitevec
+            timeout: 60m
     runs-on: windows-latest
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
@@ -428,7 +436,7 @@ jobs:
               -Package $package `
               -ShardCount 4 `
               -Tags "fts5 sqlite_vec" `
-              -Timeout 60m
+              -Timeout ${{ matrix.timeout }}
             if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
           }
 
--- a/cmd/msgvault/cmd/imap_folder_state_sync_test.go
+++ b/cmd/msgvault/cmd/imap_folder_state_sync_test.go
@@ -333,8 +333,6 @@ func TestIMAPFolderStateOptions_ForceRescanRetainsStatesAndEnumerates(t *testing
 	require.NoError(first.Close())
 
 	opts := imapFolderStateOptions(st, src, true)
-	require.Len(opts, 3,
-		"--noresume needs saved identity and alias state plus forced enumeration")
 
 	second := listedIMAPClient(t, addr, opts...)
 	ctx, cancel := context.WithTimeout(
--- a/cmd/msgvault/cmd/imap_qresync_integration_test.go
+++ b/cmd/msgvault/cmd/imap_qresync_integration_test.go
@@ -25,11 +25,16 @@ import (
 )
 
 type scriptedRFC7162Message struct {
-	UID       imapapi.UID
-	MessageID string
-	Subject   string
-	Flags     []imapapi.Flag
-	ModSeq    uint64
+	UID        imapapi.UID
+	MessageID  string
+	Subject    string
+	Body       string
+	MissingRaw bool
+	Flags      []imapapi.Flag
+	ModSeq     uint64
+	// Raw overrides the synthesized message bytes verbatim (already CRLF
+	// encoded) so tests can vary recipients, attachments, and HTML.
+	Raw string
 }
 
 type scriptedRFC7162Mailbox struct {
@@ -405,6 +410,10 @@ func writeScriptedRFC7162Fetch(
 				sequence, uid, formatScriptedRFC7162Flags(message.Flags), modSeq, len(body), body)
 			continue
 		}
+		if message.MissingRaw {
+			_, _ = fmt.Fprintf(w, "* %d FETCH (UID %d FLAGS (%s))\r\n", sequence, uid, formatScriptedRFC7162Flags(message.Flags))
+			continue
+		}
 		raw := scriptedRFC7162RawMessage(message)
 		_, _ = fmt.Fprintf(w,
 			"* %d FETCH (UID %d FLAGS (%s)%s INTERNALDATE \"01-Jan-2024 00:00:00 +0000\" RFC822.SIZE %d BODY[] {%d}\r\n%s)\r\n",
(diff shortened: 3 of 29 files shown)


salmonumbrella:
Addressed in eeecf96292a669e9baf4e08b8e0640f1d2a4aeda.

- Same-key forced relocation candidates are skipped, preserving the existing UIDVALIDITY identity checks and retry behavior.
- A matching Message-ID alone cannot replace archived content. Refresh now requires a trusted outgoing mailbox; received-mail and All Mail adoptions preserve the snapshot. Explicit trust cannot override Inbox or advertised All/Junk/Trash roles. Servers without outgoing-role discovery can use the documented mailbox setting.
- Successful relocation now archives remote images using the retained message ID and refreshed HTML.

The final branch review also caught a liveness problem: a persistently unreadable relocation target could block unrelated mail. Failed targets now remain retryable while independent mail continues; interrupted runs rebuild that protection before ordinary routing.

Regressions cover forged incoming copies, poisoned saved memberships, preferred All Mail copies, edited Drafts/Sent/All Mail flows, localized configured folders, and full-rescan snapshot stability. Mailbox trust remains a placement assumption; folders receiving incoming mail through filing rules should not be trusted.

salmonumbrella:
Confirmed. The location rekey and immediate label reconciliation need to commit together. I am fixing both adoption paths and adding failure-and-retry coverage. The documentation conflict with current main is already resolved in the working branch.

salmonumbrella:
Addressed in ca3c493d5abecda414aa762c04b0ca7fa139ea31, rebased onto current main with one commit.

Both location-only adoption paths now use one context-aware, guarded transaction for the rekey and immediate label reconciliation. A late label-write failure rolls back the identity change, preserving the old key for retry. Deferred labels remain with the mailbox finalizer.

The production regressions fail on the previous code with the old key already changed, then pass with the fix and a successful retry. The shared transaction also has database-level rollback, cancellation, and identity-guard coverage.

salmonumbrella:
I am reproducing the All-before-Sent path from an archived draft snapshot and changing explicit mailbox trust to be scoped to the IMAP account. The account-scoping regression will use two accounts with the same folder name and trust only one of them.

salmonumbrella:
Addressed in 1a1492c544a1d30ff4be1194f828214555e7b646.

- Full-enumeration deduplication now retains one full raw result from a Sent copy after an untrusted mirror. The regression starts with archived draft text and reproduces the failure on the previous head. Stale Drafts remain suppressed, preserving the existing rescan stability checks.
- Explicit trust is now `trusted_imap_sent_mailboxes`, keyed by the exact IMAP account identifier. The shared CLI/daemon setup passes only that account's entry. Tests cover two accounts with the same folder name, configured Drafts, and localized Sent folders.

The PR description includes the new account-scoped configuration syntax.

salmonumbrella:
Addressed in fe77ccc59dcd5d03aa3f859147e3100b84ae36ca.

- An edited Sent copy now refreshes a stale Drafts snapshot while the original draft still exists. Sent precedence is limited to Drafts; a configured name cannot override its advertised Drafts role. The regressions cover incremental sync, full enumeration, localized Sent configuration, and later rescans.
- A LIST entry advertising both Sent and Drafts now denies outgoing and Sent trust, including when explicitly configured. The integration regression checks that location adoption preserves the archived body and raw message.

The existing missing-source recovery tests retain legacy persisted-state fixtures so they still exercise forced fetches, failure isolation, and retry behavior.

Issue #666, opened 2026-08-21
Outbound submission: reply and send through the accounts msgvault already syncs


franklintra:
## Problem

msgvault already holds everything needed to send a correct reply, except the ability to send it.

For every synced account it has the credentials and a live connection, the full RFC822 source in `message_raw`, the thread structure in `conversations` and `messages.reply_to_message_id`, the recipient set in `message_recipients`, and the sending identity in `account_identities`.

Anything that wants to act on archived mail has to rebuild all of that somewhere else: a second credential store, a second IMAP/SMTP connection to the same servers, and its own re-derivation of `In-Reply-To` and `References` from raw source msgvault has already parsed. That duplication is the whole cost of every "email agent" integration built on top of an archive, and it is avoidable here.

## Proposal

An outbound path that composes from the archive and emits through the account the message belongs to.

**Endpoints** (mirrored in the MCP server and the CLI):

- `POST /api/v1/messages/{id}/reply` — build a reply to an archived message. Server sets `In-Reply-To`, the full `References` chain, `Re:` subject normalisation, and the recipient set for reply or reply-all. `From` resolves through `account_identities` for the source that message arrived on.
- `POST /api/v1/messages/{id}/forward` — same, preserving attachments from `attachments` rather than re-uploading them.
- `POST /api/v1/send` — fresh compose against a named source.

**Two modes, draft first:**

- `mode=draft` (default) does an IMAP `APPEND` to the account's Drafts mailbox with `\Draft`, and returns the resulting message id. The draft syncs to every client on that account, so review happens wherever the user already reads mail.
- `mode=send` performs SMTP submission, then `APPEND`s to Sent for providers that do not do it server-side.

**Coherence:** on success the outbound message is recorded in the store immediately, rather than waiting for the next sync poll to discover it. Without this, anything acting through the API sees its own writes disappear for one poll interval.

## Why this belongs here rather than in a separate tool

`internal/deletion/` already writes back to the source: `executor.go` deletes messages from the provider with manifests, batching and retry. So outbound is not a new class of behaviour for the project, and the safety machinery for "act on the real mailbox" already exists and has a shape worth reusing.

Everything else about a correct reply — identity selection, threading headers, attachment bytes, the connection itself — is already implemented for sync. Submission is the missing verb, not a missing subsystem.

## Scope and safety

The README's pitch is a complete local copy where "everything runs offline" and "no network access required". Submission should not weaken that, so:

[...]

PR #833, opened 2026-09-11
feat(api): restricted agent access for reply drafts

rodboev:
You can now issue an account-scoped credential and hand it to an agent, which then creates reply drafts through `msgvault draft-reply` against your daemon. `msgvault agent-token issue`, `list`, and `revoke` manage those credentials, and they need the owner API key.

A grant names the `draft.create` permission and the accounts it applies to, matched by the same portable source tuple a deletion manifest uses. The daemon checks it when the request arrives. Grants are immutable, live in memory, and do not expire automatically; they persist until revoked or the daemon restarts.

Delegation stays off until `[server] agent_access` is set in `config.toml`, which also requires an owner API key. That key is absent from the settings API, so no HTTP caller can enable it. Presenting an agent token commits the request to the delegated path, where an unknown or revoked one fails outright rather than falling back to keyless loopback or a session cookie.

A token scopes and revokes access for an agent that its environment already constrains. It is not a boundary against an agent with shell access as the daemon owner, which can read provider credentials or replace the daemon outright. Today the alternative is handing over the owner key, which reaches every account, raw SQL, backups, and shutdown.

Refs #666, slice 2


wesm:
working on this

PR #851, opened 2026-09-15
feat(import): add iMazing CSV archives

### LINKED ISSUE #832: Feature: import iMessage/SMS from iMazing CSV exports
## What I'm trying to do

Import my iMessage/SMS history from **iMazing** CSV exports.

## The gap

iMazing is one of the most widely used iOS backup/management tools, and its Messages export produces **CSV, PDF, TXT and XLSX** — never a raw database. msgvault has no CSV importer for any source, so those exports are unusable today.

`import-imessage` wants a macOS `chat.db`. (An iOS `sms.db` pulled out of an iMazing backup also imports cleanly in my testing, even though the docs only mention `chat.db`.) But that path only helps if you still have the backup *and* its encryption password. iOS backups are encrypted by default, and plenty of us have years-old iMazing CSV exports where the source backup or the password is long gone.

## Why it matters to me

I'm consolidating ~14 years of message history:

| Source | Coverage | Importable? |
|---|---|---|
| live `chat.db` | Nov 2022 → now, 158,382 msgs | ✅ |
| iOS backup `sms.db` | Jul 2020 → Sep 2022, 21,121 msgs | ✅ |
| **iMazing CSV exports** | **2012 → 2020** | ❌ |

So roughly eight years of my history exists only as CSV. Worth noting my Messages-in-iCloud history only reaches back to Nov 2022 — iCloud only ever received what was on-device the day it was enabled, so it will never backfill. For anyone in that boat, old exports are the only copy.

## The format

iMazing writes one CSV per conversation with a stable header:

```
Chat Session,Message Date,Delivered Date,Read Date,Service,Type,Sender ID,Sender Name,Status,Replying to,Subject,Text,Attachment,Attachment type
```

Sample row:

```
Jane Doe,2024-01-15 19:21:19,,2024-01-15 21:35:16,iMessage,Incoming,+15555550123,Jane Doe,Read,,,Example message body text,,
```

Layout:

```
export-root/
├── csv/            "Messages - <name or number>.csv"  (one per conversation)
└── attachments/    referenced files
```

Mapping onto the existing schema looks straightforward:

- `Type` (Incoming/Outgoing) → `is_from_me`
- `Service` (iMessage/SMS) → message type
- `Sender ID` → participant identifier (E.164 or email)
- `Chat Session` → conversation; group chats appear as `A & B & C`
- `Replying to` → threading
- `Attachment` / `Attachment type` → files in the sibling `attachments/` dir

## Suggested shape

```bash
msgvault import-imazing-csv <export-dir> --me +1XXXXXXXXXX [--contacts contacts.vcf]
```

## Known limitations, which I'd accept

- No stable per-message IDs, so dedup would have to be heuristic (timestamp + sender + text hash) rather than a real `source_message_id`
- Group participants derived from display names only — the same tradeoff `import-messenger` already makes with `<slug>@facebook.messenger`
- No reactions or edit history; read/delivered state limited to the columns above

Lower fidelity than a real database, no question. But for the years where the CSV is the only surviving copy, it's the difference between having that history and losing it.

salmonumbrella:
## What changed

- Add `import-imazing-csv` for iMazing Messages export roots and CSV directories, including comma, tab, and semicolon files.
- Preserve conversations, sender identity, service, delivery evidence, conservative reply links, and available attachments across deterministic reruns.
- Support an explicit IANA timezone and optional vCard enrichment, and document rerun and cross-source overlap behavior.

## Why

Some message histories survive only as iMazing CSV exports after the original Apple Messages database or encrypted backup is gone. Those exports can now join the archive without requiring the original backup.

## Usage

```bash
msgvault import-imazing-csv ~/Downloads/messages-export \
  --me +14155550100 --timezone America/Los_Angeles
```

Pass the export root or its `csv/` directory. `--me` is required; `--timezone` defaults to the local IANA zone, and `--contacts` accepts a vCard file for missing participant names.

Closes #832


salmonumbrella:
@wesm ready

PR #853, opened 2026-09-15
fix(web): shorten exploration URLs

wesm:
Shorten Web UI links by omitting defaults, keyboard focus, scroll position, and inactive workspace choices. Ordinary tabs now use readable links such as `?workspace=files&mode=full_text`; filters, layout changes, and selected items use a smaller exploration payload when needed.

Keep the complete session state in browser history so Back and Forward restore focus and choices from other tabs. Search mode stays explicit so a shared link does not pick up a different browser's saved preference.


wesm:
Review of the short-URL change, with fixes in a5d7a0aa.

**Findings: 1 P1, 2 P2, 2 P3. All fixed, none dismissed.**

- **P1** Six Playwright specs (`frontend` and `web-e2e` jobs) read keyboard focus and scroll position from the `explore` URL parameter. Those fields now live only in the history entry, so the assertions read `history.state` through a new shared helper in `web/tests/explore-state.ts`.
- **P2** The Relationships landing fallback in `AppShell.svelte` only checked for the old `explore` parameter. A shared `?workspace=relationships&mode=full_text` link looked like a default landing and would bounce to Everything when the hub was degraded. It now treats a `workspace` parameter as an explicit destination.
- **P2** The serializer pruned other-workspace fields by string prefix, which only worked because of branch order (`identityState` vs `identityQuery`, `relationshipReviewState` vs `relationshipTarget`). Replaced with an explicit field-to-workspace table. Side effect: a selected person stays in any Directory review link, not only a Fact review.
- **P3** The history entry construction was duplicated; it is now one helper.
- **P3** Tightened the Web UI doc paragraph, which said "choices from other workspaces" twice.

Locally: svelte-check, kit-ui check, 1831 vitest tests, and the full 98-test Playwright suite pass.

PR #861, opened 2026-09-16
feat(imap): retrieve, edit, and delete managed drafts

rodboev:
Retrieve, edit, and delete managed IMAP reply drafts through the CLI, with the selected daemon owning provider access and local persistence.

The daemon tracks draft ownership and revisions in SQLite or PostgreSQL. An edit saves replacement content before APPEND, publishes the new revision, then removes the exact old UID. Conditional deletion conflicts stop removal. Reused mailbox IDs preserve previously archived content.

Deletion failures before a remote write leave the draft retryable. When remote removal is confirmed, a retry can finish the local operation. Uncertain writes stay pending and retain content and known IMAP copies; explicit recovery remains a separate follow-up. Discarded drafts retain local content for `draft-get`.

Lifecycle commands remain owner-only. Committed changes refresh analytics even when cleanup or response delivery fails.

Refs #666, slice 3b.


rodboev:
The first finding above is a real race but the fix belongs in go-imap. I pushed a one commit fix to hstern's fork which was pinned to back in August as a temporary measure, but upstream still does not have his `QRESYNC`/`VANISHED`/`NOTIFY` commits, and there's no activity on his fork: my PR is the only one open.

So I think the right approach here is to fork upstream into kenn-io/go-imap, cherry-pick hstern's commits and add my `MODIFIED UID` one. @wesm @mariusvniekerk thoughts? 

Working on the second finding...

wesm:
rebasing

rodboev:
Fixing CI

wesm:
reviewing

PR #866, opened 2026-09-16
fix(api): check delegated draft permissions before running

rodboev:
The daemon checks `draft.create` before an agent's `draft-reply` request can run or make other work wait. The command handler and operation gate share this check. The draft command also checks that the token covers the requested account.

Only `draft.create` is accepted. Tests cover allowed and rejected requests and drafting for the correct account.

Refs #666.


wesm:
I narrowed this PR to the permission check that is useful today in f2b085b8.

`draft-reply` still requires `draft.create` for the account it uses. The daemon now checks the permission before starting the command or letting the request hold up other work. It still checks the account before contacting the mail server.

I removed `draft.read`, `draft.edit`, and `draft.delete`, along with the placeholder command maps and the help/docs that advertised them. Those commands do not exist yet. Issuing their permissions early would not help: tokens live in memory and are lost when the daemon restarts to install the future commands. Each permission can arrive with its working command.

I also removed the repeated vocabulary tests, the test of help wording, and the payload logs. The remaining tests cover the command handler, the operation gate, and a real HTTP draft request, including refusal when the token names a different account.

Build, vet, lint, docs checks, and the relevant Go tests pass locally. The PR title and description now match this smaller scope.

PR #871, opened 2026-09-17
feat(vector): add a SQLite semantic search accelerator

### LINKED ISSUE #870: Add a local SQLite semantic search accelerator
## The situation

SQLite semantic search scans stored vectors for each query. That retrieval work grows with the archive even when the query only needs a small result set.

Exact search works and remains useful as a fallback. Structured filters can narrow the search, but broad semantic and hybrid queries still need a way to avoid exhaustive retrieval. Rebuilding embeddings does not address that cost.

## Ask

Add a local SQLite search accelerator built from existing embeddings, with resumable construction, visible status, and an explicit switch back to exact search. Keep PostgreSQL behavior unchanged.

The implementation is ready for a PR.

salmonumbrella:
## What changed

- Add a SQLite approximate index for semantic, hybrid, and similar-message search, with exact reranking and fallback to exhaustive search.
- Add `msgvault embeddings optimize` to build or resume the index from stored embeddings without calling the embedding provider. `embeddings list` shows its state and progress.
- Keep ready indexes current as embeddings change, and expose query-embedding, retrieval, and hydration timings through the API, CLI, and MCP.
- Leave PostgreSQL search behavior unchanged.

## Why

Large SQLite archives currently scan stored vectors for each semantic query. The accelerator bounds candidate retrieval while keeping exact search available when the index is unavailable or explicitly disabled.

## Usage

After activating an embedding generation:

```bash
msgvault embeddings optimize
msgvault embeddings list
```

The default `[vector.search]` setting, `sqlite_accelerator = "auto"`, uses a ready index. Set it to `"exact"` to return to exhaustive search.

Closes #870


wesm:
working on this

Issue #728, opened 2026-08-30
Route voice notes and shared recordings from every messaging source to Docbank


salmonumbrella:
## The situation

Msgvault has one message graph but several ways speech arrives inside it: a first-class voice note, a generic audio attachment, voicemail, a meeting recording pointer, or a Loom/Cap URL pasted into any conversation. The words should be searchable the same way regardless of which importer found them.

Msgvault should not become the transcription or remote-download service. Its job is to capture enough source truth, route the exact attachment or protected recording reference to Docbank, retain the returned Docbank identity, and enforce message visibility when results come back. Docbank owns acquisition, Docling/ASR, normalized transcript evidence, embeddings, indexes, and retry state.

## What works today, and why it is not enough

The current importers expose most of the raw material, but not one complete route:

| Source | Current evidence | Gap before routing |
| --- | --- | --- |
| Beeper | Explicit `isVoiceNote`, audio metadata, duration, downloaded bytes, and sometimes a source transcript | Preserve full transcript provenance and route missing-transcript media |
| WhatsApp | Explicit `audio` and `voice_note` types, MIME, duration, and optional local media bytes | Route stored media and expose missing/unavailable media state |
| Slack | Authenticated `audio/*` downloads with bounded CAS storage | Preserve documented media fields such as duration/captions/subtype and route audio |
| Discord | Attachment MIME, duration, waveform, flags, and downloaded bytes | Preserve voice-message semantics and route eligible audio |
| Facebook Messenger exports | Explicit audio-file attachments with local paths | Route stored audio |
| SyncTech SMS/MMS | Arbitrary MMS parts with content type and stored bytes | Route audio MIME parts |
| Microsoft Teams | Typed call-recording URL plus reference attachments | Route recording references without downloading them in the Teams sync |
| Google Voice Takeout | Voicemail records and duration | Associate and import the voicemail audio when present, then route it |
| iMessage/SMS | `cache_has_attachments` only | Read the attachment join/table, safely resolve local files, retain audio metadata, and store exact bytes |

Beeper source transcripts already enter ordinary message text, so they are searchable. The other paths either archive opaque bytes, retain only a URL/metadata record, or—in iMessage and Google Voice voicemail—do not yet capture the audio at all. Loom and Cap links in ordinary message text have no provider-aware occurrence record.

## Ask

[...]

PR #876, opened 2026-09-18
feat(beeper): submit stored audio and transcripts to Docbank

rodboev:
Beeper audio can reach Docbank through an opt-in background job. Stored WAV and MP3 files backfill after `upload_consent` is set; missing transcripts remain unprocessed. OGG/Opus and M4A stay local with an `unsupported_media` code. Msgvault neither converts audio nor runs speech recognition.

The route adds HTTP delivery, two receipt tables and `[integrations.docbank]` configuration.

Each pass checks 100 attachments and 100 journal changes, then sends one Docbank request. Docbank owns processing consent and its media service; the pinned msgvault library stays unchanged. Hidden, deleted or replaced messages lose their local mapping while other messages sharing the audio keep theirs. Manual export remains available. Search and presentation follow separately.

The [status comment](https://github.com/kenn-io/msgvault/issues/728#issuecomment-5629248605) records prior capture work. This uses #783's Beeper transcript preservation and kenn-io/docbank#304's transcript support through kenn-io/docbank#346's shipped HTTP contract.

Refs #728


wesm:
Fixed the confirmed findings in `59d5f4d3`, `d9dd8d08`, and `5820a3dd`:

- Backfill now finishes and rests. Between scans, the worker follows attachment changes; a daily rescan catches late transcripts. Unchanged mappings skip write transactions, and discovery reads/parsing run outside the daemon operation lock.
- The media job no longer resets the idle timer. An unused daemon can shut down, while shutdown still cancels and waits for active uploads.
- Reactions no longer revoke audio mappings. When the last occurrence disappears before transcript delivery starts, that delivery stops waiting. Restoring an occurrence reopens it, and other messages sharing the recording keep their delivery.
- Upload copies now live under `data_dir/tmp/beeper-media` and are removed after each attempt. I kept Docbank's full-file inspector: a prefix would bypass its whole-file validation. The guide now explains its memory cost and the 1 GiB source limit; that limit is not a RAM cap.
- `operator_required` stops automatic polling and preserves the job for checking after operator intervention and a daemon restart. The guide now correctly says a status check can make up to three HTTP requests.
- Database read failures stop the pass for retry without creating a replacement gap mapping. Missing, malformed, or corrupt raw messages still record a source gap.
- Timestamp handling reuses the parsed envelope and preserves numeric offsets without labelling them UTC.

Regression tests cover these cases. Feature tests, focused race checks, `go vet`, `make lint-ci`, and `make docs-check` passed.

PR #880, opened 2026-09-19
feat(imap): recover identified draft operations

### LINKED ISSUE #862: Recover interrupted IMAP draft edits and deletions
Interrupted IMAP draft edits or deletions can leave a draft ID blocked after the mailbox has already changed. The lifecycle work in https://github.com/kenn-io/msgvault/pull/861 retains content and known provider identities, but manual mailbox cleanup leaves the ID blocked.

Add an explicit recovery command for identified copies. Recheck the caller's permissions, committed revision, mailbox generation and exact UID. Publish a known replacement or finish cleanup/deletion without repeating APPEND.

Keep replacements with unknown UIDs unresolved. Mailbox searches and adopting moved copies need a separate ownership design. Implement recovery after the lifecycle schema and failure contract land.

Refs #666, slice 3c

rodboev:
`draft-recover` resumes an interrupted IMAP draft edit or delete from its saved receipt. It can publish a known replacement or finish cleanup after the original copy is gone, without appending another copy.

The daemon checks the caller's source-scoped `draft.edit` or `draft.delete` grant before returning delegated revision or policy results. It then verifies the revision, source identity, mailbox generation, exact UID, and draft flags. Delegated responses include lifecycle metadata while omitting draft content, raw MIME, and candidate bytes.

Recovery uses only saved receipts and bytes. Unknown UIDs, moved copies, mailbox searches, and sync-archived replacements remain unresolved, so recovery cannot adopt an unrelated message.

Refs https://github.com/kenn-io/msgvault/issues/666#issuecomment-5585767456, slice 3c

Closes #862


wesm:
I checked the reported paths and pushed the fixes in 54b3e7d9 to this PR’s branch.

- Recovery can finish removing the original after an edit is saved, even if the replacement was later sent or deleted. Removing the extra replacement check also fixes the output that mislabeled the replacement as the old copy.
- Refused results retain their evidence after cancellation and report read/output failures. `pending_code` stays equal to the saved value. Refused results use `refusal_code`; pending cleanup reports the current provider result in `observation.code`.
- Delegated responses now skip content reads, so unreadable raw mail cannot turn a completed operation into an output failure. A missing draft after the source lock also returns the same `not_permitted` result as the first lookup.
- Repeated revision/source checks are shared, and the middleware comment and permission documentation now match the admitted commands.

For finding 1, I chose the documented manual recovery path. An original already marked `\Deleted` still needs an exact-UID expunge in an IMAP client before retrying recovery. Cancellation evidence does not establish who set that flag. The guide now explains the step and tells users to use the returned revision, since recovery may have saved a replacement first. A regression test exercises refusal, manual expunge, and successful completion.

`draft.edit` and `draft.delete` currently authorize recovery only. Delegated edit/delete commands remain outside this PR. An active draft with no pending operation still requires `draft.edit`, including after an aborted delete.

I removed the duplicate checks but kept the existing helper arguments. This repository explicitly disables numeric function-size and complexity limits. I kept the small permission tests as checks of permission independence and added regression coverage through the real command and HTTP paths.

The full test suite, vet, lint, and documentation checks passed.

